PR #1, opened 2018-03-26
App skeleton

jakearchibald:


webpack.config.js
```diff
@@ -0,0 +1,245 @@
...
+		resolveLoader: {
+			alias: {
+				async: path.join(__dirname, 'config/async-component-loader')
+			}
+		},
+		module: {
+			rules: [
+				{
+					test: /\.tsx?$/,
+					enforce: 'pre',
```
surma: Why?

webpack.config.js
```diff
@@ -0,0 +1,245 @@
...
+								localIdentName: '[local]__[hash:base64:5]',
+								namedExport: true,
+								camelCase: true,
+								importLoaders: 1,
+								sourceMap: isProd,
+								sass: true
+							}
+						}
+					]
+					// loader: ExtractTextPlugin.extract({
```
surma: We have some of these commented blocks. We should either remove them or add a comment explaining how and when and why they can be re-enabled

webpack.config.js
```diff
@@ -0,0 +1,245 @@
...
+		plugins: [
+			new ProgressBarPlugin({
+				format: '\u001b[90m\u001b[44mBuild\u001b[49m\u001b[39m [:bar] \u001b[32m\u001b[1m:percent\u001b[22m\u001b[39m (:elapseds) \u001b[2m:msg\u001b[22m',
+				renderThrottle: 100,
+				summary: false,
+				clear: true
+			}),
+
+			isProd && new CleanWebpackPlugin(['build']),
+			isProd && new webpack.optimize.SplitChunksPlugin({}),
```
surma: What does this plugin do? I can’t find it :(

webpack.config.js
```diff
@@ -0,0 +1,245 @@
...
+			isProd && new MiniCssExtractPlugin({}),
```
surma: Aren’t our CSS files separate already? What is the job of this plugin?

webpack.config.js
```diff
@@ -0,0 +1,245 @@
...
+					removeScriptTypeAttributes: true,
+					removeRedundantAttributes: true,
+					removeStyleLinkTypeAttributes: true,
+					removeComments: true
+				},
+				manifest,
+
+				/** @todo Finish implementing prerendering similar to that of Preact CLI. */
+				prerender() {
+					return '<div id="app_root"></div>';
```
surma: We can move our SSR markup into a serparate file later, right? Markup in `webpack.conf` seems odd to me.

webpack.config.js
```diff
@@ -0,0 +1,245 @@
...
+					// require('babel-register')({ ignore: false });
+					// return require('./config/prerender')();
+				},
+				inject: true,
+				compile: true
+			}),
+			isProd && new PreloadWebpackPlugin(),
```
surma: Why is this prod only? We should be keeping prod as close to dev as possible unless absolutely necessary.

webpack.config.js
```diff
@@ -0,0 +1,245 @@
...
+			new webpack.DefinePlugin({
```
surma: What does this do?

webpack.config.js
```diff
@@ -0,0 +1,245 @@
...
+				process: '{}'
+			}),
+			isProd && new ReplacePlugin({
```
surma: What does this do? Are we suppressing throws in prod here?

webpack.config.js
```diff
@@ -0,0 +1,245 @@
...
+				{ from: 'src/manifest.json', to: 'manifest.json' },
+				{ from: 'src/assets', to: 'assets' }
+			]),
+
+			isProd && new BundleAnalyzerPlugin({
+				analyzerMode: 'static',
+				defaultSizes: 'gzip',
+				openAnalyzer: false
+			})
+		].filter(Boolean),
```
surma: LOL! 

(I like it, but probably worth a comment)

webpack.config.js
```diff
@@ -0,0 +1,245 @@
...
+			isProd && new ReplacePlugin({
+				include: /babel-helper$/,
+				patterns: [{
+					regex: /throw\s+(new\s+)?(Type|Reference)?Error\s*\(/g,
+					value: s => `return;${Array(s.length - 7).join(' ')}(`
+				}]
+			}),
+			new CopyPlugin([
+				{ from: 'src/manifest.json', to: 'manifest.json' },
+				{ from: 'src/assets', to: 'assets' }
```
surma: `assets` contains images. Do we wanna process those? ImageOptim or something?

webpack.config.js
```diff
@@ -0,0 +1,245 @@
...
+			isProd && new BundleAnalyzerPlugin({
+				analyzerMode: 'static',
+				defaultSizes: 'gzip',
+				openAnalyzer: false
+			})
+		].filter(Boolean),
+
+		node: {
+			console: false,
+			global: true,
```
surma: Why are we polyfilling/mocking some of these? Do we need them?

config/client-boot.js
```diff
@@ -0,0 +1,27 @@
...
+if (process.env.NODE_ENV === 'development') {
+	// enable preact devtools
+	require('preact/debug');
+}
+else if (process.env.ADD_SW && 'serviceWorker' in navigator && location.protocol === 'https:') {
+	// eslint-disable-next-line no-undef
+	navigator.serviceWorker.register(__webpack_public_path__ + 'sw.js');
+}
+
+const interopDefault = m => m && m.default ? m.default : m;
```
surma: Expand this and add a couple of comments, the minifier will hopefully make it as small as it is now :D

config/client-boot.js
```diff
@@ -0,0 +1,27 @@
...
+
+let app = interopDefault(require('app-entry-point'));
```
surma: Can this be a normal static import? (Which would make `interopDefault` obsolete as well)

config/client-boot.js
```diff
@@ -0,0 +1,27 @@
...
+
+if (typeof app === 'function') {
```
surma: Can it ever be something else?

config/client-boot.js
```diff
@@ -0,0 +1,27 @@
...
+	let root = document.getElementById('app') || document.body.firstElementChild;
```
surma: We provide the markup. We don‘t need the fallback here I think.

config/client-boot.js
```diff
@@ -0,0 +1,27 @@
...
+
+	let init = () => {
+		let app = interopDefault(require('app-entry-point'));
```
surma: Didn’t we just do this?

config/client-boot.js
```diff
@@ -0,0 +1,27 @@
...
+		root = render(h(app), document.body, root);
```
surma: Overwriting `root` is somewhat confusing. They are different types, no?

src/lib/fix-pmc.js
```diff
@@ -0,0 +1,26 @@
+import { options } from 'preact';
+
+const classNameDescriptor = {
```
surma: What does this file do?

src/index.js
```diff
@@ -0,0 +1,18 @@
+import './style';
+import './lib/fix-pmc';
+import App from './components/app';
+
+export default App;
+
+if (typeof window!=='undefined') {
+	addEventListener('click', e => {
```
surma: This should be in a router component. Also we should be excluding external links from this handler.

src/index.js
```diff
@@ -0,0 +1,18 @@
...
+		let { target } = e;
+		do {
+			if (target.nodeName === 'A') {
+				history.pushState(null, null, target.pathname);
+				e.preventDefault();
+				return false;
+			}
+		} while ((target = target.parentNode));
```
surma: @jakearchibald says he has a platform method for this `do...while` loop

surma: It’s called `element.closest()`

src/components/app/index.tsx
```diff
@@ -0,0 +1,112 @@
+import { h, Component } from 'preact';
+import { updater, toggle, When } from '../../lib/util';
+import Fab from '../fab';
+import Header from '../header';
+// import Drawer from 'async!../drawer';
+const Drawer = require('async!../drawer').default;
```
surma: Why not `import(...)`?

src/components/app/index.tsx
```diff
@@ -0,0 +1,112 @@
...
+import Home from '../home';
+import * as style from './style.scss';
+
+type Props = {
+	url?: String
+}
+
+type FileObj = {
+	id: any,
```
surma: Why `any`?

src/components/app/index.tsx
```diff
@@ -0,0 +1,112 @@
...
+	data: any,
```
surma: Why `any`?

src/components/app/index.tsx
```diff
@@ -0,0 +1,112 @@
...
+	error: Error | DOMError | String,
```
surma: I think `Error` is enough?

src/components/app/index.tsx
surma: It’s a number, no?

src/components/app/index.tsx
```diff
@@ -0,0 +1,112 @@
...
+let counter = 0;
+
+export default class App extends Component<Props, State> {
+	state: State = {
+		showDrawer: false,
+		showFab: true,
+		files: []
+	};
+
+	loadFile = (file: File) => {
```
surma: Can this be a method?

src/components/app/index.tsx
```diff
@@ -0,0 +1,112 @@
...
+			error: null,
+			loading: true,
+			data: null
+		};
+
+		this.setState({
+			files: [fileObj]
+		});
+
+		let fr = new FileReader();
```
surma: Jake says you should new `new Response()` here ([example](https://github.com/jakearchibald/svgomg/blob/master/src/js/page/utils.js#L37))

I think we should move this into a `util` modules as `loadFileAsBlob()` or something.

src/components/app/index.tsx
```diff
@@ -0,0 +1,112 @@
...
+				loading: false
+			});
+			this.setState({ files });
+		};
+		fr.readAsDataURL(file);
+	};
+
+	enableDrawer = false;
+
+	openDrawer = updater(this, 'showDrawer', true);
```
surma: These should be proper methods. We can use TypeScript decorators to tie in stuff like `updater()`

src/components/app/index.tsx
```diff
@@ -0,0 +1,112 @@
...
+	toggleDrawer = updater(this, 'showDrawer', toggle);
+
+	openFab = updater(this, 'showFab', true);
+	closeFab = updater(this, 'showFab', false);
+	toggleFab = updater(this, 'showFab', toggle);
+
+	render({ url }, { showDrawer, showFab, files }) {
+		if (showDrawer) this.enableDrawer = true;
+
+		if (showFab===true) showFab = files.length>0;
```
surma: `if(showFab)`? :D

src/components/app/index.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+		return (
+			<div id="app" class={style.app}>
+				<Fab showing={showFab} />
+
+				<Header toggleDrawer={this.toggleDrawer} loadFile={this.loadFile} />
```
surma: Is `toggleDrawe` an event handler? If so, we probably wanna call it `onToggleDrawer` or something?

surma: (same for the other eventhandler-y stuff)

src/components/app/index.tsx
```diff
@@ -0,0 +1,112 @@
...
+				{/*
+					Note: this is normally where a <Router> with auto code-splitting goes.
+					Since we don't seem to need one (yet?), it's omitted.
+				*/}
+				<div class={style.content}>
+					<Home files={files} />
+				</div>
+
+				{/* This ends up in the body when prerendered, which makes it load async. */}
+				<link rel="stylesheet" href="https://fonts.googleapis.com/icon?family=Material+Icons" />
```
surma: It doesn’t make it load async, it just renders what is before it, and even that is not very widely implemented. We should figure out how we handle fonts in this project.

.babelrc
```diff
@@ -0,0 +1,33 @@
+{
+	"presets": [
+		[
+			"env",
+			{
+				"loose": true,
+				"uglify": true,
+				"modules": "commonjs",
```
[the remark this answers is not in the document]
developit: Yes - this was here for when I was trying to do prerendering without bundling via Webpack, but I've abandoned that attempt.

config/watch-timestamps-plugin.js
```diff
@@ -0,0 +1,23 @@
+let fs = require('fs');
+
+module.exports = WatchTimestampsPlugin;
+
+function WatchTimestampsPlugin(patterns) {
```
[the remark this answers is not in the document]
developit: Sure, they should be Node 4+.

config/watch-timestamps-plugin.js
```diff
@@ -0,0 +1,23 @@
+let fs = require('fs');
+
+module.exports = WatchTimestampsPlugin;
```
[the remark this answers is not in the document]
developit: Yup. I'll try to find the GH issue I retrofitted this from.

config/watch-timestamps-plugin.js
```diff
@@ -0,0 +1,23 @@
...
+
+function WatchTimestampsPlugin(patterns) {
+	this.patterns = patterns;
+}
+
+WatchTimestampsPlugin.prototype.apply = function (compiler) {
+	compiler.plugin('watch-run', (watch, callback) => {
+		let patterns = this.patterns;
+		let timestamps = watch.fileTimestamps || watch.compiler.fileTimestamps;
```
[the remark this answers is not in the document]
developit: Webpack 3 and prior vs Webpack 4+. Also I believe it may vary depending on whether the plugin is invoked in a child compiler instance.

webpack.config.js
```diff
@@ -0,0 +1,245 @@
...
+let HtmlWebpackPlugin = require('html-webpack-plugin');
+let PreloadWebpackPlugin = require('preload-webpack-plugin');
+let ReplacePlugin = require('webpack-plugin-replace');
+let CopyPlugin = require('copy-webpack-plugin');
+let WatchTimestampsPlugin = require('./config/watch-timestamps-plugin');
+let BundleAnalyzerPlugin = require('webpack-bundle-analyzer').BundleAnalyzerPlugin;
+
+let oldWrite = process.stderr.write;
+process.stderr.write = chunk => {
+	if (String(chunk).indexOf('DeprecationWarning:')!==-1) return;
```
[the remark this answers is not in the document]
developit: It just breaks the progress bar clearing, but I agree its a pretty ugly hack.

webpack.config.js
```diff
@@ -0,0 +1,245 @@
...
+	}
+	catch (e) {}
+}
+
+module.exports = function(_, env) {
+	let isProd = env.mode === 'production';
+	let nodeModules = path.join(__dirname, 'node_modules');
+	let componentStyleDirs = [
+		path.join(__dirname, 'src/components'),
+		path.join(__dirname, 'src/routes')
```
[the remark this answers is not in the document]
developit: I don't know if we are going to have routes. Originally I thought we wouldn't, but the mocks actually make it look like we will, so I think we should leave this in.

webpack.config.js
```diff
@@ -0,0 +1,245 @@
...
+	];
+
+	let babelRc = parseJson(readFile('.babelrc'));
+	babelRc.babelrc = false;
```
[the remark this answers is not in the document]
developit: It tells Webpack to avoid looking up babelrc files entirely.

webpack.config.js
```diff
@@ -0,0 +1,245 @@
...
+	if (isProd) babelRc.presets[0][1].modules = false;
```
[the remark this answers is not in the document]
developit: Maybe - it's kindof webpack specific - I was seeing Webpack's ESM scope hoisting break hot module replacement. Setting modules to "commonjs" in the babelrc (and `false` for prod) turns off that hoisting in development.

webpack.config.js
```diff
@@ -0,0 +1,245 @@
...
+
+	let manifest = parseJson(readFile('./src/manifest.json'));
+
+	return {
+		mode: env.mode || 'development',
```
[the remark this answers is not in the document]
developit: Can do, yeah. Nobody ever uses more than dev/prod.

webpack.config.js
```diff
@@ -0,0 +1,245 @@
...
+			chunkFilename: '[name].chunk.[chunkhash:5].js',
+			path: path.join(__dirname, 'build'),
+			publicPath: '/'
+		},
+		resolve: {
+			extensions: ['.ts', '.tsx', '.js', '.jsx', '.scss', '.css'],
+			alias: {
+				'app-entry-point': path.join(__dirname, 'src/index'),
+				style: path.join(__dirname, 'src/style'),
+				preact$: path.join(__dirname, 'node_modules/preact/dist/preact.js')
```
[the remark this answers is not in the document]
developit: This just uses the compressed version of Preact instead of the ESM version, since the compressed version is slightly smaller when bundled.  We could drop it, I doubt it saves more than a few bytes.

webpack.config.js
```diff
@@ -0,0 +1,245 @@
...
+			}
+		},
+		resolveLoader: {
+			alias: {
+				async: path.join(__dirname, 'config/async-component-loader')
```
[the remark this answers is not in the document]
developit: No, it is a component that wraps up dynamic imports. There are other ways to do this though if we want - my issue with dynamic import is that it's a Promise, which means it's impossible to perform a synchronous render even if the chunk is already available.

config/watch-timestamps-plugin.js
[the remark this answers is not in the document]
surma: I think we are all on Node 8, no? We should use all the goodness — our build system is not gonna turn into a library.

webpack.config.js
developit: So that it runs before Babel. May not be necessary though.

config/watch-timestamps-plugin.js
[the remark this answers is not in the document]
surma: Same as above — we decide what WebPack version we use.

webpack.config.js
developit: I'll remove them, they were from before I'd managed to get the new `mini-extract-plugin` working reliably.

webpack.config.js
developit: Webpack 4's documentation is still being written. The only reference to SplitChunksPlugin is in their blog post here:
https://medium.com/webpack/webpack-4-code-splitting-chunk-graph-and-the-splitchunks-optimization-be739a861366

- [ ] Annotate webpack config with doc links

webpack.config.js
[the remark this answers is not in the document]
surma: We reviewed this file first. Later on we found the aliases really confusing honestly. It took us a while to realize that “app-entry-point” was not a module but an alias, for example. I think we should use them sparingly.

webpack.config.js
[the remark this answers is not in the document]
surma: I don’t understand. Let’s talk this through in the meeting tomorrow! This seems like a core thing.

config/watch-timestamps-plugin.js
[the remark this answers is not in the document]
developit: yup.

config/watch-timestamps-plugin.js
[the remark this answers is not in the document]
developit: Agreed - likely this came from before I committed to v4.

webpack.config.js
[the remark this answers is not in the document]
developit: Yup - the entrypoint one is important because it makes generic HMR possible, but the style one is likely unnecessary, and the Preact one can  be removed.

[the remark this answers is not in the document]
surma: Maybe that’s better for tomorrow: How much do you value/enjoy HMR during development? I am very skeptical (I have _never_ used it so all this is just me having feelings) and it seems brittle.

webpack.config.js
developit: Webpack doesn't output CSS by default, instead it actually embeds it into the JS along with bootstrapping code that injects it as a Data URI into the head of the document.  Webpack 4 at least recognizes CSS as input by default, but [MiniCssExtractPlugin](https://twitter.com/wsokra/status/970253245733113856?lang=en) or ExtractTextWebPlugin needs to be used in order to rip the CSS back out of the bundled JS and produce files on disk.

webpack.config.js
developit: Yup. This is just temporary, I'm still working on SSR/prerendering. It'll be a plugin in `config/`.

webpack.config.js
developit: Turns this:

```js
if (process && process.env.NODE_ENV==='production')
```

... into this:

```js
if ({} && "production"==='production')
```

... which uglify can constant-fold.

webpack.config.js
developit: This is my favorite thing

webpack.config.js
developit: It strips out Babel class helper code that throws if classes are invoked without `new`, which is unnecessary for a production build, but UglifyJS does not statically remove it. If we migrate to Babel 7 (beta), this may no longer be necessary since `loose` mode omits these helpers.

webpack.config.js
developit: Maybe, though it'd be really slow to do that as part of the Webpack build.

webpack.config.js
developit: Console polyfill is just dead weight, `global` is required by some wepback loaders (it's just an alias of `window`), and the others are needed by Webpack's development tooling.

config/client-boot.js
developit: - [ ] comment

config/client-boot.js
developit: No, webpack needs a require() here since it does static analysis during build to find dependencies within `require.ensure()`.

config/client-boot.js
developit: It could be a VNode, but I don't mind making this file non-generic.

- [ ] remove generic case handling from client-boot

src/index.js
developit: Agreed - I had meant to rip this out, no need for it.

- [ ] remove garbage router code

webpack.config.js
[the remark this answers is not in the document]
developit: Yup, for sure. It's a bit opaque, and there is probably a way to at least use `require.ensure` here rather than the `async!` prefix, which reduces the amount of magic we have going on.

Another option is to dump manual async usage entirely and just let the existing mapping of `src/routes` handle it during build.

src/components/app/index.tsx
developit: I couldn't commit, haha.

- [ ] admit FileObj.id is just a number

src/components/app/index.tsx
developit: Nah, it's a DOMError which is apparently not subclassed in the TS definition.

src/components/app/index.tsx
developit: I wasn't sure which data type was going to be easiest for the compression pipeline to consume.  For now I believe I'm reading the file into a binary String, so I'll type it as such.

- [ ] Actually type FileObj.data.

src/components/app/index.tsx
developit: I'll try making a decorator. Personally I've been gravitating towards using a "store" to house data + mutators/actions.  I also threw some ideas around with @davideast last week for simplifying state+selectors, and he published something called [selectrify](https://github.com/davideast/selectrify).

src/components/app/index.tsx
developit: I assume you mean the left-hand side? If so I agree!

- [ ] `onToggleDrawer`

src/components/app/index.tsx
developit: Agreed, plus without prerendering my comment is inaccurate anyway (though I believe stylesheets added by script are async, unlike those dumped into body).

src/components/app/index.tsx
surma: I’d be in favor of decorators.

webpack.config.js
```diff
@@ -0,0 +1,245 @@
...
+let WatchTimestampsPlugin = require('./config/watch-timestamps-plugin');
+let BundleAnalyzerPlugin = require('webpack-bundle-analyzer').BundleAnalyzerPlugin;
+
+let oldWrite = process.stderr.write;
+process.stderr.write = chunk => {
+	if (String(chunk).indexOf('DeprecationWarning:')!==-1) return;
+	return oldWrite.call(process.stderr, chunk);
+};
+
+function parseJson(text, fallback) {
```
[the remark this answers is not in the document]
developit: - [ ] Error out if required json configs are missing/corrupted.

webpack.config.js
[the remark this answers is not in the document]
developit: I've removed the modules stuff entirely for now - it's possible the HMR issue I was seeing was just a webpack bug, I'm no longer seeing it.

src/index.js
developit: removed it anyway.

src/components/app/index.tsx
developit: Ended up being as simple as `new Response(file).text()`, not worth a util.

developit:
Alrighty @surma - that should be all the feedback acted on!

config/prerender-loader.js
```diff
@@ -0,0 +1,64 @@
...
+    let props = {};
+    // console.log(mod);
+    let vnode = preact.createElement(mod, props);
+    let frag = document.createElement('div');
+    frag.innerHTML = renderToString(vnode);
+    root.parentNode.replaceChild(frag.firstChild, root);
+
+    let html = dom.serialize();
+    callback(null, html);
+    // return html = `module.exports = ${JSON.stringify(html)}`;
```
surma: Can these comments all go?

src/components/app/style.scss
```diff
@@ -0,0 +1,25 @@
...
+
+  .header {
+    flex: 0 0 auto;
+    position: relative;
+  }
+
+  .content {
+    flex: 1 1 auto;
+    contain: size layout style;
+    overflow: auto;
```
surma: Do we really wanna scroll on content? It usually ends up causing problems. I’d prefer scrolling on body and having sidebar/topbar `position: fixed`.

src/components/drawer/index.tsx
```diff
@@ -0,0 +1,63 @@
+import { h, Component } from 'preact';
+import MdlDrawer from 'preact-material-components-drawer';
+import 'preact-material-components/Drawer/style.css';
+import List from 'preact-material-components/List';
+// import 'preact-material-components/List/style.css';
```
surma: remove?

src/components/app/index.tsx
```diff
@@ -0,0 +1,136 @@
+import { h, Component } from 'preact';
+import { When, bind } from '../../lib/util';
+import Fab from '../fab';
+import Header from '../header';
+// import Drawer from 'async!../drawer';
```
surma: Remove?

config/prerender.js
```diff
@@ -0,0 +1,20 @@
...
+
+let appPath = path.join(__dirname, '../src/index');
+
+module.exports = function(options) {
+  options = options || {};
+  let url = typeof options==='string' ? options : options.url;
+  global.history = {};
+  global.location = { href: url, pathname: url };
+
+  // let app = require('app-entry-point');
```
surma: Remove?

src/components/drawer/style.scss
```diff
@@ -0,0 +1,29 @@
+@import '~style/helpers.scss';
+
+:global {
+  // @import '~preact-material-components/Drawer/style.css';
```
surma: Remove?

src/components/drawer/style.scss.d.ts
```diff
@@ -0,0 +1,14 @@
+export const mdcListItemSecondaryText: string;
```
surma: Are these files generated? If so, should we really commit them?

src/components/header/style.scss
```diff
@@ -0,0 +1,52 @@
+@import '~style/helpers.scss';
+:global {
+  @import '~preact-material-components/Toolbar/mdc-toolbar.scss';
+}
+
+.toolbar {
+  // height: $toolbar-height;
```
surma: Remove comments?

src/components/home/index.tsx
```diff
@@ -0,0 +1,36 @@
+import { h, Component } from 'preact';
+// import Button from 'preact-material-components/Button';
```
surma: Remove?

src/components/home/style.scss
```diff
@@ -0,0 +1,20 @@
+@import '~style/helpers.scss';
+
+// :global {
```
surma: Remove?

src/index.html
```diff
@@ -0,0 +1,16 @@
...
+		<title>Squoosh</title>
+		<meta name="viewport" content="width=device-width,initial-scale=1">
+		<meta name="mobile-web-app-capable" content="yes">
+		<meta name="apple-mobile-web-app-capable" content="yes">
+		<meta name="theme-color" content="#673ab8">
+		<link rel="manifest" href="/manifest.json">
+	</head>
+	<body>
+		<div id="app" prerender></div>
+		<!-- <link rel="stylesheet" href="https://fonts.googleapis.com/icon?family=Material+Icons" /> -->
```
surma: Remove?

src/style/index.scss
```diff
@@ -0,0 +1,27 @@
+// @import './material-icons.scss';
```
surma: Remove?

src/style/material-icons.scss
```diff
@@ -0,0 +1,28 @@
+@font-face {
+  font-family: 'Material Icons';
+  font-style: normal;
+  font-weight: 400;
+  src: local('Material Icons'),
+    local('MaterialIcons-Regular'),
+    url(https://example.com/MaterialIcons-Regular.woff2) format('woff2'),
```
surma: Quotes like in `index.scss`?

src/style/reset.scss
```diff
@@ -0,0 +1,12 @@
+button, a, img, input, select, textarea {
+  -webkit-tap-highlight-color: rgba(0,0,0,0);
+}
+
+a, button, img, [inert], .inert {
+  user-select: none;
+  -webkit-user-select: none;
+  user-drag: none;
+  -webkit-user-drag: none;
```
surma: Should we use something like [autoprefixer](https://github.com/postcss/autoprefixer)?

jakearchibald:
In general, I think this PR is trying to do too much.

We don't really need large parts of the app complete, we just need a build system that supports a mix of assets, preact & web components.

I worry that by bringing in things like icon fonts, and making layout decisions, is just delaying this being able to land.

Any chance of keeping it to the basics? Then we can start planning and dividing up the rest of the work.

src/style/material-icons.scss
jakearchibald: I don't think this PR should contain anything like icon fonts.

src/style/reset.scss
jakearchibald: I don't think we should decide that as part of this PR.

src/components/app/style.scss
jakearchibald: I think we've got too much layout in the PR. If it's either-or, I'd rather we landed something that simply demonstrated that CSS could be included, and didn't otherwise worry about layout.

Layout is something we can plan & do in another PR.

surma:
I’m absolutely fine with punting on the specifics and picking those back up when they get touched.

Mostly want the dead code gone, tho.

src/lib/fix-pmc.js
jakearchibald: If we still need this, some comments would be handy, just to communicate the intent.

src/lib/util.ts
```diff
@@ -0,0 +1,49 @@
...
+};
+
+export class When extends Component<WhenProps, WhenState> {
+  state: WhenState = {
+    ready: !!this.props.value
+  };
+
+  render({ value, children = [] }: WhenProps, { ready }: WhenState) {
+    let child = children[0];
+    if (value && !ready) this.setState({ ready: true });
```
jakearchibald: What's the benefit of `ready` here? Can't you just test `value`?

src/index.html
```diff
@@ -0,0 +1,16 @@
...
+		<meta charset="utf-8">
+		<title>Squoosh</title>
+		<meta name="viewport" content="width=device-width,initial-scale=1">
+		<meta name="mobile-web-app-capable" content="yes">
+		<meta name="apple-mobile-web-app-capable" content="yes">
+		<meta name="theme-color" content="#673ab8">
+		<link rel="manifest" href="/manifest.json">
+	</head>
+	<body>
+		<div id="app" prerender></div>
```
jakearchibald: Why invent an attribute here rather than use a class name (or a data attribute)?

developit: fixed in subsequent branch (prerender)

src/lib/util.ts
developit: would create an infinite loop if the component loaded was undefined

developit:
note: rest of the stuff started here is getting its own PR's.

PR #11, opened 2018-03-29
Code review for two-up, pinch-zoom, trackDragging

jakearchibald:
You can `npm i && npm run start` in the components folder to build the test page.

See comments for which files to ignore.


components/.gitignore
```diff
@@ -0,0 +1,2 @@
+node_modules
```
jakearchibald: Only for the demo stuff. This will be removed in the final merge.

components/MultiPanel/index.ts
```diff
@@ -0,0 +1,12 @@
+import './styles.css';
```
jakearchibald: This is a stub I created for @kosamari. Feel free to ignore.

components/MultiPanel/styles.css
```diff
@@ -0,0 +1,3 @@
+multi-panel {
```
jakearchibald: This is a stub I created for @kosamari. Feel free to ignore.

components/package.json
```diff
@@ -0,0 +1,33 @@
+{
```
jakearchibald: This is just so I can build the test page. Will be removed in the final merge in favour of the build system @developit is working on.

components/test/car.svg
```diff
@@ -0,0 +1 @@
..."M563 396.1c59.9 105.6 72.6 148.1 82.3 302.3C659.9 571.5 617.9 460 563 396z"/><path fill="#a00" d="M460.2 642.1c22.2 56.2 40 120 38.4 172.6 13.2-48-4.7-142.3-38.4-172.6z"/><path fill="#2b0000" d="M439.2 613a485.8 485.8 0 0 1 59.1 194.3c7-75.1-21.9-151.4-59.1-194.3z"/><path fill="red" d="M449.3 622.8c10.7 17.8 62.6 26.8 66.2-5.4 3.6-32.2-39.4-35.8-59-25-31.3 17-7.2 28.6-7.2 30.4z"/><path fill="#a00" d="M449.3 622.8c9 18.3 66.1 27.7 66-8-.2-32.4-42.2-12.7-61.8-11.4-20.2 1.5-9.4 15.8-4.2 19.4z"/><path fill="#6d0000" d="M449.3 622.8c17.2 22.2 59.2 17.4 53.6-2.5-7-24.9-29.4 2.7-47.2-4.3-23-8.9-13.3 5.4-6.4 6.8z"/><rect width="64.8" height="23.2" x="382.1" y="322.8" fill="#1a1a1a" color="#000" opacity=".5" rx="5.3" ry="3.9" transform="matrix(.96122 .27577 -.20533 .9787 0 0)"/><path fill="#6d0000" d="M29.6 784.4c-3.7 10.6-8.2 36.8-1.7 45.8 1.5-12.2 2-28.4 1.7-45.8z"/><path fill="#6d0000" d="M24.5 826.3c3.4 11.3 130.5 59 141.2 55.7-76.4-26.9-139-56.7-141.2-55.7z"/><path fill="#963500" style="marker:none" d="M332 782.8c-1.1.2-3 .6-4.2 1.3-3.4 1.8-7 2.7-10.8 3a249.7 249.7 0 0 0-6 60.3 86.9 86.9 0 0 0 34.3-9.6c20.3-10 4.6-57.6-13.2-55z" color="#000" opacity=".5" overflow="visible"/></g></svg>
```
jakearchibald: Just for the test page.

components/test/custom.d.ts
```diff
@@ -0,0 +1,4 @@
+declare module "*.svg" {
```
jakearchibald: Just for the test page.

components/test/index.html
```diff
@@ -0,0 +1,10 @@
+<!DOCTYPE html>
```
jakearchibald: Just for the test page.

components/test/index.ts
```diff
@@ -0,0 +1,69 @@
+import '@webcomponents/custom-elements';
```
jakearchibald: Just for the test page.

components/test/style.css
```diff
@@ -0,0 +1,41 @@
+.container {
```
jakearchibald: Just for the test page.

components/tsconfig.json
```diff
@@ -0,0 +1,13 @@
+{
```
jakearchibald: This will be merged with the overall project rules.

components/tslint.json
```diff
@@ -0,0 +1,3 @@
+{
```
jakearchibald: This will be merged with the overall project rules.

components/webpack.config.js
```diff
@@ -0,0 +1,40 @@
+const path = require('path');
```
jakearchibald: This will be dropped in favour of @developit's build system.

components/test/index.ts
jakearchibald: Although we'll need the event retargeting stuff.

components/PinchZoom/index.ts
```diff
@@ -0,0 +1,274 @@
...
+  /**
+   * Called when the direct children of this element change.
+   * Until we have have shadow dom support across the board, we
+   * require a single element to be the child of <pinch-zoom>, and
+   * that's the element we pan/scale.
+   */
+  private _stageElChange () {
+    this._positioningEl = undefined;
+
+    if (this.children.length === 0) {
```
surma: Okay, _crazy_ idea: `this.children.length === 1 && this.children[0] instanceof HTMLElement`

components/PinchZoom/index.ts
```diff
@@ -0,0 +1,274 @@
...
+      return;
+    }
+
+    this._positioningEl = el;
+
+    if (this.children.length > 1) {
+      console.warn('<pinch-zoom> must not have more than one child.');
+    }
+
+    this.setTransform();
```
surma: But... calling `setTransform` with the default values means it will just bail, did you mean to do something else here?

components/PinchZoom/index.ts
```diff
@@ -0,0 +1,274 @@
...
+        if (pointerTracker.currentPointers.length === 2) return false;
+        event.preventDefault();
+        return true;
+      },
+      move: previousPointers => {
+        this._onPointerMove(previousPointers, pointerTracker.currentPointers);
+      }
+    });
+
+    this.addEventListener('wheel', event => this._onWheel(event));
```
surma: Willing to use `@bind` instead?

components/PinchZoom/index.ts
```diff
@@ -0,0 +1,274 @@
...
+      console.warn('<pinch-zoom> must not have more than one child.');
+    }
+
+    this.setTransform();
+  }
+
+  private _onWheel (event: WheelEvent) {
+    event.preventDefault();
+
+    const thisRect = this.getBoundingClientRect();
```
surma: Is there any reasonable way we can cache this? Listen to `document.onresize`?

components/PinchZoom/index.ts
```diff
@@ -0,0 +1,274 @@
...
+    const { ctrlKey, deltaMode } = event;
+
+    if (deltaMode === 1) { // 1 is "lines", 0 is "pixels"
+      // Firefox uses "lines" for some types of mouse
+      deltaY *= 15;
+    }
+
+    // ctrlKey is true when pinch-zooming on a trackpad.
+    const divisor = ctrlKey ? 100 : 300;
+    const scaleDiff = -deltaY / divisor + 1;
```
surma: I’d prefer `1 - deltaY / divisor`

components/PinchZoom/index.ts
```diff
@@ -0,0 +1,274 @@
...
+    scaleDiff = 1,
+    originX = 0,
+    originY = 0
+  }: {
+    panX?: number,
+    panY?: number,
+    scaleDiff?: number,
+    originX?: number,
+    originY?: number
+  } = {}) {
```
surma: Same as above, please.

components/PinchZoom/index.ts
```diff
@@ -0,0 +1,274 @@
...
+   * Update the stage with a given scale/x/y.
+   */
+  setTransform (scale: number = this._scale, x: number = this._x, y: number = this._y, {
+    allowChangeEvent = false
+  }: {
+    /**
+     * Fire a 'change' event if values are different to current values
+     */
+    allowChangeEvent?: boolean
+  } = {}) {
```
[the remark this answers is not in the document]
surma: Or we even go full TypeScript:

```typescript
const setTransformDefaultOpts = {
    allowChangeEvent: false;
};
function setTransform(scale: number, x: number, y: number, opts: Partial<typeof setTransformDefaultOpts>?) {
    opts = { ...setTransformDefaultOpts, ...opts };
    /* ... */
}
```

components/PinchZoom/index.ts
```diff
@@ -0,0 +1,274 @@
...
+
+    let matrix = createMatrix()
+      // Translate according to panning.
+      .translate(panX, panY)
+      // Scale about the origin.
+      .translate(originX, originY)
+      .scale(scaleDiff)
+      .translate(-originX, -originY);
+
+    // Calculate the final position & clamp to bounds.
```
surma: Maybe add a comment per case?

components/PinchZoom/index.ts
```diff
@@ -0,0 +1,274 @@
...
+  }
+
+  get scale () {
+    return this._scale;
+  }
+
+  /**
+   * Update the stage with a given scale/x/y.
+   */
+  setTransform (scale: number = this._scale, x: number = this._x, y: number = this._y, {
```
surma: Wouldn’t it be easier if `setTransform` took a matrix and we took the a matrix as our source-of-truth?

components/PinchZoom/index.ts
```diff
@@ -0,0 +1,274 @@
...
+    this._applyChange({
+      scaleDiff,
+      originX: event.clientX - thisRect.left,
+      originY: event.clientY - thisRect.top
+    });
+  }
+
+  private _onPointerMove (previousPointers: Pointer[], currentPointers: Pointer[]) {
+    // Combine next points with previous points
+    const thisRect = this.getBoundingClientRect();
```
surma: I don’t think we can afford to call `gBCR` on a move event. We really should be looking into caching these.

components/PinchZoom/styles.css
```diff
@@ -0,0 +1,14 @@
...
+  display: block;
+  overflow: hidden;
+  touch-action: none;
+  --scale: 1;
+  --x: 0;
+  --y: 0;
+}
+
+pinch-zoom > * {
+  transform: translate(var(--x), var(--y)) scale(var(--scale));
```
surma: I really like this approach of going via custom props!

components/TwoUp/index.ts
```diff
@@ -0,0 +1,94 @@
...
+import { PointerTracker, Pointer } from '../../utils/trackDragging';
+
+/**
+ * A split view that the user can adjust. The first child becomes
+ * the left-hand side, and the second child becomes the right-hand side.
+ */
+export default class TwoUp extends HTMLElement {
+  private readonly _handle = document.createElement('div');
+  /**
+   * The position of the split.
```
surma: In pixels? Percent?

components/TwoUp/index.ts
```diff
@@ -0,0 +1,94 @@
...
+          pointerTracker.startPointers[0],
+          pointerTracker.currentPointers[0]
+        );
+      }
+    });
+  }
+
+  connectedCallback () {
+    this._childrenChange();
+    if (!this._everConnected) {
```
surma: IIUC this code _should_ be run on every connect.

components/TwoUp/index.ts
```diff
@@ -0,0 +1,94 @@
...
+    // Watch for children changes.
+    // Note this won't fire for initial contents,
+    // so _childrenChange is also called in connectedCallback.
+    new MutationObserver(() => this._childrenChange())
+      .observe(this, { childList: true });
+
+    // Watch for pointers on the handle.
+    const pointerTracker: PointerTracker = new PointerTracker(this._handle, {
+      start: (pointer, event) => {
+        // We only want to track 1 pointer.
```
surma: “We only want to track 1 pointer.” but then it early-exits when there is only pointer.

components/TwoUp/styles.css
```diff
@@ -0,0 +1,40 @@
+two-up {
+  display: grid;
+  --split-point: 0;
+}
+
+two-up > * {
+  grid-area: 1/1;
```
surma: I know what this does because of your tweet. But a comment would be good.

components/TwoUp/styles.css
```diff
@@ -0,0 +1,40 @@
...
+  top: 50%;
+  left: 50%;
+  transform: translate(-50%, -50%);
+  width: 80px;
+  height: 40px;
+  background: red;
+  border-radius: 20px;
+}
+
+two-up > :nth-child(1):not(.two-up-handle) {
```
surma: Is the second `:not()` necessary considering the code makes sure that the handle is always the last child?

utils/trackDragging.ts
```diff
@@ -0,0 +1,234 @@
...
+   * Track pointers across a particular element
+   *
+   * @param element Element to monitor.
+   * @param callbacks
+   */
+  constructor (private _element: HTMLElement, {
+    start = () => true,
+    move = noop,
+    end = noop
+  }: PointerTrackerCallbacks) {
```
surma: Same here.

utils/trackDragging.ts
```diff
@@ -0,0 +1,234 @@
...
+    this._startCallback = start;
+    this._moveCallback = move;
+    this._endCallback = end;
+
+    // Bind listener methods
+    this._pointerStart = this._pointerStart.bind(this);
```
surma: `@bind`?

utils/trackDragging.ts
```diff
@@ -0,0 +1,234 @@
...
+
+  /**
+   * Listener for mouse/pointer starts. Bound to the class in the constructor.
+   *
+   * @param event This will only be a MouseEvent if the browser doesn't support
+   * pointer events.
+   */
+  private _pointerStart (event: PointerEvent | MouseEvent) {
+    // Only interested in left-button presses.
+    if (event.button !== 0) return;
```
surma: Maybe define an `enum` for `LEFT_BUTTON`?

utils/trackDragging.ts
```diff
@@ -0,0 +1,234 @@
...
+    }
+  }
+
+  /**
+   * Listener for pointer/mouse/touch move events.
+   * Bound to the class in the constructor.
+   */
+  private _move (event: PointerEvent | MouseEvent | TouchEvent) {
+    const previousPointers = this.currentPointers.slice();
+    const changedPointers = ('changedTouches' in event) ?
```
surma: Can you use `isPointerEvent` here?

components/PinchZoom/index.ts
```diff
@@ -0,0 +1,274 @@
+import './styles.css';
+import { PointerTracker, Pointer } from '../../utils/trackDragging';
+
+interface Point {
+  clientX: number;
+  clientY: number;
+}
+
+function getDistance (a: Point, b?: Point): number {
```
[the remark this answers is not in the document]
jakearchibald: Convenience really, so the distance is zero when there's only one pointer.

components/PinchZoom/index.ts
jakearchibald: Absolutely

components/PinchZoom/index.ts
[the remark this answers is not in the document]
surma: But can that ever happen? Wouldn’t we rather throw a compile-time error when you call this function with just one argument?

components/PinchZoom/index.ts
jakearchibald: In person discussion:

This shouldn't take a matrix as it can express things that we don't want to support here, eg rotation & 3d stuff.

However, using a matrix privately to store state makes sense.

utils/PointerTracker.ts
```diff
@@ -0,0 +1,234 @@
...
+   * Track pointers across a particular element
+   *
+   * @param element Element to monitor.
+   * @param callbacks
+   */
+  constructor (private _element: HTMLElement, {
+    start = () => true,
+    move = noop,
+    end = noop
+  }: PointerTrackerCallbacks) {
```
surma: Same here

utils/PointerTracker.ts
```diff
@@ -0,0 +1,234 @@
...
+    this._startCallback = start;
+    this._moveCallback = move;
+    this._endCallback = end;
+
+    // Bind listener methods
+    this._pointerStart = this._pointerStart.bind(this);
```
surma: @bind?

utils/PointerTracker.ts
```diff
@@ -0,0 +1,234 @@
...
+
+  /**
+   * Listener for mouse/pointer starts. Bound to the class in the constructor.
+   *
+   * @param event This will only be a MouseEvent if the browser doesn't support
+   * pointer events.
+   */
+  private _pointerStart (event: PointerEvent | MouseEvent) {
+    // Only interested in left-button presses.
+    if (event.button !== 0) return;
```
surma: Maybe define an `enum` for `LEFT_BUTTON`?

utils/PointerTracker.ts
```diff
@@ -0,0 +1,234 @@
...
+    }
+  }
+
+  /**
+   * Listener for pointer/mouse/touch move events.
+   * Bound to the class in the constructor.
+   */
+  private _move (event: PointerEvent | MouseEvent | TouchEvent) {
+    const previousPointers = this.currentPointers.slice();
+    const changedPointers = ('changedTouches' in event) ?
```
surma: Can you use `isPointerEvent` here?

components/PinchZoom/index.ts
```diff
@@ -0,0 +1,274 @@
...
+
+  get scale () {
+    return this._scale;
+  }
+
+  /**
+   * Update the stage with a given scale/x/y.
+   */
+  setTransform (scale: number = this._scale, x: number = this._x, y: number = this._y, {
+    allowChangeEvent = false
```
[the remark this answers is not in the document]
jakearchibald: I thought the same, but doesn't `fireChangeEvent` make it sound like it *will* fire a change event, whereas in reality it might not, if the provided values match the current state, or are clamped to be equal to the current state?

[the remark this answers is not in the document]
surma: I think the word “change” covers that.

components/PinchZoom/index.ts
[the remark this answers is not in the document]
jakearchibald: After in person discussion, we decided:

**BAD JakeScript**

```ts
private _applyChange ({
  panX = 0,
  panY = 0,
  scaleDiff = 1,
  originX = 0,
  originY = 0
}: {
  panX?: number,
  panY?: number,
  scaleDiff?: number,
  originX?: number,
  originY?: number
} = {}) {
  // …
}
```

**HaloumiScript**

```ts
interface ApplyChangeOpts {
  panX?: number;
  panY?: number;
  scaleDiff?: number;
  originX?: number;
  originY?: number;
}
```

```ts
private _applyChange (opts: ApplyChangeOpts = {}) {
  const {
    panX = 0, panY = 0,
    originX = 0, originY = 0,
    scaleDiff = 1
  } = opts;
  // …
}
```

Any objections / better ideas @developit @kosamari?

components/PinchZoom/index.ts
```diff
@@ -0,0 +1,274 @@
...
+    this._x = x;
+    this._y = y;
+    this._scale = scale;
+
+    this.style.setProperty('--x', this._x + 'px');
+    this.style.setProperty('--y', this._y + 'px');
+    this.style.setProperty('--scale', this._scale + '');
+
+    if (allowChangeEvent) {
+      const event = new Event('change', { bubbles: true });
```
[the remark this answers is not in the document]
jakearchibald: Asked in web platform kin.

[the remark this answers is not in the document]
jakearchibald: Seems like the only reason to use `CustomEvent` is for `.detail`, which we don't need here (and wouldn't want to use in TypeScript).

components/PinchZoom/index.ts
jakearchibald: I don't know why `_positioningEl` needs to be an `HTMLElement`. I complained at one point, but I've refactored since, so it shouldn't be an issue.

components/PinchZoom/index.ts
jakearchibald: I might still need this, but some stuff has gone awry in a refactor.

I need to:

* Ensure bounds are respected when calling `setTransform`.
* `setTransform` should be callable before adding `<pinch-zoom>` to the document, but bounds will need to be rechecked once it's added to the document.

components/PinchZoom/index.ts
jakearchibald: From in-person chat: There's no way to know bounds haven't changed after caching.

This is a common problem though, and we should have a think about what we need on the platform to make this easier.

components/PinchZoom/index.ts
jakearchibald: Same caching issue as above.

components/TwoUp/index.ts
jakearchibald: It's pixels now, so I'll update the comment. At some point I need to come up with a better way to handle resizing, since it's currently possible to lose the handle. One way will be to use a percent value.

components/TwoUp/index.ts
jakearchibald: That's because `currentPointers` doesn't include this pointer, as we haven't accepted it yet. Maybe that's confusing?

components/TwoUp/index.ts
jakearchibald: If the handle is set to 0, then the element is moved, I wouldn't expect the handle to change position.

components/TwoUp/styles.css
jakearchibald: If there are no children it'll be both the last and first child.

components/TwoUp/index.ts
surma: A bit yeah. But it might be that I have’t fully grasped this class yet.

components/PinchZoom/index.ts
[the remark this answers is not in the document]
jakearchibald: It happens a lot – if you're dragging using the mouse, or with one finger on the screen.

Alternatively I could throw an if around every usage of `getDistance` and `getMidpoint`, but that feels messier.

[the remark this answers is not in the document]
surma: Fair.

components/PinchZoom/index.ts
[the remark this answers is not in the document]
jakearchibald: @developit @kosamari what do you think?

```js
pinchZoom.setTransform(1, 100, 100, {
  allowChangeEvent: true
});
```

Or `fireChangeEvent`?

If you set this property to true, it may not fire a change event, eg if the supplied or computed values are the same as the current transform.

components/PinchZoom/index.ts
jakearchibald: Sold

components/PinchZoom/index.ts
jakearchibald: Sold

components/PinchZoom/index.ts
developit: A primitive for that would be amazing...

components/PinchZoom/index.ts
developit: random thought: What about deferring it until the next render, then reacting only if it changed?

utils/PointerTracker.ts
developit: Maybe amend when I merge - I exported it from `src/lib/util.ts`

jakearchibald:
I still need to take a look at `setTransform` in pinch-zoom, but my brain's tired. I botched a refactor. I need to ensure that transforms can be set before the element is connected, and that bounds are checked when `setTransform` is called.

Also, bounds should be checked on connect.

jakearchibald:
Thanks for the review @surma! Good to see some conventions for the rest of the project coming out of this.

utils/PointerTracker.ts
jakearchibald: TIL `const enum` in TypeScript.

I'd avoided enums previously because of the objects they create.

utils/PointerTracker.ts
jakearchibald: This is checking for a touch event, not a pointer event.

I could replace this with: `self.TouchEvent && event instanceof TouchEvent`, but I went for the property check since it's smaller.

components/PinchZoom/index.ts
developit: @todo maybe try ResizeObserver here?

jakearchibald: Resize observer is useful in cases where we need to immediately react to a change in layout size. Whereas in this case, we only need it on-demand, and we need the position relative to the viewport which resize observer doesn't give us.

Besides, resize observer is Chrome-only, so it doesn't make sense with our target browser support.

surma: (Just for completeness sake: If we had broad RO support, you could use ROs to cache the most recent layout values and not have to worry about forcing sync layout.)

utils/PointerTracker.ts
```diff
@@ -0,0 +1,237 @@
...
+  /** ID for this pointer */
+  id: number = -1;
+
+  constructor (nativePointer: Touch | PointerEvent | MouseEvent) {
+    this.pageX = nativePointer.pageX;
+    this.pageY = nativePointer.pageY;
+    this.clientX = nativePointer.clientX;
+    this.clientY = nativePointer.clientY;
+
+    if (self.Touch && nativePointer instanceof Touch) {
```
surma: Why not put this in a `isTouchPointer()` and `isPointerEvent()`?

utils/PointerTracker.ts
```diff
@@ -0,0 +1,237 @@
...
+    }
+  }
+
+  /**
+   * Listener for pointer/mouse/touch move events.
+   * Bound to the class in the constructor.
+   */
+  private _move (event: PointerEvent | MouseEvent | TouchEvent) {
+    const previousPointers = this.currentPointers.slice();
+    const changedPointers = ('changedTouches' in event) ? // Shortcut for 'is touch event'.
```
surma: re-use `isTouchEvent`?

utils/PointerTracker.ts
jakearchibald: It's only used once. Happy to put it in a function if you'd prefer though.

jakearchibald:
Closing this, as I'm rolling the components into master.

utils/PointerTracker.ts
jakearchibald: There's no such method

PR #12, opened 2018-04-03
CSS Inlining

developit:
This adds a new plugin (which I think could be released on its own, and I have a funny name for). It plugs into the output of `html-webpack-plugin`, finds inline & external stylesheets, inlines them, then trims out selectors & rules not used by the document.

In this app, the effect is a drop from 13kb to 1.2kb of critical CSS.

(note: currently targeted at `app-skeleton` branch)

Still to do:

- [x] If the stylesheet being inlined is external, replace the `<link>` tag with an async one (preload -> style).

/cc @surma

### DIFF
--- a/config/critters-webpack-plugin.js
+++ b/config/critters-webpack-plugin.js
@@ -235,7 +235,7 @@ function getElementsByTagName(tagName) {
 const ElementExtensions = {
   nodeName: {
     get() {
-      return this.tagName;
+      return this.tagName.toUpperCase();
     }
   },
   insertBefore(child, referenceNode) {
@@ -279,7 +279,12 @@ const DocumentExtensions = {
   // nwmatcher requires that it at least report a correct nodeType of DOCUMENT_NODE.
   nodeType: {
     get() {
-      return 11;
+      return 9;
+    }
+  },
+  nodeName: {
+    get() {
+      return '#document';
     }
   },
   createElement(name) {
@@ -295,10 +300,10 @@ const DocumentExtensions = {
     return node;
   },
   querySelector(sel) {
-    return this.$match.first(sel);
+    return this.$match.first(sel, this.documentElement);
   },
   querySelectorAll(sel) {
-    return this.$match.select(sel);
+    return this.$match.select(sel, this.documentElement);
   },
   getElementsByTagName,
   // nwmatcher uses inexistence of `document.addEventListener` to detect IE:
--- a/config/critters-webpack-plugin.js
+++ b/config/critters-webpack-plugin.js
@@ -5,94 +5,20 @@ const nwmatcher = require('nwmatcher');
 const css = require('css');
 const prettyBytes = require('pretty-bytes');
 
-const treeUtils = parse5.treeAdapters.htmlparser2;
+const treeAdapter = parse5.treeAdapters.htmlparser2;
 
 const PLUGIN_NAME = 'critters-webpack-plugin';
 
 const PARSE5_OPTS = {
-  treeAdapter: treeUtils
-};
-
-function defineProperties(obj, properties) {
-  for (const i in properties) {
-    const value = properties[i];
-    Object.defineProperty(obj, i, typeof value === 'function' ? { value: value } : value);
-  }
-}
-
-const ElementExtensions = {
-  nodeName: {
-    get: function() {
-      return this.tagName;
-    }
-  },
-  insertBefore: function (child, referenceNode) {
-    if (!referenceNode) return this.appendChild(child);
-    treeUtils.insertBefore(this, child, referenceNode);
-    return child;
-  },
-  appendChild: function (child) {
-    treeUtils.appendChild(this, child);
-    return child;
-  },
-  removeChild: function (child) {
-    treeUtils.detachNode(child);
-  },
-  setAttribute: function (name, value) {
-    if (this.attribs == null) this.attribs = {};
-    if (value == null) value = '';
-    this.attribs[name] = value;
-  },
-  removeAttribute: function(name) {
-    if (this.attribs != null) {
-      delete this.attribs[name];
-    }
-  },
-  getAttribute: function (name) {
-    return this.attribs != null && this.attribs[name];
-  },
-  hasAttribute: function (name) {
-    return this.attribs != null && this.attribs[name] != null;
-  },
-  getAttributeNode: function (name) {
-    const value = this.getAttribute(name);
-    if (value!=null) return { specified: true, value: value };
-  },
-  getElementsByTagName: getElementsByTagName
-};
-
-const DocumentExtensions = {
-  nodeType: {
-    get: function () {
-      return 11;
-    }
-  },
-  createElement: function (name) {
-    return treeUtils.createElement(name, null, []);
-  },
-  createTextNode: function (text) {
-    const scratch = this.$$scratchElement;
-    treeUtils.insertText(scratch, text);
-    const node = scratch.lastChild;
-    treeUtils.detachNode(node);
-    return node;
-  },
-  querySelector: function (sel) {
-    return this.$match.first(sel);
-  },
-  querySelectorAll: function (sel) {
-    return this.$match.select(sel);
-  },
-  getElementsByTagName: getElementsByTagName,
-  // https://github.com/dperini/nwmatcher/blob/3edb471e12ce7f7d46dc1606c7f659ff45675a29/src/nwmatcher.js#L353
-  addEventListener: Object
+  treeAdapter
 };
 
 /** Critters: Webpack Plugin Edition!
  *  @class
  *  @param {Object} options
  *  @param {Boolean} [options.external=true]  Fetch and inline critical styles from external stylesheets
- *  @param {Boolean} [options.async=true]     If `false`, only already-inline stylesheets will be reduced to critical rules.
+ *  @param {Boolean} [options.async=false]    Convert critical-inlined external stylesheets to load asynchronously (via link rel="preload")
+ *  @param {Boolean} [options.minify=false]   Minify resulting critical CSS using cssnano
  */
 module.exports = class CrittersWebpackPlugin {
   constructor(options) {
@@ -100,59 +26,60 @@ module.exports = class CrittersWebpackPlugin {
   }
 
   apply(compiler) {
-    const self = this;
     const outputPath = compiler.options.output.path;
-    compiler.hooks.compilation.tap(PLUGIN_NAME, function (compilation) {
-      compilation.hooks.htmlWebpackPluginAfterHtmlProcessing.tapAsync(PLUGIN_NAME, function (htmlPluginData, callback) {
+
+    // hook into the compiler to get a Compilation instance...
+    compiler.hooks.compilation.tap(PLUGIN_NAME, compilation => {
+      // ... which is how we get an "after" hook into html-webpack-plugin's HTML generation.
+      compilation.hooks.htmlWebpackPluginAfterHtmlProcessing.tapAsync(PLUGIN_NAME, (htmlPluginData, callback) => {
+        // Parse the generated HTML in a DOM we can mutate
         const document = parse5.parse(htmlPluginData.html, PARSE5_OPTS);
+        makeDomInteractive(document);
 
-        defineProperties(document, DocumentExtensions);
-        document.documentElement = document.childNodes[document.childNodes.length - 1];
+        // `external:false` skips processing of external sheets
+        const externalSheets = this.options.external===false ? [] : document.querySelectorAll('link[rel="stylesheet"]');
 
-        const scratch = document.$$scratchElement = document.createElement('div');
-        const elementProto = Object.getPrototypeOf(scratch);
-        defineProperties(elementProto, ElementExtensions);
-        elementProto.ownerDocument = document;
-
-        document.$match = nwmatcher({ document });
-        document.$match.configure({
-          CACHING: false,
-          USE_QSAPI: false,
-          USE_HTML5: false
-        });
-
-        const externalSheets = document.querySelectorAll('link[rel="stylesheet"]');
-
-        Promise.all(externalSheets.map(function(link) {
-          if (self.options.external===false) return;
+        Promise.all(externalSheets.map(link => {
           const href = link.getAttribute('href');
+
+          // skip network resources
           if (href.match(/^(https?:)?\/\//)) return Promise.resolve();
+
+          // path on disk
           const filename = path.resolve(outputPath, href.replace(/^\//, ''));
+
+          // try to find a matching asset by filename in webpack's output (not yet written to disk)
           const asset = compilation.assets[path.relative(outputPath, filename).replace(/^\.\//, '')];
+
+          // wait for a disk read if we had to go to disk
           const promise = asset ? Promise.resolve(asset.source()) : readFile(filename);
-          return promise.then(function (sheet) {
+          return promise.then(sheet => {
+            // the reduced critical CSS gets injected into a new <style> tag
             const style = document.createElement('style');
-            style.$$name = href;
             style.appendChild(document.createTextNode(sheet));
             link.parentNode.insertBefore(style, link.nextSibling);
-            if (self.options.async) {
+
+            // drop a reference to the original URL onto the tag (used for reporting to console later)
+            style.$$name = href;
+
+            // the `async` option changes any critical'd <link rel="stylesheet"> tags to async-loaded equivalents
+            if (this.options.async) {
               link.setAttribute('rel', 'preload');
               link.setAttribute('as', 'style');
               link.setAttribute('onload', "this.rel='stylesheet'");
             }
           });
         }))
-          .then(function() {
+          .then(() => {
+            // go through all the style tags in the document and reduce them to only critical CSS
             const styles = document.$match.byTag('style');
-            return Promise.all(styles.map(function (style) {
-              return self.processStyle(style, document);
-            }));
+            return Promise.all(styles.map(style => this.processStyle(style, document)));
           })
-          .then(function () {
+          .then(() => {
             const html = parse5.serialize(document, PARSE5_OPTS);
             callback(null, { html });
           })
-          .catch(function (err) {
+          .catch((err) => {
             callback(err);
           });
       });
@@ -161,16 +88,22 @@ module.exports = class CrittersWebpackPlugin {
 
   processStyle(style, document) {
     let done = Promise.resolve();
-    let sheet = style.childNodes.length>0 && style.childNodes.map(getNodeValue).join('\n');
+
+    // basically `.textContent`
+    let sheet = style.childNodes.length>0 && style.childNodes.map( node => node.nodeValue ).join('\n');
+
+    // store a reference to the previous serialized stylesheet for reporting stats
+    const before = sheet;
+
+    // Skip empty stylesheets
     if (!sheet) return done;
 
     const ast = css.parse(sheet);
 
-    const before = sheet;
-
-    visit(ast, function (rule) {
+    // Walk all CSS rules, transforming unused rules to comments (which get removed)
+    visit(ast, (rule) => {
       if (rule.type==='rule') {
-        rule.selectors = rule.selectors.filter(function (sel) {
+        rule.selectors = rule.selectors.filter((sel) => {
           // Remove unknown pseudos as they break nwmatcher
           sel = sel.replace(/::?(?:[a-z-]+)([.[#~&^:*]|\s|\n|$)/gi, '$1');
           return document.querySelector(sel, document) != null;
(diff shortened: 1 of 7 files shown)


config/html-webpack-inline-critical-css-plugin.js
```diff
@@ -0,0 +1,256 @@
...
+
+const PLUGIN_NAME = 'html-webpack-inline-critical-css-plugin';
+
+const PARSE5_OPTS = {
+  treeAdapter: treeUtils
+};
+
+function defineProperties(obj, properties) {
+  for (let i in properties) {
+    let value = properties[i];
```
mathiasbynens: Why are these not `const`?

config/html-webpack-inline-critical-css-plugin.js
```diff
@@ -0,0 +1,256 @@
...
+  nodeType: {
+    get: function () {
+      return 11;
+    }
+  },
+  createElement: function (name) {
+    return treeUtils.createElement(name, null, []);
+  },
+  createTextNode: function (text) {
+    let scratch = this.$$scratchElement;
```
mathiasbynens: `const`

config/html-webpack-inline-critical-css-plugin.js
```diff
@@ -0,0 +1,256 @@
...
+    treeUtils.insertText(scratch, text);
+    let node = scratch.lastChild;
```
mathiasbynens: Same here

mathiasbynens: …and throughout the rest of the file, really

config/html-webpack-inline-critical-css-plugin.js
jakearchibald: +1 we should only use `let` if the value is changed.

config/html-webpack-inline-critical-css-plugin.js
```diff
@@ -0,0 +1,256 @@
...
+        let externalSheets;
+        externalSheets = document.querySelectorAll('link[rel="stylesheet"]');
+
+        Promise.all(externalSheets.map(function(link) {
+          const href = link.getAttribute('href');
+          if (href.match(/^(https?:)?\/\//)) return Promise.resolve();
+          const filename = path.resolve(outputPath, href.replace(/^\//, ''));
+          let asset = compilation.assets[path.relative(outputPath, filename).replace(/^\.\//, '')];
+          let promise = asset ? Promise.resolve(asset.source()) : readFile(filename);
+          return promise.then(function (sheet) {
```
jakearchibald: Would this be easier to read with async functions?

config/html-webpack-inline-critical-css-plugin.js
developit: whoop, old habits die hard. I imagine there's a lint rule to remind me of this.

config/html-webpack-inline-critical-css-plugin.js
developit: yup. I was avoiding the transpile step. that said, I would vastly prefer async here.

developit:
@jakearchibald Might be worth adding an option to scan for strings that look like classnames in inline JS?  Or even simpler, an option to force the existence of selector parts.

Right now this is "css required to style the static html".

config/html-webpack-inline-critical-css-plugin.js
```diff
@@ -0,0 +1,255 @@
...
+    if (!sheet) return done;
+
+    const ast = css.parse(sheet);
+
+    const before = sheet;
+
+    visit(ast, function (rule) {
+      if (rule.type==='rule') {
+        rule.selectors = rule.selectors.filter(function (sel) {
+          if (sel.match(/:(hover|focus|active)([.[#~&^:*]|\s|\n|$)/)) return false;
```
jakearchibald: Does this mean `:hover` rules etc are always lost?

config/html-webpack-inline-critical-css-plugin.js
```diff
@@ -0,0 +1,255 @@
...
+
+const treeUtils = parse5.treeAdapters.htmlparser2;
+
+const PLUGIN_NAME = 'html-webpack-inline-critical-css-plugin';
+
+const PARSE5_OPTS = {
+  treeAdapter: treeUtils
+};
+
+function defineProperties(obj, properties) {
```
surma: Why are you polyfilling `defineProperties`?

config/html-webpack-inline-critical-css-plugin.js
```diff
@@ -0,0 +1,255 @@
...
+  }
+}
+
+const ElementExtensions = {
+  nodeName: {
+    get: function() {
+      return this.tagName;
+    }
+  },
+  appendChild: function (child) {
```
surma: I can haz `appendChild(child) { /*...*/ }` syntax?

config/html-webpack-inline-critical-css-plugin.js
```diff
@@ -0,0 +1,255 @@
...
+
+  apply(compiler) {
+    const self = this;
+    const outputPath = compiler.options.output.path;
+    compiler.hooks.compilation.tap(PLUGIN_NAME, function (compilation) {
+      compilation.hooks.htmlWebpackPluginAfterHtmlProcessing.tapAsync(PLUGIN_NAME, function (htmlPluginData, callback) {
+        const document = parse5.parse(htmlPluginData.html, PARSE5_OPTS);
+
+        defineProperties(document, DocumentExtensions);
+        document.documentElement = document.childNodes[document.childNodes.length - 1];
```
surma: ... Is that guaranteed to be correct?

config/html-webpack-inline-critical-css-plugin.js
```diff
@@ -0,0 +1,255 @@
...
+          CACHING: false,
+          USE_QSAPI: false,
+          USE_HTML5: false
+        });
+
+        const externalSheets = document.querySelectorAll('link[rel="stylesheet"]');
+
+        Promise.all(externalSheets.map(function(link) {
+          const href = link.getAttribute('href');
+          if (href.match(/^(https?:)?\/\//)) return Promise.resolve();
```
surma: This any stylesheet that is included via a full URL is ignored, right?

config/html-webpack-inline-critical-css-plugin.js
```diff
@@ -0,0 +1,255 @@
...
+  // https://github.com/dperini/nwmatcher/blob/3edb471e12ce7f7d46dc1606c7f659ff45675a29/src/nwmatcher.js#L353
+  addEventListener: Object
+};
+
+module.exports = class HtmlWebpackInlineCriticalCssPlugin {
+  constructor(options) {
+    this.options = options || {};
+  }
+
+  apply(compiler) {
```
surma: Willing to make this async and use `async`/`await`?

config/html-webpack-inline-critical-css-plugin.js
developit: yes. I was waffling on them and I think I might remove this clause - hover and focus are valid states for a static HTML document.

config/html-webpack-inline-critical-css-plugin.js
```diff
@@ -0,0 +1,255 @@
...
+        document.$match = nwmatcher({ document });
+        document.$match.configure({
+          CACHING: false,
+          USE_QSAPI: false,
+          USE_HTML5: false
+        });
+
+        const externalSheets = document.querySelectorAll('link[rel="stylesheet"]');
+
+        Promise.all(externalSheets.map(function(link) {
```
surma: Maybe put the map function in a named top-level function so that the name describes what its doing?

config/html-webpack-inline-critical-css-plugin.js
```diff
@@ -0,0 +1,255 @@
...
+            callback(null, { html });
+          })
+          .catch(function (err) {
+            callback(err);
+          });
+      });
+    });
+  }
+
+  processStyle(style, document) {
```
surma: Comment?

config/html-webpack-inline-critical-css-plugin.js
```diff
@@ -0,0 +1,255 @@
...
+          const href = link.getAttribute('href');
+          if (href.match(/^(https?:)?\/\//)) return Promise.resolve();
+          const filename = path.resolve(outputPath, href.replace(/^\//, ''));
+          const asset = compilation.assets[path.relative(outputPath, filename).replace(/^\.\//, '')];
+          const promise = asset ? Promise.resolve(asset.source()) : readFile(filename);
+          return promise.then(function (sheet) {
+            const style = document.createElement('style');
+            style.$$name = href;
+            style.appendChild(document.createTextNode(sheet));
+            link.parentNode.appendChild(style);  // @TODO insertBefore
```
surma: If you `return style` here, then you don’t need to `byTag()` in the next `.then()`, no?

config/html-webpack-inline-critical-css-plugin.js
surma: I thought this meant that they are always kept, but I am also not 100% sure. Comments please? 😄

config/html-webpack-inline-critical-css-plugin.js
```diff
@@ -0,0 +1,255 @@
...
+        // If there are no remaining rules, replace the parent rule with a comment.
+        if (rule.rules.length===0) {
+          rule.type = 'comment';
+          rule.comment = '';
+          delete rule.rules;
+        }
+      }
+    });
+
+    sheet = css.stringify(ast, { compress: true });
```
surma: `... {compress: this.options.compress}`?

config/html-webpack-inline-critical-css-plugin.js
```diff
@@ -0,0 +1,255 @@
...
+
+    if (this.options.minimize || this.options.compress || this.options.minify) {
+      const cssnano = require('cssnano');
```
surma: Move this to the top.

config/html-webpack-inline-critical-css-plugin.js
```diff
@@ -0,0 +1,255 @@
...
+  processStyle(style, document) {
+    let done = Promise.resolve();
+    let sheet = style.childNodes.length>0 && style.childNodes.map(getNodeValue).join('\n');
+    if (!sheet) return done;
+
+    const ast = css.parse(sheet);
+
+    const before = sheet;
+
+    visit(ast, function (rule) {
```
surma: The callback could be async as well

config/html-webpack-inline-critical-css-plugin.js
```diff
@@ -0,0 +1,255 @@
...
+  node.rules.forEach(function (rule) {
+    if (rule.rules) {
+      visit(rule, fn);
+    }
+    fn(rule);
+  });
+}
+
+
+function readFile(file) {
```
surma: WDYT about:

```js
const {promisify} = require('utils');
const fs = require('fs');
const readFile = promisify(fs.readFile);
```

config/html-webpack-inline-critical-css-plugin.js
```diff
@@ -0,0 +1,255 @@
...
+
+function getNodeValue(node) {
+  return node.nodeValue;
+}
+
+function notComment(rule) {
+  return rule.type !== 'comment';
+}
+
+function getElementsByTagName(tagName) {
```
surma: I mean, it doesn’t really matter. But a recursive function could be easier to read:

```js
function getElementsByTagName(tagName) {
  if (this.nodeType !== 1) return []; // Only elements are relevant
  tagName = tagName.toUpperCase();
  return [
    ...((this.tagName == tagName || tagName === '*') ? tagName : [])), // Add current element if it matches tag
    ...this.children.map(child => getElementByTagName(tagName).call(child)) // Check children recursively
  ];
}
```

config/html-webpack-inline-critical-css-plugin.js
```diff
@@ -0,0 +1,255 @@
...
+};
+
+function defineProperties(obj, properties) {
+  for (const i in properties) {
+    const value = properties[i];
+    Object.defineProperty(obj, i, typeof value === 'function' ? { value: value } : value);
+  }
+}
+
+const ElementExtensions = {
```
surma: Comment?

config/html-webpack-inline-critical-css-plugin.js
```diff
@@ -0,0 +1,255 @@
...
+    return this.attribs != null && this.attribs[name] != null;
+  },
+  getAttributeNode: function (name) {
+    const value = this.getAttribute(name);
+    if (value!=null) return { specified: true, value: value };
+  },
+  getElementsByTagName: getElementsByTagName
+};
+
+const DocumentExtensions = {
```
surma: Comment?

surma:
Oh and I forgot: Can we have some tests infra for this? I’d love to use this, but I’d like to be able to test (esp. if you wanna go ahead and publish this and we start running into corner cases).

config/html-webpack-inline-critical-css-plugin.js
developit: Wasn't - most of the values are functions that get assigned as pure `value`s, this made that a bit easier (rather than having to specific `{ value() {} }` for each prototype method.

config/html-webpack-inline-critical-css-plugin.js
developit: Just looked and its Node 4+ so yes!

config/html-webpack-inline-critical-css-plugin.js
surma: Oh I see. Comment please :D

config/html-webpack-inline-critical-css-plugin.js
```diff
@@ -0,0 +1,255 @@
...
+        document.$match.configure({
+          CACHING: false,
+          USE_QSAPI: false,
+          USE_HTML5: false
+        });
+
+        const externalSheets = document.querySelectorAll('link[rel="stylesheet"]');
+
+        Promise.all(externalSheets.map(function(link) {
+          const href = link.getAttribute('href');
```
mathiasbynens: Use `link.href` to always get a normalized URL string (even when the attribute value is a relative URL, for example).

config/html-webpack-inline-critical-css-plugin.js
surma: Out of curiosity — what is your baseline for build tooling like this? My gut feeling would be LTS, so Node 8.

surma: Nvm, asked this in chat.

config/html-webpack-inline-critical-css-plugin.js
developit: nah, this DOM (parse5) is extremely rudimentary and only supports attributes.

config/html-webpack-inline-critical-css-plugin.js
developit: Nope. I'll have it filter by tagName.

config/html-webpack-inline-critical-css-plugin.js
developit: Yes. I might also add an attribute option to filter by, or make this filterable via a regex passed as an option.

config/html-webpack-inline-critical-css-plugin.js
developit: Just annoying to bind.

config/html-webpack-inline-critical-css-plugin.js
developit: It scans for all style tags, not just those embedded by this function.

config/html-webpack-inline-critical-css-plugin.js
developit: I need to compress it in order to have it remove the comments (that's how I'm doing AST node removal).

config/html-webpack-inline-critical-css-plugin.js
developit: But then it'll be imported regardless of whether it gets used

surma:
I’m gonna wait with my 2nd review until you push the async/await :)

config/html-webpack-inline-critical-css-plugin.js
developit: yup

config/html-webpack-inline-critical-css-plugin.js
developit: I've removed the cssnano integration - never resulted in any compression for me.

config/html-webpack-inline-critical-css-plugin.js
developit: remove this entirely

config/html-webpack-inline-critical-css-plugin.js
developit: how come? it's a synchronous walk.

config/html-webpack-inline-critical-css-plugin.js
developit: yup, if we're doing 8+ that's all this does.

config/html-webpack-inline-critical-css-plugin.js
developit: easier to read but quite a bit slower. also the weird edge cases are to handle the fact that this dom doesn't normalize the case of tagName

developit: Whoop, here I was spouting "facts" without testing. Recursion is *much* faster here:

<a href="https://esbench.com/bench/5ac3b647f2949800a0f619e1"></a>

I'll update with the ones from that [bench](https://esbench.com/bench/5ac3b647f2949800a0f619e1).

surma: Even if it wasn’t, I’d prefer the more readable/concise version (unless the difference is on the magnitude of seconds). It’s a build step after all.

surma: Also we can probably safe the `bind()` and turn the element into a normal parameter, if that makes it more readable.

developit: The bind is because this gets mixed into Element.prototype.

My benchmark was flawed, stack is around 5x faster, or 2x faster than an optimized recursive version:



(same url: https://esbench.com/bench/5ac3b647f2949800a0f619e1)

config/critters-webpack-plugin.js
```diff
@@ -0,0 +1,304 @@
...
+
+        // `external:false` skips processing of external sheets
+        if (this.options.external!==false) {
+          const externalSheets = document.querySelectorAll('link[rel="stylesheet"]');
+          externalStylesProcessed = Promise.all(externalSheets.map(
+            link => this.embedLinkedStylesheet(link, compilation, outputPath)
+          ));
+        }
+
+        externalStylesProcessed
```
surma: async/await?

config/critters-webpack-plugin.js
```diff
@@ -0,0 +1,304 @@
...
+    return fn(rule)!==false;
+  });
+}
+
+
+/** Enhance an htmlparser2-style DOM with basic manipulation methods. */
+function makeDomInteractive(document) {
+  defineProperties(document, DocumentExtensions);
+  // Find the first <html> element within the document
+  document.documentElement = document.childNodes.filter( child => String(child.tagName).toLowerCase()==='html' )[0];
```
surma: Have you linted this file with semistandard? The spaces look suspicious here

config/critters-webpack-plugin.js
```diff
@@ -0,0 +1,304 @@
...
+
+    // path on disk
+    const filename = path.resolve(outputPath, href.replace(/^\//, ''));
+
+    // try to find a matching asset by filename in webpack's output (not yet written to disk)
+    const asset = compilation.assets[path.relative(outputPath, filename).replace(/^\.\//, '')];
+
+    // wait for a disk read if we had to go to disk
+    const promise = asset ? Promise.resolve(asset.source()) : readFile(filename, 'utf8');
+    return promise.then(sheet => {
```
surma: async/await?

config/critters-webpack-plugin.js
```diff
@@ -0,0 +1,304 @@
...
+      link.parentNode.insertBefore(style, link.nextSibling);
+
+      // drop a reference to the original URL onto the tag (used for reporting to console later)
+      style.$$name = href;
+
+      // the `async` option changes any critical'd <link rel="stylesheet"> tags to async-loaded equivalents
+      if (this.options.async) {
+        link.setAttribute('rel', 'preload');
+        link.setAttribute('as', 'style');
+        link.setAttribute('onload', "this.rel='stylesheet'");
```
surma: What does this `onload` handler do o_O

config/critters-webpack-plugin.js
```diff
@@ -0,0 +1,304 @@
...
+        }
+      }
+
+      // If there are no remaining rules, remove the whole rule.
+      return !rule.rules || rule.rules.length!==0;
+    });
+
+    sheet = css.stringify(ast, { compress: this.options.compress!==false });
+
+    return done.then(() => {
```
surma: async/await? 😬

surma:
Wait... did I just put “async/await?” everywhere and you are still working on that? Apologies if that’s the case. I thought you pushed a commit. Sorryyy

config/critters-webpack-plugin.js
developit: coming :)

config/critters-webpack-plugin.js
developit: I'll switch over - right now it is using my fork but needn't be.

config/critters-webpack-plugin.js
developit: swaps the preload to a stylesheet once it's loaded:
https://www.filamentgroup.com/lab/async-css.html

developit:
lol still workin on it. will be up in a jiffy - my day got torched yesterday.

config/html-webpack-inline-critical-css-plugin.js
mathiasbynens: Ooh, gotcha. Fingers crossed dealing with raw attribute values doesn’t break anything, I guess!

PR #17, opened 2018-04-17
Only use Babel to process .js files.

developit:


jakearchibald:
@developit Do we need to use babel on JS files?

.babelrc
```diff
@@ -17,10 +17,8 @@
 		]
 	],
 	"plugins": [
-		"syntax-dynamic-import",
 		"transform-decorators-legacy",
```
surma: Are we using decorators outside TS?

PR #20, opened 2018-04-17
Add firebase hosting

developit:


jakearchibald:
I'm ok with this PR but we shouldn't consider it a done deal. We should also look at serving straight from a storage bucket.

I'm curious about:

* Differences in cost.
* Differences in using a custom domain.
* Control over response headers, and the ease of setting them (the current headers are bad).

We also need to make sure this is set up internally so billing goes to the right place.

developit:
Definitely just an interim thing. I'd assumed we'd get this set up somewhere more official soon, just wanted a place with proper h2 + http redirects so lighthouse was happy.

surma:
My 2c: I like FB Hosting as a target. You have full control over headers (more so than GCS) and it’s pretty frictionless. But I’ll go with whatever.

developit:
That was the other killer feature here. We could grab the `firebase.json` Link header generation bit from preact-cli and get automatic h/2 push.

surma:
Gonna be honest: I don’t wanna deal with h2 push and all its problems. I’d rather not bring it into this project.

developit:
Even so, the Link header for preload can still provide an advantage over preloading via meta tags, right?

surma:
Sorry, I kinda side-tracked this PR with my comment — my comment does not veto this PR at all.

(And yes, I think being able to control HTTP Link headers can be valuable.)

surma:
@developit Please rebase & fix once you are back at work :)

surma:
🎉

Issue #14, opened 2018-04-15
Some styles not getting typescript definitions


jakearchibald:
https://github.com/GoogleChromeLabs/squoosh/tree/components-merge I've started trying to merge components in here.

* `src/components/app/style.scss` - definitions generated
* `src/components/output/custom-els/PinchZoom/styles.css` - definitions generated
* `src/components/output/style.scss` - **no definitions generated**

I can't figure out why definitions are missing for that last resource. Any ideas?

developit:
This should work as of the `prerendering` branch. The issue was that `ts-loader` was running prior to the `typings-for-css-modules-loader`.

jakearchibald:
It isn't working in https://github.com/GoogleChromeLabs/squoosh/pull/22, which is based on preloading.

jakearchibald:
I've uncommented the line in #22 that triggers the build error.

jakearchibald:
https://github.com/GoogleChromeLabs/squoosh/pull/22/commits/552beb5fb8d57f1a3d890388697b763b21034a5a

jakearchibald:
Update: It's only working if you use the import, and even then it fails the first time.

PR #22, opened 2018-04-18
Removing everything that isn't skeletonyy

jakearchibald:
I'm trying to remove stuff all the frontend stuff that isn't ready / hasn't been discussed / isn't part of what we're building.

I'll leave notes against particular changes…


src/components/app/style.scss
```diff
@@ -1,25 +1,3 @@
...
-  }
-
-  .content {
-    flex: 1 1 auto;
-    contain: size layout style;
-    overflow: auto;
-    -webkit-overflow-scrolling: touch;
-  }
+.app h1 {
+  color: green;
```
jakearchibald: This is just to demo how styles work, but it should be clear I'm not making any style decisions.

src/components/output/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+  componentDidUpdate({ img }: Props) {
+    if (img !== this.props.img) this.updateCanvas(this.props.img);
+  }
+
+  render({ img }: Props, { }: State) {
+    return (
+      <div>
+        <canvas ref={c => this.canvas = c as HTMLCanvasElement} width={img.width} height={img.height} />
```
jakearchibald: This is where I'll be introducing the web components for pinch-zooming etc

src/index.tsx
```diff
@@ -20,10 +20,6 @@ if (process.env.NODE_ENV === 'development') {
       root = render(<App />, document.body, root);
     });
   });
-} else if ('serviceWorker' in navigator && location.protocol === 'https:') {
-  addEventListener('load', () => {
-    navigator.serviceWorker.register(__webpack_public_path__ + 'sw.js');
-  });
```
jakearchibald: I don't think it's time to add a service worker yet. Also we wouldn't want it to be https-only, as testing locally is important.

webpack.config.js
```diff
@@ -234,22 +232,6 @@ module.exports = function (_, env) {
...
-        exclude: [
-          'report.html',
-          'manifest.json',
-          /(report\.html|manifest\.json|\.precache-manifest\..*\.json)$/,
-          /\.(?:map|pem|DS_Store)$/
-        ],
-        // allow for offline client-side routing:
-        navigateFallback: '/',
-        navigateFallbackBlacklist: [/\.[a-z0-9]+$/i]
       })
```
jakearchibald: I don't think it's time to add a service worker yet. Also, they shouldn't be prod only, else it becomes easy to forget and break the site.

webpack.config.js
```diff
@@ -277,8 +259,6 @@ module.exports = function (_, env) {
       compress: true,
       // Request paths not ending in a file extension serve index.html:
       historyApiFallback: true,
-      // Don't output server address info to console on startup:
-      noInfo: true,
```
jakearchibald: Important to know the port number.

jakearchibald:
@developit does this make sense? I feel bad hitting the delete key a lot, but I think we should start simple so others on the project can get their heads around it, and it's clear which decisions have been made.

src/components/output/index.tsx
developit: yup!

webpack.config.js
developit: Any harm? I did it because we know this will be a PWA, and our requirements are pretty basic - static files with deterministic hashed names, all precached aggressively.

developit:
I'm good with simplicity. I don't think it's worth removing Workbox since that is a nontrivial setup that we'll just end up adding again later since this is by definition a PWA (in that it must work offline).  I think we should leave the SW in, remove the development and https conditions, and wait until it gives us pain.

webpack.config.js
surma: > static files with deterministic hashed names, all precached aggressively

tbqh that’s not basic in my book. 

It will be interesting for a lot of people to document how we went from a non-PWA preact app to a full PWA one.

jakearchibald:
Like my choice of default serif fonts & placeholder text, the use of an unstyled `<input type="file">` is just a means to load an image, not part of the final design or interaction.

I'd rather we got ourselves into a state where we could iterate on the compression options, than be blocked perfecting the intro screen.

developit:
In that case I think we ought to merge all these open PRs then! ;)

jakearchibald:
@developit this is now based on your latest prerendering branch. Happy with this?

https://github.com/GoogleChromeLabs/squoosh/issues/14 is still an issue here though.

PR #24, opened 2018-04-24
WIP : multi panels (for question comment purpose)

kosamari:
Just making PR because commenting UI is better on PR.

Approach I took was to accept any HTML elements as children of `<multi-panels>` 

`<multi-panels>` will simply take even indexed elements as heading, and element following(thus odd index) as content.  Allowing user of this component more freedom in markup.

note: I don't really understand TypeScript(yet) and this might very well be not great code(yet) !

### DIFF
--- a/components/MultiPanel/index.ts
+++ b/components/MultiPanel/index.ts
@@ -9,6 +9,10 @@ export default class MultiPanel extends HTMLElement {
 
   constructor () {
     super();
+    
+    // add EventListners
+    this.addEventListener('click', this._onClick);
+    this.addEventListener('keydown', this._onKeyDown);
 
     // Watch for children changes.
     new MutationObserver(() => this._childrenChange())
@@ -17,57 +21,20 @@ export default class MultiPanel extends HTMLElement {
 
   connectedCallback () {
     this._childrenChange();
-
-    const children : Element[] = Array.from(this.children);
-    for (let i : number = 0; i < children.length; i++){
-      const child : Element = children[i]
-
-      if(i % 2 === 0){  
-        // even index = heading element
-        child.classList.add('panel-heading');
-        
-        // for A11y
-        child.id = `panel-heading-${i}`;
-        child.setAttribute('tabindex', '-1');
-        child.setAttribute('aria-controls', `panel-content-${i}`);
-      } else {
-        // odd index = content element
-        child.classList.add('panel-content');
-
-        // for A11y
-        child.id = `panel-content-${i-1}`;
-        child.setAttribute('aria-labelledby', `panel-heading-${i-1}`);
-      } 
-    }
-
-    // make the first heading focusable.
-    children[0].setAttribute('tabindex', '0');
-
-    // add EventListners
-    this.addEventListener('click', this._onClick);
-    this.addEventListener('keydown', this._onKeyDown);
-  }
-
-  // remove EventListeners
-  disconnectedCallback() {
-    this.removeEventListener('click', this._onClick);
-    this.removeEventListener('keydown', this._onKeyDown);
   }
   
   // Click event handler
-  _onClick(event) {
-    const heading : Element = event.target;
-    if (this._isHeading(heading)) {
-      this._expand(heading);
-    }
+  _onClick(event:Event) {
+    const el: Element = event.target;
+    this._expand(this._getClosestHeading(el));
   }
 
   // KeyDown event handler
   _onKeyDown(event) {
-    const currentHeading : Element = event.target;
+    const selectedEl : Element = event.target;
     
-    // if clicke event is not on heading element, ignore
-    if (!this._isHeading(currentHeading)) {
+    // if keydown event is not on heading element, ignore
+    if (!this._getClosestHeading(selectedEl)) {
       return;
     }
 
@@ -97,9 +64,7 @@ export default class MultiPanel extends HTMLElement {
 
       case 'Enter':
       case ' ':
-        if (this._isHeading(document.activeElement)) {
-            this._expand(document.activeElement);
-        }
+        this._expand(this._getClosestHeading(document.activeElement));
         break;
 
       // Any other key press is ignored and passed back to the browser.
@@ -108,7 +73,7 @@ export default class MultiPanel extends HTMLElement {
     }
 
     event.preventDefault();
-    currentHeading.setAttribute('tabindex', '-1');
+    selectedEl.setAttribute('tabindex', '-1');
     if (newHeading) {
       newHeading.setAttribute('tabindex', '0');
       newHeading.focus();
@@ -116,6 +81,11 @@ export default class MultiPanel extends HTMLElement {
   }
 
   _expand (heading : Element) {
+
+    if (heading === undefined) {
+      return;
+    }
+
     const content = heading.nextElementSibling;
     
     // heading elment should always have nextElementSibling (checked on _childrenChange)
@@ -137,47 +107,91 @@ export default class MultiPanel extends HTMLElement {
   // children of multi-panel should always in even nuber (heading/content pair)
   // if children are odd numbers, add a div at the end to prevent potential error.
   _childrenChange () {
+    console.log('hihihihihi')
     if (this.children.length % 2 !== 0) {
       console.error('detected odd number of elements inside multi-panel, please make sure you have heading/content pair')
-      this.appendChild(document.createElement('div'))
+    }
+
+    const children : Element[] = Array.from(this.children);
+    let tabindexAlreadySet : boolean = false;
+    let randomId : string;
+
+    children.forEach((child, i) => {
+      if(i % 2 === 0){
+        // even index = heading element
+
+        // if panel-heading class is already assigned to the element (= new child was added later),
+        // skip the ID/tabindex setting activities & just flag tabindexAlreadySet
+        if (!child.classList.contains('panel-heading')){
+          // for A11y
+          randomId = Math.random().toString(36).substr(2, 9);
+          child.classList.add('panel-heading');
+          child.setAttribute('tabindex', '-1');
+          child.id = `panel-heading-${randomId}`;
+          child.setAttribute('aria-controls', `panel-content-${randomId}`);
+        }  else {
+          tabindexAlreadySet = true
+        }       
+      } else {
+        // odd index = content element
+        // if panel-content class is already assigned to the element (= new child was added later), skip
+        if (!child.classList.contains('panel-content')){
+          // for A11y
+          child.classList.add('panel-content');
+          child.id = `panel-content-${randomId}`;
+          child.setAttribute('aria-labelledby', `panel-heading-${randomId}`);
+        }
+      } 
+    })
+
+    // Only if tab index was never set (= initial load) make the first heading focusable.
+    // otherwise keep tab index where already is assigned.
+    if(!tabindexAlreadySet) {
+      children[0].setAttribute('tabindex', '0');
     }
   }
 
-  // called to check if an element passed is panel heading
-  _isHeading (heading : Element) {
-    return heading.classList.contains('panel-heading');
-  }
-
-  // return list of panel heading elements
-  _allHeadings () {
-    return Array.from(this.querySelectorAll('.panel-heading'));
+  _getClosestHeading (el : Element) {
+    const closestEl = el.closest('multi-panel > *')
+    if (closestEl.classList.contains('panel-heading')) {
+      return closestEl
+    }
+    return undefined
   }
 
   // returns headding that is before currently selected one.
   _prevHeading () {
-    const headings : Element[] = this._allHeadings();
-    let newIdx : number = headings.findIndex(headings => headings === document.activeElement) - 1;
-    return headings[(newIdx + headings.length) % headings.length];
+    // activeElement would be the currently selected headding
+    // 2 elemements before that would be the previouse heading unless it is the first element.
+    if (this.firstElementChild === document.activeElement) {
+      return this.firstElementChild
+    }
+    return document.activeElement.previousElementSibling.previousElementSibling;
   }
 
   // returns headding that is after currently selected one.
   _nextHeading () {
-    const headings : Element[] = this._allHeadings();
-    let newIdx  : number = headings.findIndex(heading => heading === document.activeElement) + 1;
-    return headings[newIdx % headings.length];
+    // activeElement would be the currently selected headding
+    // 2 elemements after that would be the next heading.
+    return document.activeElement.nextElementSibling.nextElementSibling;
   }
 
   // returns first heading in multi-panel.
   _firstHeading () {
-    const headings : Element[] = this._allHeadings();
-    return headings[0];
+    // first element is always first headding
+    return this.firstElementChild;
   } 
 
   // returns last heading in multi-panel.
   _lastHeading () {
-    const headings : Element[] = this._allHeadings();
-    return headings[headings.length - 1];
-  }   
+    // if the last element is headding, return last element
+    const lastEl = this.lastElementChild;
+    if (lastEl.classList.contains('panel-heading')) {
+      return lastEl;
+    }
+    // otherwise return 2nd from the last
+    return this.lastElementChild.previousElementSibling;
+  }
 }
 
 customElements.define('multi-panel', MultiPanel); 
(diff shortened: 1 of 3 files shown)


components/MultiPanel/index.ts
```diff
@@ -1,12 +1,183 @@
...
+
+  // remove EventListeners
+  disconnectedCallback() {
+    this.removeEventListener('click', this._onClick);
+    this.removeEventListener('keydown', this._onKeyDown);
   }
+  
+  // Click event handler
+  _onClick(event) {
+    const heading : Element = event.target;
```
kosamari: I don't know what to do with types when we are doing simple event listeners...
When I do 
```
_onClick(event : Event) {
  const heading : Element = event.target
}
```
then type script yells `Type 'EventTarget | null' is not assignable to type 'Element'`

components/MultiPanel/index.ts
```diff
@@ -1,12 +1,183 @@
...
+      // Any other key press is ignored and passed back to the browser.
+      default:
+        return;
+    }
+
+    event.preventDefault();
+    currentHeading.setAttribute('tabindex', '-1');
+    if (newHeading) {
+      newHeading.setAttribute('tabindex', '0');
+      newHeading.focus();
```
kosamari: Type script is yelling `Property 'focus' does not exist on type 'Element'` here... but I can't find the way to make return of `_allHeadings` to `HTMLElement` ...

components/MultiPanel/index.ts
```diff
@@ -1,12 +1,183 @@
...
+      case 'Home':
+        newHeading = this._firstHeading();
+        break;
+
+      case 'End':
+        newHeading = this._lastHeading();
+        break;
+
+      case 'Enter':
+      case ' ':
```
kosamari: If we want to support IEs and FF before 37, then `Spacebar` might be needed here.
https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key/Key_Values

jakearchibald: We don't need to support those browsers for this project, but in terms of using this component outside this project, I'll leave it up to you.

components/MultiPanel/index.ts
```diff
@@ -1,12 +1,183 @@
...
+
+    // Watch for children changes.
+    new MutationObserver(() => this._childrenChange())
+      .observe(this, { childList: true });
   }
 
   connectedCallback () {
+    this._childrenChange();
+
+    const children : Element[] = Array.from(this.children);
```
jakearchibald: The following code assumes the internals of the multi-panel are complete once the element is connected, and never change after that. Ideally that shouldn't be assumed. It'd be better to observe this stuff in `_childrenChange` and react to elements coming and going.

components/MultiPanel/index.ts
```diff
@@ -1,12 +1,183 @@
...
+    for (let i : number = 0; i < children.length; i++){
+      const child : Element = children[i]
+
+      if(i % 2 === 0){  
+        // even index = heading element
+        child.classList.add('panel-heading');
+        
+        // for A11y
+        child.id = `panel-heading-${i}`;
```
jakearchibald: If the child already has an ID we shouldn't overwrite it.

components/MultiPanel/index.ts
```diff
@@ -1,12 +1,183 @@
...
+    // Watch for children changes.
+    new MutationObserver(() => this._childrenChange())
+      .observe(this, { childList: true });
   }
 
   connectedCallback () {
+    this._childrenChange();
+
+    const children : Element[] = Array.from(this.children);
+    for (let i : number = 0; i < children.length; i++){
```
mathiasbynens: Nit: use `forEach` instead of a hand-written loop. `forEach` is highly optimized for proper arrays in V8, so it’s likely faster, and it still gives you access to the `index` variable (unlike `for`-`of`).

components/MultiPanel/index.ts
```diff
@@ -1,12 +1,183 @@
...
+    if (!this._isHeading(currentHeading)) {
+      return;
+    }
+
+    // don’t handle modifier shortcuts used by assistive technology.    
+    if (event.altKey) {
+      return;
+    }
+    let newHeading;
+    switch (event.key) {
```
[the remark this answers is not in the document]
mathiasbynens: Nice work — I love it! So readable 👍🏻

components/MultiPanel/index.ts
```diff
@@ -1,12 +1,183 @@
...
+        child.classList.add('panel-content');
+
+        // for A11y
+        child.id = `panel-content-${i-1}`;
+        child.setAttribute('aria-labelledby', `panel-heading-${i-1}`);
+      } 
+    }
+
+    // make the first heading focusable.
+    children[0].setAttribute('tabindex', '0');
```
jakearchibald: Is https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details a better primitive here in terms of accessibility? @surma did you ever look into this?

It might be that `<details>` buggers up too much when it comes to animation.

components/MultiPanel/index.ts
```diff
@@ -1,12 +1,183 @@
...
+  // return list of panel heading elements
+  _allHeadings () {
+    return Array.from(this.querySelectorAll('.panel-heading'));
+  }
+
+  // returns headding that is before currently selected one.
+  _prevHeading () {
+    const headings : Element[] = this._allHeadings();
+    let newIdx : number = headings.findIndex(headings => headings === document.activeElement) - 1;
+    return headings[(newIdx + headings.length) % headings.length];
```
mathiasbynens: Maybe document the edge cases here? E.g. when headings contains a single element, this code just returns that.

components/MultiPanel/index.ts
mathiasbynens: (https://v8project.blogspot.com/2017/09/elements-kinds-in-v8.html if you’re interested in the background)

jakearchibald: Hah, I recommended for-of below. Which is better? Fwiw I think we should be optimising for readability over what Chrome does fastest on a particular day.

components/MultiPanel/index.ts
```diff
@@ -1,12 +1,183 @@
...
+    this.removeEventListener('click', this._onClick);
+    this.removeEventListener('keydown', this._onKeyDown);
   }
+  
+  // Click event handler
+  _onClick(event) {
+    const heading : Element = event.target;
+    if (this._isHeading(heading)) {
+      this._expand(heading);
+    }
```
jakearchibald: This breaks if the user clicks an element inside the heading. As in:

```html
<h1>Hello <em>world</em></h1>
```

Clicking `world` won't work.

Instead of `_isHeading`, maybe write something like `_getClosestHeading`? `el.closest('multi-panel > *')` will help here.

`_getClosestHeading` should return undefined/null if the user didn't click within a heading.

components/MultiPanel/index.ts
mathiasbynens: > Fwiw I think we should be optimising for readability over what Chrome does fastest on a particular day.

To be clear, I fully agree with this, and this matches our general V8 messaging.

In this case, a handwritten `for` loop expresses _implementation_, whereas `for-of` + `entries` or `forEach` would express _intent_. Expressing intent often leads to more readable code, IMHO.

`forEach` likely being faster than `for` in this case is an indirect result of that (and a nice little bonus). By expressing intent rather than implementation, you generally give the JS engine more information to work with, often leading to greater behind-the-scenes optimization potential.

components/MultiPanel/index.ts
```diff
@@ -1,12 +1,183 @@
...
+    new MutationObserver(() => this._childrenChange())
+      .observe(this, { childList: true });
   }
 
   connectedCallback () {
+    this._childrenChange();
+
+    const children : Element[] = Array.from(this.children);
+    for (let i : number = 0; i < children.length; i++){
+      const child : Element = children[i]
```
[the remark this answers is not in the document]
mathiasbynens: Between that or `forEach((child, index) => { … })`, I don’t really have a preference. I do think either of those are nicer than the hand-written `for` loop though

components/MultiPanel/index.ts
jakearchibald: You can go one of two ways with this. If you think `event.target` can only ever be an `Element`, you can do:

```ts
const heading = event.target as Element;
```

See https://www.typescriptlang.org/docs/handbook/basic-types.html#type-assertions.

However, if TypeScript is being helpful, you can make use of [type guards](https://www.typescriptlang.org/docs/handbook/advanced-types.html#type-guards-and-differentiating-types).

```ts
if (!(event.target instanceof Element)) return;
const heading: Element = event.target;
```

The above works, because TypeScript infers that `event.target` must be an `Element`.

components/MultiPanel/index.ts
```diff
@@ -1,12 +1,183 @@
...
+  }
+
+  // KeyDown event handler
+  _onKeyDown(event) {
+    const currentHeading : Element = event.target;
+    
+    // if clicke event is not on heading element, ignore
+    if (!this._isHeading(currentHeading)) {
+      return;
+    }
```
jakearchibald: Might be worth looking at [user defined type guards](https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards) here.

That way, you can tell TypeScript that, if `_isHeading` returns true, the type of `currentHeading` must be `HTMLElement` or whatever.

components/MultiPanel/index.ts
```diff
@@ -1,12 +1,183 @@
...
+
+    // don’t handle modifier shortcuts used by assistive technology.    
+    if (event.altKey) {
+      return;
+    }
```
jakearchibald: Good catch!

components/MultiPanel/index.ts
jakearchibald: Note: if you do the `_closestHeading` thing I mentioned above, `_isHeading` is still the best thing to use in this case.

components/MultiPanel/index.ts
```diff
@@ -1,12 +1,183 @@
...
+  _nextHeading () {
+    const headings : Element[] = this._allHeadings();
+    let newIdx  : number = headings.findIndex(heading => heading === document.activeElement) + 1;
+    return headings[newIdx % headings.length];
+  }
+
+  // returns first heading in multi-panel.
+  _firstHeading () {
+    const headings : Element[] = this._allHeadings();
+    return headings[0];
```
jakearchibald: I think this would be faster:

```ts
const children = this.children;
if (children.length % 1) {
  // Odd number of children. Last one must be a heading.
  return children[children.length - 1];
}
// Otherwise, the second-last child is the heading.
return children[children.length - 2];
```

components/MultiPanel/index.ts
```diff
@@ -1,12 +1,183 @@
...
+  } 
+
+  // returns last heading in multi-panel.
+  _lastHeading () {
+    const headings : Element[] = this._allHeadings();
+    return headings[headings.length - 1];
```
jakearchibald: I think this would be faster:

```ts
const children = this.children;
if (children.length % 1) {
  // Odd number of children. Last one must be a heading.
  return children[children.length - 1];
}
// Otherwise, the second-last child is the heading.
return children[children.length - 2];
```

components/MultiPanel/index.ts
jakearchibald: Ohh, and you're going to get duplication of IDs if you have two `<multi-panel>`s. Maybe generate a random ID?

components/MultiPanel/index.ts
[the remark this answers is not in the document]
jakearchibald: No preference here either.

components/MultiPanel/index.ts
developit: While we're on the subject: generally when iterating over children, it's fastest to use traversal:

```js
let child = this.firstElementChild;
while (child) {
  // stuff
  child = child.nextElementSibling;
}
```

I find this offers a nice balance of readability and performance, since it avoids the array casting entirely.

components/MultiPanel/index.ts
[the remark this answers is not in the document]
developit: Why use an Array at all?

```js
let child = this.firstElementChild;
while (child) {
  // stuff
  child = child.nextElementSibling;
}
```

- nice balance of readability and performance
- avoids the array casting entirely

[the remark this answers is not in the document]
mathiasbynens: Array operations are heavily optimized in V8 for actual arrays. The one-off cost of casting to an array is usually worth the later performance benefits when operating on the array (as opposed to an array-like object). (shameless https://www.youtube.com/watch?v=m9cTaYI95Zc plug in case anyone wants more info)

TL;DR Don’t think of the `Array.from()` as overhead.

kosamari:
Status : 
Current code has all refraction of the previous code review (minus `<details>` consideration).

This component currently assumes there is always a set of `heading` and  `content`.
and expands corresponding content via `nextElementSibling`.

If developer using this component does something like below (per @jakearchibald's review) and user click on heading before content is created, it should be fine (simply no`nextElementSibling`)
```
multiPanel.append(parse(`<h1>Hello!</h1>`));
const data = await fetch(content).then(r => r.json());
multiPanel.append(parse(data.html));
```
 
however, if developer inserts element in async way like above in the middle of multi panel children... it will mess up which panel to be 'expanded'.

Not sure expecting dev user to "always append 2 elements together" is a reasonable expectation (code will be simpler !) or we should bullet proof this with unique ID approach and control element gets expanded based on ID.

note: currently unique id is assigned just for A11y but not for control.

components/MultiPanel/index.ts
```diff
@@ -1,12 +1,197 @@
...
+  _childrenChange () {
+    if (this.children.length % 2 !== 0) {
+      console.error('detected odd number of elements inside multi-panel, please make sure you have heading/content pair')
+    }
+
+    const children : Element[] = Array.from(this.children);
+    let tabindexAlreadySet : boolean = false;
+    let randomId : string;
+
+    children.forEach((child, i) => {
```
jakearchibald: I think this would be more readable as:

```js
for (let i = 0; i < children.length; i += 2) {
  const heading = children[i];
  const content = children[i+1];
  const randomId = Math.random().toString(36).substr(2, 9);

  // …
}
```

As you don't need to do the odd/even check, and `randomId` gets better scoping.

components/MultiPanel/index.ts
```diff
@@ -1,12 +1,197 @@
...
+      if(i % 2 === 0){
+        // even index = heading element
+
+        // if panel-heading class is already assigned to the element (= new child was added later),
```
jakearchibald: Earlier?

components/MultiPanel/index.ts
```diff
@@ -1,12 +1,197 @@
...
+        // skip the ID/tabindex setting activities & just flag tabindexAlreadySet
+        if (!child.classList.contains('panel-heading')){
```
jakearchibald: If there's an if statement with a short part & a long part, I think it's more readable to handle the short part first. In some cases it means you can avoid the `else`. In this case:

```js
if (child.classList.contains('panel-heading')) {
  tabIndexAlreadySet = true;
  return;
}

child.classList.add('panel-heading');
// …
```

The `return` will need to be `continue` if you change this to a for-loop.

components/MultiPanel/index.ts
```diff
@@ -1,12 +1,197 @@
...
+    const children : Element[] = Array.from(this.children);
+    let tabindexAlreadySet : boolean = false;
+    let randomId : string;
+
+    children.forEach((child, i) => {
+      if(i % 2 === 0){
+        // even index = heading element
+
+        // if panel-heading class is already assigned to the element (= new child was added later),
+        // skip the ID/tabindex setting activities & just flag tabindexAlreadySet
```
jakearchibald: I think this strategy has problems if you replace just a heading, or just the content. Or, if you swap some headings around but not the content.

components/MultiPanel/index.ts
kosamari: handled by adding `case 'Spacebar':`

components/MultiPanel/index.ts
kosamari: handled this in suggested way.

kosamari:
closing this in lieu of #95

PR #27, opened 2018-05-04
Merging pinch-zoom

jakearchibald:


src/components/output/custom-els/PinchZoom/missing-types.d.ts
```diff
@@ -0,0 +1,16 @@
...
+interface Window {
+  PointerEvent: typeof PointerEvent;
+  Touch: typeof Touch;
+}
+
+declare namespace JSX {
+  interface IntrinsicElements {
+    "pinch-zoom": any
+  }
+}
```
jakearchibald: This seems to be the right thing to do https://www.typescriptlang.org/docs/handbook/jsx.html#intrinsic-elements

jakearchibald:
This is all as previously reviewed, except for the JSX intrinsic type.

PR #29, opened 2018-05-15
Using @bind

jakearchibald:
I should have done this before merging. Please forgive my trespasses.

### DIFF
--- a/src/lib/PointerTracker/index.ts
+++ b/src/lib/PointerTracker/index.ts
@@ -1,3 +1,4 @@
+import { bind } from '../util';
 const enum Button { Left }
 
 export class Pointer {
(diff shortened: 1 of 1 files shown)


src/lib/PointerTracker/index.ts
```diff
@@ -1,3 +1,4 @@
+import { bind } from '../util';
```
surma: Just for me: We are doing extension-less imports because webpack, right?

jakearchibald: A vscode extension I used added this import for me 😄.

Not wedded to extension-less imports. Any feels here @developit @kosamari?

surma: FWIW I tend to use `.js` extension. TypeScript automatically prefers `.ts` files with the same name if they exist. But no strong feels on my side.

developit: TBH it doesn't matter to me. It's extremely uncommon to use extensions in Webpack, though.

mathiasbynens: Drive-by opinion: explicit > implicit

surma: >  It's extremely uncommon to use extensions in Webpack

Probably the strongest argument so far, considering one of our goals was to meet developers where they are currently at.

PR #34, opened 2018-05-16
Integrating two-up

jakearchibald:
Ready for review!

### DIFF
--- a/src/components/output/index.tsx
+++ b/src/components/output/index.tsx
@@ -1,6 +1,10 @@
 import { h, Component } from 'preact';
+import PinchZoom from './custom-els/PinchZoom';
 import './custom-els/PinchZoom';
+import './custom-els/TwoUp';
 import * as style from './style.scss';
+import { bind } from '../../lib/util';
+import { twoUpHandle } from './custom-els/TwoUp/styles.css';
 
 type Props = {
   img: ImageBitmap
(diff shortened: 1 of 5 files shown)


jakearchibald:
Pixelated bugs:

Chrome: https://bugs.chromium.org/p/chromium/issues/detail?id=843615.
Safari: https://bugs.webkit.org/show_bug.cgi?id=185683.

I need to add `image-rendering: optimizespeed;` for Firefox.
I need to fix my test to work in Edge 😢.

src/components/output/index.tsx
```diff
@@ -10,34 +14,88 @@ type State = {};
...
-
-  updateCanvas(img: ImageBitmap) {
-    if (!this.canvas) return;
-    const ctx = this.canvas.getContext('2d');
-    if (!ctx) return;
-    ctx.clearRect(0, 0, this.canvas.width, this.canvas.height);
-    ctx.drawImage(img, 0, 0);
+  updateCanvases(img: ImageBitmap) {
+    for (const [i, canvas] of [this.canvasLeft, this.canvasRight].entries()) {
+      if (!canvas) return;
```
surma: `continue`? (If there is `canvasRight`, but not `canvasLeft`?)

src/components/output/index.tsx
```diff
@@ -10,34 +14,88 @@ type State = {};
 
 export default class App extends Component<Props, State> {
   state: State = {};
-  canvas?: HTMLCanvasElement;
+  canvasLeft?: HTMLCanvasElement;
+  canvasRight?: HTMLCanvasElement;
+  pinchZoomLeft?: PinchZoom;
```
surma: Considering we render those in the JSX, can they ever be null?

src/components/output/index.tsx
```diff
@@ -10,34 +14,88 @@ type State = {};
...
+    if (!this.pinchZoomLeft) return;
+    // If the event is on the handle of the two-up, let it through.
+    if (targetEl.closest('.' + twoUpHandle)) return;
+    // If we've already retargeted this event, let it through.
+    if (this.retargetedEvents.has(event)) return;
+    // Stop the event in its tracks.
+    event.stopImmediatePropagation();
+    event.preventDefault();
+    // Clone the event & dispatch
+    const clonedEvent = new (event.constructor as typeof Event)(event.type, event);
```
surma: Surprised you need to force the type of `event.constructor`, but I’m guessing you did try without.

src/components/output/index.tsx
```diff
@@ -10,34 +14,88 @@ type State = {};
...
+        <two-up
+          // Event redirecting. See onRetargetableEvent.
+          onTouchStartCapture={this.onRetargetableEvent}
+          onTouchEndCapture={this.onRetargetableEvent}
+          onTouchMoveCapture={this.onRetargetableEvent}
+          onPointerDownCapture={this.onRetargetableEvent}
+          onMouseDownCapture={this.onRetargetableEvent}
+          onWheelCapture={this.onRetargetableEvent}
+        >
+          <pinch-zoom onChange={this.onPinchZoomLeftChange} ref={p => this.pinchZoomLeft = p as PinchZoom}>
```
surma: Can’t you do `p: PinchZoom => this.pinchZoomLeft = p`?

src/components/output/index.tsx
jakearchibald: They're nullable in terms of TypeScript, since they're defined *after* the constructor.

src/components/output/index.tsx
jakearchibald: I should add a comment linking to https://github.com/Microsoft/TypeScript/issues/3841.

src/components/output/index.tsx
jakearchibald: Nah, since ref provides an `Element`. Typing the arg is different to casting, and it's casting that I need here.

Casting seems fair here, since it won't ever be something other than a canvas (famous last words).

src/components/output/index.tsx
jakearchibald: I guess this is one of those "this should never happen" cases.

Should we throw errors in these cases?

surma: If it’s never supposed to happen, I’d say throw. Silent return can be frustrating

jakearchibald: Agreed. I'll fix other places where I've done this.

developit:
_(aside: ugh, I made some comments here but never submitted the review)_

jakearchibald:
@developit will it not let you submit it now?

PR #35, opened 2018-05-17
Load mozjpeg codec and encode image

surma:
Here we go! I am compressing an image with MozJPEG (with hardcoded quality set to 2%) and it works. Huzzah 🎉 

I haven’t done WebP as I wanted to make sure the stuff I came up with are ok with y’all. So I’ll add that in a separate PR if you don’t mind once this lands.


package.json
```diff
@@ -4,6 +4,8 @@
   "version": "0.0.0",
   "license": "apache-2.0",
   "scripts": {
+    "build:mozjpeg_enc": "cd codecs/mozjpeg_enc && npm run build",
+    "build:codecs": "npm run build:mozjpeg_enc",
```
surma: @developit Currently I just have separate npm scripts to build all codecs and assume that these tasks have been run before building the app. WDYT? Do you have a better idea?

src/components/app/index.tsx
```diff
@@ -23,13 +25,31 @@ export default class App extends Component<Props, State> {
...
   @bind
   async onFileChange(event: Event) {
     const fileInput = event.target as HTMLInputElement;
     if (!fileInput.files || !fileInput.files[0]) return;
     // TODO: handle decode error
-    const img = await createImageBitmap(fileInput.files[0]);
-    this.setState({ img });
+    const bitmap = await createImageBitmap(fileInput.files[0]);
+    const data = await this.getImageData(bitmap);
+    const encoder = new MozJpegEncoder();
```
surma: The way it’s currently written, we should just be able to load these encoders in a worker and invoke `encode` via Comlink.

codecs/mozjpeg_enc/mozjpeg_enc.d.ts
```diff
@@ -0,0 +1 @@
+export default function(opts: {}): EmscriptenWasm.Module;
```
jakearchibald: Don't we need to give types to the options?

src/components/app/index.tsx
```diff
@@ -23,13 +25,31 @@ export default class App extends Component<Props, State> {
...
   }
 
+  private async getImageData(bitmap: ImageBitmap): Promise<ImageData> {
+    // Make canvas same size as image
+    const canvas = document.createElement('canvas');
+    [canvas.width, canvas.height] = [bitmap.width, bitmap.height];
+    // Draw image onto canvas
+    const ctx = canvas.getContext('2d');
+    if (!ctx) {
+      throw new Error("Could not create canvas contex");
```
jakearchibald: "context"

src/components/app/index.tsx
```diff
@@ -23,13 +25,31 @@ export default class App extends Component<Props, State> {
     }
   }
 
+  private async getImageData(bitmap: ImageBitmap): Promise<ImageData> {
+    // Make canvas same size as image
+    const canvas = document.createElement('canvas');
+    [canvas.width, canvas.height] = [bitmap.width, bitmap.height];
```
jakearchibald: I still think this technique is smarter than it is good, and it'd be better as two lines. Happy to be overridden by @developit or @kosamari though.

src/components/app/index.tsx
```diff
@@ -23,13 +25,31 @@ export default class App extends Component<Props, State> {
     }
   }
 
+  private async getImageData(bitmap: ImageBitmap): Promise<ImageData> {
```
jakearchibald: This doesn't read/write any class state, so it'd be better as a function.

src/lib/codec-wrappers/codec.ts
```diff
@@ -0,0 +1,7 @@
+export interface Encoder {
+  encode(data: ImageData): Promise<ArrayBuffer | SharedArrayBuffer>;
```
jakearchibald: What causes the return type to be `ArrayBuffer` *or* `SharedArrayBuffer`?

jakearchibald: Also, if this going to be used by anything? Won't they all have different options?

src/lib/codec-wrappers/mozjpeg-enc.ts
```diff
@@ -0,0 +1,72 @@
...
+        noInitialRun: false,
+        locateFile(url: string): string {
+          // Redirect the request for the wasm binary to whatever webpack gave us.
+          if(url.endsWith('.wasm')) {
+            return wasmBinaryUrl;
+          }
+          return url;
+        },
+        onRuntimeInitialized() {
+          // This is a bug I discovered. To be filed.
```
jakearchibald: More info on this.

src/components/app/index.tsx
surma: Nah, let’s keep it explicit. I don’t mind. I copy-pasted this from the `example.html` that I copied from my article :3

src/components/app/index.tsx
surma: Right, good call. I should probably start a `imageutils` file or something. We are gonna write a lot more of these I reckon.

src/lib/codec-wrappers/codec.ts
surma: I’ll re-check `ArrayBuffer` vs `SharedArrayBuffer`, but I think TS was complaining about _something_ (and was, strictly speaking, correct).

src/components/app/index.tsx
jakearchibald: Also, we should do this on the worker side if we can (will be fun to play with offscreen canvas)

src/lib/codec-wrappers/codec.ts
jakearchibald: > I did this because I thought it would be nice to have a consistent encode() call and rather have a stateful encoder with individual calls to set options.

Yeah, fair enough!

src/components/app/index.tsx
surma: `OffscreenCanvas` isn't stable anywhere tho.

surma: ... could we built an "Offthread" canvas  with an iframe? Cc @developit

jakearchibald: I guess we should see how much jank it causes us first.

The iframe idea is interesting! OOPIFs are a thing now, so it could work. I wonder if transferrables lose some of their benefit when jumping across processes.

surma:
@jakearchibald PTAL :)

@developit still would really appreciate your input on the build system here.

src/lib/codec-wrappers/codec.ts
surma: Yeah, `SharedArrayBuffer` was a left-over. I just removed it :)

jakearchibald:
I'm happy with this, but since it involves build I'd like @developit to review & sign-off.

package.json
developit: TBH that seems fine to me, unless we think the codecs will be changed as frequently as app code. Otherwise it seems nice to keep those things separate and keep the builds fast.  Are the results from these builds committed?

surma: Build-speed was the main argument for me as well. And yes, build-artifacts are not commited.

jakearchibald:
Approved, since @developit didn't disapprove

src/lib/util.ts
```diff
@@ -24,3 +24,22 @@ export function bind(target: any, propertyKey: string, descriptor: PropertyDescr
     }
   };
 }
+
+/**
+ * Turns a given `ImageBitmap` into `ImageData`.
+ */
+export async function bitmapToImageData(bitmap: ImageBitmap): Promise<ImageData> {
+  // Make canvas same size as image
+  // TODO: Move this off-thread if possible with `OffscreenCanvas` or iFrames?
```
developit: We could just cache this for now so it only runs when the source image changes. OffscreenCanvas is still behind a flag and iframe would have to be served from a different origin to have any benefit :(

PR #38, opened 2018-05-23
Fix codec integration

developit:
With these fixes in place I have the build running locally on a chromebook.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "scripts": {
     "build:mozjpeg_enc": "cd codecs/mozjpeg_enc && npm run build",
     "build:codecs": "npm run build:mozjpeg_enc",
-    "start": "webpack serve --hot",
+    "start": "webpack serve --host 0.0.0.0 --hot",
     "build": "webpack -p",
     "lint": "eslint src"
   },
@@ -85,6 +85,7 @@
     "classnames": "^2.2.5",
     "material-components-web": "^0.32.0",
     "material-radial-progress": "git+https://gist.github.com/02134901c77c5309924bfcf8b4435ebe.git",
+    "mozjpeg": "^6.0.0",
     "preact": "^8.2.7",
     "preact-i18n": "^1.2.0",
     "preact-material-components": "^1.3.7",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -85,6 +85,7 @@
     "classnames": "^2.2.5",
     "material-components-web": "^0.32.0",
     "material-radial-progress": "git+https://gist.github.com/02134901c77c5309924bfcf8b4435ebe.git",
+    "mozjpeg": "^6.0.0",
```
surma: Why is this needed? And if it is, would it make more sense to have it in the `codecs/mozjpeg_enc/package.json`?

src/lib/codec-wrappers/codec.ts
```diff
@@ -1,5 +1,5 @@
 export interface Encoder {
-  encode(data: ImageData): Promise<ArrayBuffer>;
+  encode(data: ImageData): Promise<ArrayBuffer | SharedArrayBuffer>;
```
surma: Iiiinteresting. I removed the `SharedArrayBuffer` and it still compiled. 

I just cleaned the entire repo from build artifacts and now I am seeing the same error. I guess we have some caching in the build system somewhere.

surma: On second thought: I’d actually prefer to enforce `ArrayBuffer` as wasm cannot run on `SharedArrayBuffer`, so it’s not really a possibility anyways.

@jakearchibald: I now remember why this was a thing before: The wasm/Emscripten modules export `HEAP8`, `HEAP16` etc as views onto the memory, which are `Int8Array`, `Int16Array` etc. So `HEAP8.buffer` has type `ArrayBuffer | SharedArrayBuffer`. 

I’ll push a commit to this branch.

surma:
I’m gonna go ahead and merge this. Please still take a quick look at that question of mine, @developit  :D

package.json
surma: Gotcha. I’ll remove it.

developit:
Fixed on master.

package.json
surma: ... or you will :D

Issue #30, opened 2018-05-15
Create config component


jakearchibald:
It'll need:

* A drop-down select, to select the codec.
* Toggles.
* Range sliders (eg "quality").

This will be a good point for us to review and test the material design implementation, and decide if the off-the-shelf solution is good enough.

No need hooking the form up to anything until https://github.com/GoogleChromeLabs/squoosh/issues/25 is complete, then we can look at how to define configs for each codec.

jakearchibald:
Since @surma has started merging in the codecs, this can now involve looking at the actual options the codec provides, and creating a realistic form.

surma:
Should we figure out some kind of self-description per codec? 

Like `getOptions()` and that returns some sort of JSON that describes the available options from which when then generate the options panel?

developit:
@surma either that or just have components that implement each codec's settings so we can customize as-needed. Not sure about you but every time I've done form generation from a schema, I've always ended up regretting it, haha.

jakearchibald:
I've also experienced the schema-form-generation pain. However, the form should be 'close' to the codec, so we remember to update one if we update the other.

jakearchibald:
I think this is done in terms of its own issue. We'll create additional issues per codec.

PR #39, opened 2018-05-25
Options UI

developit:
Implements basic support for options (#30), configurable for both the left and right panels. One of the modes is "original", which is a pass-through that displays the source image.


src/components/app/index.tsx
```diff
@@ -2,17 +2,58 @@ import { h, Component } from 'preact';
...
-import {MozJpegEncoder} from '../../lib/codec-wrappers/mozjpeg-enc';
+// import { Encoder } from '../../lib/codec-wrappers/codec';
+import { MozJpegEncoder } from '../../lib/codec-wrappers/mozjpeg-enc';
+
+export type ImageType = 'original' | 'jpeg';
+
+export type CodecOptions = any;
+
+type Encoders = {
+  [type: string]: any  // @todo Encoder
```
developit: This is a "TypeScript is frustrating" moment. For some reason, MozJpegEncoder isn't considered to have the same interface as Encoder?  It's likely just me doing something stupid.

surma: `MozJpegEncoder` doesn’t implement `Encoder`, an instance of `MozJpegEncoder` does 😄 

You could do `new () => Encoder` instead of any here.

developit:
Aside from the design obviously being poor, I'd like to collocate the options component for each codec with the codec itself.  I'm thinking that will be best if I move them into `src/codecs/[codec]/{index,options}.ts`

src/components/app/style.scss
```diff
@@ -1,3 +1,58 @@
-.app h1 {
-  color: green;
+.app {
+  position: absolute;
+  left: 0;
+  top: 0;
+  width: 100%;
+  height: 100%;
+  overflow: hidden;
+  contain: strict;
```
surma: Why?

src/components/app/style.scss
```diff
@@ -1,3 +1,58 @@
-.app h1 {
-  color: green;
+.app {
+  position: absolute;
+  left: 0;
+  top: 0;
+  width: 100%;
+  height: 100%;
+  overflow: hidden;
```
surma: I don’t feel like we should go `overflow: hidden` without a good reason. For mobile we definitely want to be using the root scroller and this would make that very hard.

src/components/app/style.scss
```diff
@@ -1,3 +1,58 @@
-.app h1 {
-  color: green;
+.app {
+  position: absolute;
```
surma: I feel like this file could be condensed down a lot with flexbox and grid.

src/components/app/index.tsx
```diff
@@ -2,17 +2,58 @@ import { h, Component } from 'preact';
...
-  state: State = {};
+  state: State = {
+    loading: false,
+    leftType: 'original',
+    rightType: 'jpeg',
+    leftOptions: {},
+    rightOptions: {}
+  };
+
+  optionsUpdateTimer?: NodeJS.Timer | number | null;
```
surma: - Don’t think you need `null` if you are using `?`
- Why `NodeJS.Timer`? 
- What does this prop stand for?

src/components/app/index.tsx
```diff
@@ -2,17 +2,58 @@ import { h, Component } from 'preact';
...
+
+  compressCounter = 0;
+
+  retries = 0;
```
surma: I don’t think we should be doing retries. Either it’s a bug in the way I wrote or invoke my C code, or it’s a Chrome bug. If it’s a Chrome bug and it can’t be fixed, only then should we consider writing a retry strategy, and it should be modular instead of being part of the app container.

src/components/app/index.tsx
```diff
@@ -2,17 +2,58 @@ import { h, Component } from 'preact';
...
+export type ImageType = 'original' | 'jpeg';
+
+export type CodecOptions = any;
+
+type Encoders = {
+  [type: string]: new () => Encoder
+};
+
+const EncoderNames = {
+  original: 'Original Image',
```
surma: Make this a function that just always returns a string to have a consistent type for this dictionary?

src/components/app/index.tsx
```diff
@@ -25,33 +66,131 @@ export default class App extends Component<Props, State> {
...
+  getEncoderName(type: ImageType, options: CodecOptions) {
+    const name = EncoderNames[type];
+    if (typeof name === 'function') return name(options);
+    return name;
+  }
+
+  @bind
+  setLeftType(leftType: ImageType) {
+    this.setState({ leftType, leftOptions: {} });
+    this.optionsUpdated();
```
surma: Considering we see this pattern a lot, should we make a second decorator where a function will just return the partial update passed to `setState()` and that calls `optionsUpdated()` implicitly?

Not sure if making this DRY is worth it, just a thought. @kosamari @jakearchibald ?

src/components/app/index.tsx
```diff
@@ -25,33 +66,131 @@ export default class App extends Component<Props, State> {
...
+    const sourceData = await bitmapToImageData(sourceImg);
+    this.setState({ sourceImg, sourceData, loading: false });
+    this.updateImages();
   }
 
-  render({ }: Props, { img }: State) {
+  async updateImages() {
+    const { sourceData, leftType, rightType, leftOptions, rightOptions } = this.state;
+    if (!sourceData) return;
+    const id = ++this.compressCounter;
```
surma: What is this `id` for?

src/components/app/index.tsx
```diff
@@ -25,33 +66,131 @@ export default class App extends Component<Props, State> {
...
+    const [leftImg, rightImg] = await Promise.all([
+      this.updateCompressedImage(sourceData, leftType, leftOptions),
+      this.updateCompressedImage(sourceData, rightType, rightOptions)
+    ]);
+    if (this.compressCounter !== id) return;
+    this.setState({ leftImg, rightImg, loading: false });
+  }
+
+  async updateCompressedImage(sourceData: ImageData, type: ImageType, options: CodecOptions, retries= 0) {
+    if (type === 'original') {
```
surma: Just wondering: Worth creating a pseudo-encode in the `AllEncoders` dictionary that’s just a pass-through?

src/components/app/index.tsx
```diff
@@ -25,33 +66,131 @@ export default class App extends Component<Props, State> {
...
+    // }
+    const encoder = new AllEncoders[type]();
+    try {
+      const compressedData = await encoder.encode(sourceData, options);
+      const blob = new Blob([compressedData], { type: 'image/jpeg' });
+      return await createImageBitmap(blob);
+    } catch (err) {
+      // console.log(`failed to encode ${type} (retries: ${retries}): ${err}`);
+      if (retries < 5) {
+        await this.updateCompressedImage(sourceData, type, options, retries + 1);
```
surma: Why don’t you just loop here?

src/components/app/index.tsx
```diff
@@ -25,33 +66,131 @@ export default class App extends Component<Props, State> {
...
+    //   console.error(`Unknown encoder: ${type}`);
+    //   return;
+    // }
+    const encoder = new AllEncoders[type]();
+    try {
+      const compressedData = await encoder.encode(sourceData, options);
+      const blob = new Blob([compressedData], { type: 'image/jpeg' });
+      return await createImageBitmap(blob);
+    } catch (err) {
+      // console.log(`failed to encode ${type} (retries: ${retries}): ${err}`);
```
surma: Remove dead code

src/components/app/style.scss
```diff
@@ -1,3 +1,58 @@
...
+  .rightOptions { right: 10px; }
+}
+
+.welcome {
+  position: absolute;
+  display: inline-block;
+  left: 50%;
+  top: 50%;
+  padding: 20px;
+  transform: translate(-50%, -50%);
```
surma: We don’t do this kind of vertical centering anymore ;)

src/components/options/index.tsx
```diff
@@ -0,0 +1,102 @@
...
+  original: () => null
+};
+
+type ChildProps = {
+  options: CodecOptions,
+  updateOption(e: Event): void,
+  setOption(key: string, value: any): void
+};
+
+CodecConfigurations.jpeg = ({ options, updateOption }: ChildProps) => (
```
surma: Yeah, would be nicer to move this to a file next to the codec wrapper (or even into the codec wrapper file? Your call).

Also I think we should call the codec for what it is: MozJpeg

src/components/options/index.tsx
```diff
@@ -0,0 +1,102 @@
...
+  };
+
+  componentWillReceiveProps({ options }: Props) {
+    if (options !== this.props.options) {
+      this.setState({ options });
+    }
+  }
+
+  @bind
+    setType(e: Event) {
```
surma: formatting seems wonky here

src/components/options/index.tsx
```diff
@@ -0,0 +1,102 @@
...
+    const el = e.currentTarget as HTMLSelectElement;
+    this.props.onTypeChange(el.value);
+  }
+
+  @bind
+    updateOption(e: Event) {
+    const el = e.currentTarget as HTMLInputElement;
+    this.setOption(el.name, /(rad|box)/i.test(el.type) ? el.checked : el.value);
```
surma: We can afford a switch-case here.

src/components/options/style.scss
```diff
@@ -0,0 +1,34 @@
+.options {
+    width: 180px;
+    padding: 10px;
+    background: rgba(50,50,50,0.8);
+    border: 1px solid #222;
+    box-shadow: inset 0 0 1px #fff, 0 0 1px #fff;
+    border-radius: 3px;
+    color: #eee;
+    overflow: auto;
```
surma: I think we should try to avoid scrolling elements as much as possible.

src/components/output/index.tsx
```diff
@@ -20,26 +21,29 @@ export default class App extends Component<Props, State> {
...
-      if (i === 1) {
-        // This is temporary, to show the images are different
-        ctx.filter = 'hue-rotate(180deg)';
-      }
-      ctx.clearRect(0, 0, canvas.width, canvas.height);
-      ctx.drawImage(img, 0, 0);
-    }
+  updateCanvas(canvas: HTMLCanvasElement, img?: ImageBitmap) {
+    let ctx = canvas.getContext('2d');
+    if (!ctx) return;
```
surma: Silent errors are evil, please handle that the same way as the old code.

src/components/output/style.scss
```diff
@@ -1,3 +1,21 @@
+@mixin fill {
+  position: absolute;
+  left: 0;
+  top: 0;
+  width: 100%;
+  height: 100%;
+  overflow: hidden;
+  contain: strict;
```
surma: Same as the other file

src/lib/codec-wrappers/codec.ts
```diff
@@ -1,5 +1,5 @@
 export interface Encoder {
-  encode(data: ImageData): Promise<ArrayBuffer>;
+  encode(data: ImageData, options: any): Promise<ArrayBuffer>;
```
surma: Jake and I discussed in the original PR as well and I said I’d like to keep the `encode()` function consistent across all codecs and have individual methods for each option.

src/lib/codec-wrappers/mozjpeg-enc.ts
```diff
@@ -57,13 +61,13 @@ export class MozJpegEncoder implements Encoder {
...
 
-  async encode(data: ImageData): Promise<ArrayBuffer> {
+  async encode(data: ImageData, options: EncodeOptions): Promise<ArrayBuffer> {
     const m = await this.emscriptenModule;
     const api = await this.api;
 
     const p = api.create_buffer(data.width, data.height);
     m.HEAP8.set(data.data, p);
-    api.encode(p, data.width, data.height, 2);
+    api.encode(p, data.width, data.height, options.quality!=null ? options.quality : 7);
```
surma: We should throw (or something) if a required option is not given. The UI can define a default, the module itself shouldn’t imo.

src/components/app/index.tsx
jakearchibald: I think it's too early for that. We'll be splitting this up so that a change to the left doesn't result in a reencode on the right.

However, @developit, this looks like a case where we want to look at state before/after, and set additional state as a result. As in, "If left's options are different, then update `leftImg`". Is this a common pattern?

We could either split into `leftOptionsUpdated` and `rightOptionsUpdate`, or give `optionsUpdated` the ability to know what's changed.

jakearchibald: (Either way, I don't think a decorator is saving us much, it's just moving the call up a couple of lines up)

src/components/app/index.tsx
```diff
@@ -25,33 +66,131 @@ export default class App extends Component<Props, State> {
...
+  setRightOptions(rightOptions: any) {
+    this.setState({ rightOptions });
+    this.optionsUpdated();
+  }
+
+  optionsUpdated() {
+    if (!this.optionsUpdateTimer) {
+      this.optionsUpdateTimer = setTimeout(this.commitOptionsUpdated, 500);
+    }
+  }
```
jakearchibald: I'd like us to have a think about the best kind of debouncing to use in these cases, but not needed for this PR.

src/components/app/index.tsx
jakearchibald: Let's just ditch the retries bit.

src/components/app/index.tsx
```diff
@@ -25,33 +66,131 @@ export default class App extends Component<Props, State> {
...
+
+  render({ }: Props, { loading, leftType, leftOptions, rightType, rightOptions, leftImg, rightImg }: State) {
     return (
       <div id="app" class={style.app}>
-        {img ?
-          <Output img={img} />
-          :
-          <div>
+        {leftImg && rightImg ? (
```
jakearchibald: Eventually we'll go to the output view even if only one of these are ready. Not needed for this PR.

src/components/app/style.scss
jakearchibald: Just to write down what we talked about in the office: I think we should use per-element scrollers for our compression option elements. If iOS gives us troubles, we'll work around them (potentially by reducing niceness on that platform). Basically, I'm sick of scrolling being broken on iOS.

src/components/options/index.tsx
jakearchibald: Agree on both counts.

src/components/options/index.tsx
jakearchibald: Same goes for other methods in this class

src/components/options/index.tsx
```diff
@@ -0,0 +1,102 @@
+import { h, Component } from 'preact';
+import * as style from './style.scss';
+import { bind } from '../../lib/util';
+import { ImageType, CodecOptions } from '../app';
+
+const CodecConfigurations: {
```
jakearchibald: `codecConfigurations` or whatever we agree in the earlier comment.

src/components/options/index.tsx
jakearchibald: Or an if.

Also, I don't think we have anything that matches rad/box yet, right?

src/components/app/style.scss
```diff
@@ -1,3 +1,58 @@
-.app h1 {
-  color: green;
+.app {
```
jakearchibald: Slap a comment like "Temp styles. These will be reviewed before going live" at the top & I'm happy for this to merge.

src/components/options/index.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+  @bind
+    setOption(key: string, value?: any) {
+    const options = Object.assign({}, this.state.options);
+    options[key] = value;
+    this.setState({ options });
+    this.props.onOptionsChange(options);
+  }
+
+  render({ class: c, name, type }: Props, { options }: State) {
```
jakearchibald: Rename `c` to `className`? We've got minifiers for the rest.

src/components/options/index.tsx
```diff
@@ -0,0 +1,102 @@
...
+                    </select>
+                </label>
+                {CodecOptionComponent && (
+                    <CodecOptionComponent
+                        options={options}
+                        setOption={this.setOption}
+                        updateOption={this.updateOption}
+                    />
+                )}
+                {/* <pre>{JSON.stringify(options,null,2)}</pre> */}
```
jakearchibald: Remove dead code

src/components/options/index.tsx
```diff
@@ -0,0 +1,102 @@
...
+    options[key] = value;
+    this.setState({ options });
+    this.props.onOptionsChange(options);
+  }
+
+  render({ class: c, name, type }: Props, { options }: State) {
+    const CodecOptionComponent = CodecConfigurations[type];
+
+    return (
+            <div class={`${style.options}${c ? (' ' + c) : ''}`}>
```
jakearchibald: Indenting's gone weird in this section

src/components/options/index.tsx
```diff
@@ -0,0 +1,102 @@
...
+                name="quality"
+                type="range"
+                min="1"
+                max="100"
+                step="1"
+                value={options.quality}
+                onInput={updateOption}
+            />
+        </label>
+    </div>
```
jakearchibald: Inconsistent indenting above

src/components/options/style.scss
```diff
@@ -0,0 +1,34 @@
+.options {
```
jakearchibald: As before, slap a "this is temporary" comment at the top and I'm happy for this to merge.

src/components/output/index.tsx
```diff
@@ -20,26 +21,29 @@ export default class App extends Component<Props, State> {
...
-      const ctx = canvas.getContext('2d');
-      if (!ctx) throw Error('Expected 2d canvas context');
-      if (i === 1) {
-        // This is temporary, to show the images are different
-        ctx.filter = 'hue-rotate(180deg)';
-      }
-      ctx.clearRect(0, 0, canvas.width, canvas.height);
-      ctx.drawImage(img, 0, 0);
-    }
+  updateCanvas(canvas: HTMLCanvasElement, img?: ImageBitmap) {
```
jakearchibald: Doesn't look like this needs to be a member of the class (no state usage)

src/lib/codec-wrappers/codec.ts
jakearchibald: I'm happy for it to land as-is, and we can have a discussion when we start to integrate options.

src/style/index.scss
```diff
@@ -5,6 +5,8 @@ html, body {
   width: 100%;
   padding: 0;
   margin: 0;
+  font: 14px/1.3 Roboto,'Helvetica Neue',arial,helvetica,sans-serif;
```
jakearchibald: Let's move these changes into their own rule and put the usual disclaimer there.

src/components/options/style.scss
surma: nvm this as per @jakearchibald’s comment

src/components/app/index.tsx
```diff
@@ -2,17 +2,58 @@ import { h, Component } from 'preact';
...
+type Encoders = {
+  [type: string]: new () => Encoder
+};
+
+const EncoderNames = {
+  original: 'Original Image',
+  jpeg: (options: CodecOptions) => `JPEG ${options.quality || ''}`
+};
+
+const AllEncoders: Encoders = {
```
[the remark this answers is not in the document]
surma: That was the motivation for not having an options parameter for the `encode()` call, so that all encoders could be handled transparently.

src/components/app/style.scss
developit: also trying to use fixed for the fixed portions of the UI would be a nightmare with auto-hiding address bar.

src/components/app/index.tsx
developit: It only works with this combination. I believe it's because we're combining the Node and web global typescript defs - setTimeout() returns a NodeJS.Timer, but cancelTimeout expects an int.

src/components/app/index.tsx
developit: I'm fine with that.

src/components/app/index.tsx
developit: I had a few different implementations of this, including the abstracted "setter with optionsUpdated", a componentDidUpdate() hook with a comparison, etc.  TBH I don't care which we go with, these were just the most explicit.

src/components/app/index.tsx
developit: I'd be up for abstracting this (along with the retry strategy if we keep one) out.

src/components/app/index.tsx
developit: Ah yeah that's cleaner.

src/components/options/index.tsx
developit: yup, I'll collocate these. I'm going to hoist that directory though.

src/components/options/index.tsx
developit: Yeah, I just realized I committed this from pixelbook and hadn't installed tslint or eslint in my editor.

src/components/options/index.tsx
developit: Yeah, just there's no real way to determine which should be used for a given element.

src/lib/codec-wrappers/codec.ts
developit: I wasn't sure what we wanted the lifecycle of these encoder classes to be. If we reuse one instance for multiple encodes, methods are pain.  Also given the fact that state is managed elsewhere, duplicating it here just causes extra work.

src/lib/codec-wrappers/mozjpeg-enc.ts
developit: I'm fine with that, or have it default to 100

src/components/app/index.tsx
```diff
@@ -2,17 +2,58 @@ import { h, Component } from 'preact';
...
-  img?: ImageBitmap
+  sourceImg?: ImageBitmap,
+  sourceData?: ImageData,
+  leftImg?: ImageBitmap,
+  rightImg?: ImageBitmap,
+  loading: boolean
+  leftType: ImageType,
+  rightType: ImageType,
+  leftOptions: CodecOptions,
+  rightOptions: CodecOptions
```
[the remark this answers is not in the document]
developit: It's not necessarily bad practice, but it can make it more difficult to implement shouldComponentUpdate intelligently, and people tend to avoid storing objects in state because mutating them in-place can break shouldComponentUpdate entirely (old===new).  I'm not sure we care that much, so I'm open to nesting to make things cleaner.  It would make the methods to update an option simpler, and would also help avoid coupling the whole app to a left/right UI (maybe we have 4 panels down the line).

src/components/app/index.tsx
```diff
@@ -2,17 +2,58 @@ import { h, Component } from 'preact';
...
 };
 
 export default class App extends Component<Props, State> {
-  state: State = {};
+  state: State = {
+    loading: false,
+    leftType: 'original',
+    rightType: 'jpeg',
+    leftOptions: {},
+    rightOptions: {}
```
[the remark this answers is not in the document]
developit: Yup - wasn't sure how to conditionally type them given that each encoder has its own options type.

src/lib/codec-wrappers/mozjpeg-enc.ts
jakearchibald: Throwing if quality is missing seems like the right rule for now. We can revisit this as we add more compression options.

src/components/app/index.tsx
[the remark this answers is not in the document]
jakearchibald: Agreed. In terms of `shouldComponentUpdate`, we could agree to changing the parent if a property needs changing, immutable-style. I think there are libraries to help with that. Do you have a favourite?

src/components/app/index.tsx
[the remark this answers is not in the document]
jakearchibald: There are TypeScript tricks for this (similar to how it defines the event object type for `addEventListener`), so we can look at this later.

src/components/app/index.tsx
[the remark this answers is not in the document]
developit: [immutability-helpers](https://npm.im/immutability-helpers) is decent.

src/style/index.scss
developit: could just assume all of the styling is temporary, right? we don't have a completed design.

src/components/app/index.tsx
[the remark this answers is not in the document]
developit: Removed the types entirely for now.

src/style/index.scss
jakearchibald: That's fair for this resource, but some styles (such as image-rendering) are intentional.

src/components/app/index.tsx
developit: Removed.

src/lib/codec-wrappers/mozjpeg-enc.ts
developit: Done.

developit:
Alright, let me know if anything else is needed!

src/codecs/codec.ts
```diff
@@ -1,5 +1,5 @@
 export interface Encoder {
-  encode(data: ImageData): Promise<ArrayBuffer>;
+  encode(data: ImageData, options: any): Promise<ArrayBuffer | ImageData>;
```
jakearchibald: We should avoid `any` unless we really can't. IMO this should be:

```ts
export interface Encoder<OptionsType> {
  encode(data: ImageData, options: OptionsType): Promise<ArrayBuffer | ImageData>;
}
```

Also `ArrayBuffer | ImageData` seems weird. Is this for the "original" encoder. It's probably better to treat "original" as something other than an encoder than mess up the interface.

src/codecs/identity/encoder.ts
```diff
@@ -0,0 +1,11 @@
+import { Encoder } from '../codec';
+
+type EncodeOptions = {};
+
+export default class IdentityEncoder implements Encoder {
```
jakearchibald: This would become

```ts
export default class IdentityEncoder implements Encoder<EncodeOptions> {
```

src/codecs/mozjpeg/options.tsx
```diff
@@ -0,0 +1,27 @@
+import { h } from 'preact';
+import { CodecOptions } from '../../components/app';
+
+type Props = {
+  options: CodecOptions,
+  updateOption(e: Event): void,
```
jakearchibald: For event-driven callbacks, I tend to name them `onWhatever`. Is this a good convention? +@surma.

src/components/app/index.tsx
```diff
@@ -2,56 +2,176 @@ import { h, Component } from 'preact';
...
+import Options from '../options';
 
-import {MozJpegEncoder} from '../../lib/codec-wrappers/mozjpeg-enc';
+import { Encoder } from '../../codecs/codec';
+import IdentityEncoder from '../../codecs/identity/encoder';
+import MozJpegEncoder from '../../codecs/mozjpeg/encoder.worker';
+
+export type ImageType = 'original' | 'MozJpeg';
+
+export type CodecOptions = any;
```
jakearchibald: We should avoid an `any` type here, but I'm happy for this to merge as long as an issue is created to fix this (I'm happy to do the fixing).

src/components/output/custom-els/PinchZoom/index.ts
```diff
@@ -268,7 +268,7 @@ export default class PinchZoom extends HTMLElement {
   }
 
   /** Transform the view & fire a change event */
-  private _applyChange (opts: ApplyChangeOpts = {}) {
+  private _applyChange(opts: ApplyChangeOpts = {}) {
```
jakearchibald: I was using the standardjs style, but I dislike the space after the function name,

What lint rules are you using? Check them in.

src/codecs/codec.ts
surma: I think we can treat original as an encoder. `ImageData` doesn’t really make sense, so they should all use `ArrayBuffer` imo.

package.json
```diff
@@ -83,6 +83,9 @@
   },
   "dependencies": {
     "classnames": "^2.2.5",
+    "comlink": "^3.0.3",
+    "comlink-loader": "^1.0.0",
+    "immunize": "git+https://gist.github.com/9ee47f9a6122669b9a1b77587af04ff8.git",
```
surma: lol

src/components/app/index.tsx
```diff
@@ -2,56 +2,176 @@ import { h, Component } from 'preact';
...
+
+const ENCODER_NAMES = {
```
surma: Add a comment where this string is used. Also, can we merge `ENCODER_NAMES` and `ENCODERS` or do they need to be separate?

src/components/app/index.tsx
```diff
@@ -2,56 +2,176 @@ import { h, Component } from 'preact';
...
+  original: 'Original Image',
+  MozJpeg: 'JPEG'
+};
+
+const ENCODERS = {
+  original: IdentityEncoder,
+  MozJpeg: MozJpegEncoder
+};
+
+type Image = {
```
surma: Nit: I prefer `interface` over `type`, as they can be extended.

surma: Same for some other files in this PR.

src/components/app/index.tsx
```diff
@@ -2,56 +2,176 @@ import { h, Component } from 'preact';
...
 
 type Props = {};
 
 type State = {
-  img?: ImageBitmap
+  sourceFile?: File,
+  sourceImg?: ImageBitmap,
+  sourceData?: ImageData,
+  images: Array<Image>,
+  loading: boolean,
```
surma: Since this is _the_ app state, can we add some comments explaining the props? Like, what does `loading` stand for and what is the desired effect?

src/components/app/index.tsx
surma: And/or declare a type (especially if the two merge)

src/components/app/index.tsx
```diff
@@ -2,56 +2,176 @@ import { h, Component } from 'preact';
...
+    ]
+  };
 
   constructor() {
     super();
     // In development, persist application state across hot reloads:
     if (process.env.NODE_ENV === 'development') {
       this.setState(window.STATE);
-      this.componentDidUpdate = () => {
+      let oldCDU = this.componentDidUpdate;
```
surma: Care to explain? 😅

src/components/app/index.tsx
```diff
@@ -2,56 +2,176 @@ import { h, Component } from 'preact';
...
+const ENCODERS = {
+  original: IdentityEncoder,
+  MozJpeg: MozJpegEncoder
+};
+
+type Image = {
+  type: ImageType,
+  options: CodecOptions,
+  data?: ImageBitmap,
+  counter: number,
```
surma: Can we add some comments to this as well? What does counter do?

surma: ... okay found it. Still, comment would be useful.

src/components/app/index.tsx
```diff
@@ -2,56 +2,176 @@ import { h, Component } from 'preact';
...
+  loading: boolean
```
surma: Nit: Call this `currentlyEncoding` or something?

src/components/app/index.tsx
```diff
@@ -2,56 +2,176 @@ import { h, Component } from 'preact';
...
+      if (compressedData instanceof ArrayBuffer) {
+        imageData = new Blob([compressedData], {
+          type: ENCODERS[type].mimeType || ''
+        });
+      } else {
+        imageData = compressedData;
+      }
+      return await createImageBitmap(imageData);
+    } catch (err) {
+      console.error(`Encoding error (type=${type}): ${err}`);
```
surma: Add a `TODO` to inform the user of an error?

src/components/app/index.tsx
```diff
@@ -2,56 +2,176 @@ import { h, Component } from 'preact';
...
+    }
+  }
+
+  render({ }: Props, { loading, error, images }: State) {
+    for (let image of images) {
```
surma: `loading = images.any(image => image.loading)`? :3

src/components/options/index.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+  @bind
+  updateOption(e: Event) {
+    const el = e.currentTarget as HTMLInputElement;
+    this.setOption(el.name, /(rad|box)/i.test(el.type) ? el.checked : el.value);
+  }
+
+  @bind
+  setOption(key: string, value?: any) {
+    const options = Object.assign({}, this.state.options);
```
surma: Just wondering: should we be using `immunize` for all these state-copy-thingies for consistency?

src/components/output/index.tsx
surma: Yeah, it’s a good utility function.

webpack.config.js
```diff
@@ -97,6 +98,10 @@ module.exports = function (_, env) {
             }
           ]
         },
+        {
+          test: /\.worker.[tj]sx?$/,
+          loader: 'comlink-loader'
```
surma: 🎉

webpack.config.js
```diff
@@ -111,16 +116,16 @@ module.exports = function (_, env) {
         {
           // All the codec files define a global with the same name as their file name. `exports-loader` attaches those to `module.exports`.
           test: /\/codecs\/.*\.js$/,
-          loader: 'exports-loader',
+          loader: 'exports-loader'
         },
         {
           test: /\/codecs\/.*\.wasm$/,
```
surma: I think this rule can be removed now, no?

src/components/output/index.tsx
```diff
@@ -7,39 +7,43 @@ import { bind } from '../../lib/util';
...
+    let ctx = canvas.getContext('2d');
```
jakearchibald: `const`

webpack.config.js
developit: wasn't sure - are we not using it anymore?

src/components/options/index.tsx
developit: maybe. I actually just removed it, but if it's cleaner and everyone is on board I'm find adding it back in.

src/components/app/style.scss
developit: We can fix this when we write the actual CSS.

src/components/app/index.tsx
```diff
@@ -25,33 +66,131 @@ export default class App extends Component<Props, State> {
...
 
-  render({ }: Props, { img }: State) {
+  async updateImages() {
+    const { sourceData, leftType, rightType, leftOptions, rightOptions } = this.state;
+    if (!sourceData) return;
+    const id = ++this.compressCounter;
+    this.setState({ loading: true });
+    const [leftImg, rightImg] = await Promise.all([
+      this.updateCompressedImage(sourceData, leftType, leftOptions),
+      this.updateCompressedImage(sourceData, rightType, rightOptions)
```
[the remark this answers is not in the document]
developit: It now updates individual outputs in response to per-image changes.

src/components/output/index.tsx
developit: - [x] move to utility function

src/style/index.scss
developit: added

src/codecs/codec.ts
developit: @surma if you can get that to work, I agree. I couldn't.

developit: - [x] generics for encoder options

src/components/app/index.tsx
developit: I've done the simplest thing for now:

```js
export type CodecOptions = IdentityEncodeOptions | MozJpegEncodeOptions;
```

src/components/output/custom-els/PinchZoom/index.ts
developit: The rules are in our package.json. Surma's PR swaps them out for tslint.

package.json
developit: removed :P

jakearchibald:
@developit @surma @kosamari Ready for review! I've also rebased on master.

src/codecs/encoders.ts
```diff
@@ -0,0 +1,15 @@
...
+export type EncoderState = IdentityEncodeData | MozJPEGEncodeData;
+export type EncoderOptions = IdentityEncodeOptions | MozJPEGEncodeOptions;
+export type EncoderType = keyof typeof encoderMap;
+
+export const encoderMap = {
+  [identity.type]: identity,
+  [mozJPEG.type]: mozJPEG
+};
+
+export const encoders = Array.from(Object.values(encoderMap));
```
jakearchibald: We'll need to edit this file as we add new encoders. It isn't as DRY as I'd like, but I couldn't get TypeScript to do better.

src/codecs/mozjpeg/encoder.ts
```diff
@@ -0,0 +1,17 @@
...
+export const mimeType = 'image/jpeg';
+export const extension = 'jpg';
+export const defaultOptions: EncodeOptions = { quality: 7 };
+
+export async function encode(data: ImageData, options: EncodeOptions = {}) {
+  // This is horrible because TypeScript doesn't realise that
+  // Encoder has been comlinked.
+  const encoder = new Encoder() as any as Promise<Encoder>;
+  return (await encoder).encode(data, options);
+}
```
jakearchibald: All encoders will export these things, except identity which doesn't encode, or have a static extensions / mime type etc.

src/codecs/mozjpeg/encoder.ts
```diff
@@ -0,0 +1,17 @@
...
+export const mimeType = 'image/jpeg';
+export const extension = 'jpg';
+export const defaultOptions: EncodeOptions = { quality: 7 };
+
+export async function encode(data: ImageData, options: EncodeOptions) {
+  // This is horrible because TypeScript doesn't realise that
+  // Encoder has been comlinked.
+  const encoder = new EncoderWorker() as any as Promise<EncoderWorker>;
+  return (await encoder).encode(data, options);
+}
```
jakearchibald: All encoders will export these things, except identity which doesn't encode, or have a static extensions / mime type etc.

src/codecs/mozjpeg/options.tsx
```diff
@@ -0,0 +1,35 @@
...
+            max="100"
+            step="1"
+            value={'' + options.quality}
+            onChange={this.onChange}
+          />
+        </label>
+      </div>
+    );
+  }
+}
```
jakearchibald: I removed state from this component. The form elements are still controlled, but by the app.

src/codecs/mozjpeg/encoder.ts
```diff
@@ -0,0 +1,17 @@
...
+export const type = 'mozjpeg';
+export const label = 'MozJPEG';
+export const mimeType = 'image/jpeg';
+export const extension = 'jpg';
+export const defaultOptions: EncodeOptions = { quality: 7 };
+
+export async function encode(data: ImageData, options: EncodeOptions) {
+  // This is horrible because TypeScript doesn't realise that
+  // Encoder has been comlinked.
+  const encoder = new EncoderWorker() as any as Promise<EncoderWorker>;
```
surma: shouldn’t `await new EncoderWorker()` just work?

src/components/app/index.tsx
```diff
@@ -2,56 +2,192 @@ import { h, Component } from 'preact';
...
+    let image = images[index];
+
+    // Each time we trigger an async encode, the ID changes.
+    const id = ++image.counter;
+    image.loading = true;
+    this.setState({ });
+    const result = await this.updateCompressedImage(source, image.encoderState);
+    image = this.state.images[index];
+    // If another encode has been initiated since we started, ignore this one.
+    if (image.counter !== id) return;
```
surma: Maybe I’m misunderstanding, but if `image.counter < id` it means an older encoding job has finished and we shouldn’t early return?

jakearchibald: Agreed. I messed up the counter pretty well. Will fix.

jakearchibald:
I've created two counters, one for loading and one for loaded. This means we can render the result of one encode even if another is in progress, but they can't render out of order.

PR #42, opened 2018-05-29
Commit binaries

### LINKED ISSUE #37: Build codecs as part of post-install
If this proves super-painful, we can just check them in.

### LINKED ISSUE #41: Document codec build dependencies
I'm guessing it's `automake` and `docker` for mozjpeg?

surma:
In the end building the codecs on-demand is probably not worth the effort. The files are small enough to be checked in as well. I might revisit this decision at a later point, but for now I wanna move forward.

Closes #37 
Closes #41


jakearchibald:
I'm ok with this. If we find that it bloats the repo, we can store the wasms in cloudstorage and fetch them as part of install.

We should, however, make sure that we can all build the codecs, so we can confirm the docs are good.

PR #43, opened 2018-05-29
Testing

surma:
Okay so... testing. This PR adds infrastructure to run end-to-end tests using puppeteer as well as unit tests with Karma. 

It’s quite the ball-ache due to the typescript+webpack setup. So I’d love for @developit to take a look at this and hopefully tell me that I’m dumb and it’s all much much easier than this 😅

I also added a `Dockerfile` that can be used to build a container to run the tests (mostly so you can run the test suite locally in the same way Travis would. Makes debugging Travis a lot easier). However, since Travis will always run on a fresh clone, we _need_ to fix that bug now that styles are not being written on the first build.

The tests in here are mostly dummies so we can talk about the infrastructure. Might be worth talking about in a face-to-face, as it’s quite the monstrous setup. LMK.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,20 @@
+FROM selenium/node-chrome:latest
+
+USER root
+
+RUN apt-get update -qqy \
+  && apt-get -qqy install python git build-essential \
+  && rm -rf /var/lib/apt/lists/* /var/cache/apt/* \
+  && rm /bin/sh && ln -s /bin/bash /bin/sh \
+  && chown seluser /usr/local
+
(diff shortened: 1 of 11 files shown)


developit:
This is a promise to have this reviewed by tomorrow EOD

developit:
Update: a few fixes locally, but I also tried just dropping `karmatic` in and it works!*

I'm going to open a PR so you can compare.

* with a bugfix I'm releasing right now

surma:
Closing this in favor of #67

PR #50, opened 2018-06-13
Decoder for webp

surma:
Tested in Safari, Firefox and Chrome \o/


surma:
I’ll do the proper JS wrapper in a separate PR

codecs/webp_dec/.gitignore
```diff
@@ -0,0 +1,2 @@
+/*.wasm
```
jakearchibald: Weren't we going to start checking the wasm in?

codecs/webp_dec/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+## API
+
+### `int version()`
+
+Returns the version of libwebp as a number. va.b.c is encoded as 0x0a0b0c
+
+### `uint8_t* create_buffer(int size)`
+
+Allocates an buffer for the file data.
```
jakearchibald: "a buffer"

codecs/webp_dec/example.html
```diff
@@ -0,0 +1,44 @@
+<!doctype html>
+<script src='webp_dec.js'></script>
+<script>
+  const Module = webp_dec();
```
jakearchibald: I'm not going to review this file in terms of project types, but isn't this not-a-constructor, so should start with a lower-case?

The reason I ask, is I saw @developit doing a similar thing in the options PR. I'm worried I've missed out on some change to consensus I always thought it was:

```
ThisIsAClass
THIS_IS_A_PRIMITIVE_CONSTANT
thisIsEverythingElse
```

codecs/webp_dec/example.html
```diff
@@ -0,0 +1,44 @@
...
+
+  async function loadFile(src) {
+    const resp = await fetch(src);
+    return await resp.arrayBuffer();
```
jakearchibald: Just `return` here is fine. The `await` is redundant.

codecs/webp_dec/.gitignore
surma: Yeah, that’s currently only in the testing branch. I’ll move that to master once this lands.

codecs/webp_dec/example.html
surma: It’s mostly this way ’cause that’s how all the Emscripten code examples are written. By default, Emscripten glue creates a global called `Module`. We set a flag so that instead it registers a function that returns that `Module`, hence the naming. It’s nbd to change that, tho

codecs/webp_dec/example.html
surma: Oh cool. I thought it would make the stack trace look different, but it doesn’t. TIL.

jakearchibald: https://jakearchibald.com/2017/await-vs-return-vs-return-await/ 😄

PR #56, opened 2018-06-22
Adding in Drag and Drop support to fix #45

### LINKED ISSUE #45: Drag and drop file loading
This feature will be available on all screens of the app, not just the intro screen.

There should be some kind of drop indicator, but we shouldn't worry too much about the specific design yet.

PaulKinlan:
* Defines a `<file-drop>` element that lets you filter on File mimetypes via an `accepts` attr (it can take image/png, or wildcards.). When a matching file is found, the first file in the dataTransfer object is attached to a new 'onFileDrop event.
* Implements `<file-drop>` in app, it wraps the entire app Component and hooks up with the logic to take the file element and load it when the 'onFileDrop' event triggers.

Current known issue, the entire App is position absolute, so I've just made my `<file-drop>` default to display: block, width & height 100%... kinda doesn't fell right.

The file-drop element also emulates the pseudo classes that drag and drop is supposed to support and it applies them only once on dragEnter and dragLeave to avoid lots of transitions and layout recalcs etc.

Learnt a bit of typescript and preact... innit.

### DIFF
--- a/src/components/app/custom-els/FileDrop/index.ts
+++ b/src/components/app/custom-els/FileDrop/index.ts
@@ -1,0 +1,116 @@
+import { bind } from '../../../../lib/util';
+import './styles.css';
+
+function firstMatchingItem(list: DataTransferItemList, acceptVal: string): DataTransferItem | undefined {
+  // Split accepts values by ',' then by '/'. Trim everything & lowercase.
+  const accepts = acceptVal.toLowerCase().split(',').map(accept => {
+    return accept.trim().split('/').map(part => part.trim());
+  }).filter(acceptParts => acceptParts.length === 2); // Filter invalid values
+
+  return Array.from(list).find(item => {
+    if (item.kind !== 'file') return false;
+
+    // 'Parse' the type.
+    const [typeMain, typeSub] = item.type.toLowerCase().split('/').map(s => s.trim());
+
+    for (const [acceptMain, acceptSub] of accepts) {
+      // Look for an exact match, or a partial match if * is accepted, eg image/*.
+      if (typeMain === acceptMain && (acceptSub === '*' || typeSub === acceptSub)) {
+        return true;
+      }
+    }
+    return false;
+  });
+}
+
+interface FileDropEventInit extends EventInit {
+  file: File;
+}
+
+export class FileDropEvent extends Event {
+  private _file: File;
+  constructor(typeArg: string, eventInitDict: FileDropEventInit) {
+    super(typeArg, eventInitDict);
+    this._file = eventInitDict.file;
+  }
+
+  get file(): File {
+    return this._file;
+  }
+}
+
+/*
+  Example Usage.
+  <file-drop
+    accept='image/*'
+    class='drop-valid|drop-invalid'
+  >
+   [everything in here is a drop target.]
+  </file-drop>
+
+  dropElement.addEventListner('dropfile', (event) => console.log(event.detail))
+*/
+export class FileDrop extends HTMLElement {
+
+  private _dragEnterCount = 0;
+
+  constructor() {
+    super();
+    this.addEventListener('dragover', (event) => event.preventDefault());
+    this.addEventListener('drop', this._onDrop);
+    this.addEventListener('dragenter', this._onDragEnter);
+    this.addEventListener('dragend', () => this._reset());
+    this.addEventListener('dragleave', this._onDragLeave);
+  }
+
+  get accept() {
+    return this.getAttribute('accept') || '';
+  }
+
+  set accept(val: string) {
+    this.setAttribute('accept', val);
+  }
+
+  @bind
+  private _onDragEnter(event: DragEvent) {
+    this._dragEnterCount++;
+    if (this._dragEnterCount > 1) return;
+
+    // We don't have data, attempt to get it and if it matches, set the correct state.
+    const dragDataItem = firstMatchingItem(event.dataTransfer.items, this.accept);
+    if (dragDataItem) {
+      this.classList.add('drop-valid');
+    } else {
+      this.classList.add('drop-invalid');
+    }
+  }
+
+  @bind
+  private _onDragLeave() {
+    this._dragEnterCount--;
+    if (this._dragEnterCount === 0) {
+      this._reset();
+    }
+  }
+
+  @bind
+  private _onDrop(event: DragEvent) {
+    event.preventDefault();
+    this._reset();
+    const dragDataItem = firstMatchingItem(event.dataTransfer.items, this.accept);
+    if (!dragDataItem) return;
+
+    const file = dragDataItem.getAsFile();
+    if (file === null) return;
+
+    this.dispatchEvent(new FileDropEvent('filedrop', { file }));
+  }
+
+  private _reset() {
+    this._dragEnterCount = 0;
+    this.classList.remove('drop-valid');
+    this.classList.remove('drop-invalid');
+  }
+}
+
+customElements.define('file-drop', FileDrop);
--- a/src/components/app/custom-els/FileDrop/missing-types.d.ts
+++ b/src/components/app/custom-els/FileDrop/missing-types.d.ts
@@ -1,0 +1,20 @@
+import { FileDropEvent, FileDrop } from ".";
+
+
+declare global {
+
+  interface HTMLElementEventMap {
+    'filedrop': FileDropEvent;
+  }
+
+  namespace JSX {
+    interface IntrinsicElements {
+      'file-drop': FileDropAttributes;
+    }
+
+    interface FileDropAttributes extends HTMLAttributes {
+      accept?: string;
+      onfiledrop?: ((this: FileDrop, ev: FileDropEvent) => any) | null;
+    }
+  }
+}
--- a/src/components/app/custom-els/FileDrop/styles.css
+++ b/src/components/app/custom-els/FileDrop/styles.css
@@ -1,0 +1,3 @@
+file-drop {
+  display: block;
+}
--- a/src/components/app/index.tsx
+++ b/src/components/app/index.tsx
@@ -35,8 +35,8 @@ interface State {
   error?: string;
 }
 
-function initialState(): State {
-  return {
+export default class App extends Component<Props, State> {
+  state: State = {
     loading: false,
     images: [
       {
@@ -53,10 +53,6 @@ function initialState(): State {
       }
     ]
   };
-}
-
-export default class App extends Component<Props, State> {
-  state: State = initialState();
 
   constructor() {
     super();
@@ -117,7 +113,6 @@ export default class App extends Component<Props, State> {
   async onFileDrop(event: FileDropEvent) {
     const { file } = event;
     if (!file) return;
-    this.setState(initialState());
     await this.updateFile(file);
   }
 
--- a/src/components/app/index.tsx
+++ b/src/components/app/index.tsx
@@ -3,6 +3,8 @@ import { bind, bitmapToImageData } from '../../lib/util';
 import * as style from './style.scss';
 import Output from '../output';
 import Options from '../options';
+import { FileDropEvent } from './custom-els/FileDrop';
+import './custom-els/FileDrop';
 
 import * as mozJPEG from '../../codecs/mozjpeg/encoder';
 import * as identity from '../../codecs/identity/encoder';
@@ -33,8 +35,8 @@ interface State {
   error?: string;
 }
 
-export default class App extends Component<Props, State> {
-  state: State = {
+function initialState(): State {
+  return {
     loading: false,
     images: [
       {
@@ -51,6 +53,10 @@ export default class App extends Component<Props, State> {
       }
     ]
   };
+}
+
+export default class App extends Component<Props, State> {
+  state: State = initialState();
 
   constructor() {
     super();
@@ -104,6 +110,18 @@ export default class App extends Component<Props, State> {
     const fileInput = event.target as HTMLInputElement;
     const file = fileInput.files && fileInput.files[0];
     if (!file) return;
+    await this.updateFile(file);
+  }
+
+  @bind
+  async onFileDrop(event: FileDropEvent) {
+    const { file } = event;
+    if (!file) return;
+    this.setState(initialState());
+    await this.updateFile(file);
+  }
+
+  async updateFile(file: File) {
     this.setState({ loading: true });
     try {
       const bmp = await createImageBitmap(file);
@@ -112,7 +130,7 @@ export default class App extends Component<Props, State> {
       this.setState({
         source: { data, bmp, file },
         error: undefined,
-        loading: false
+        loading: false,
       });
     } catch (err) {
       this.setState({ error: 'IMAGE_INVALID', loading: false });
@@ -173,29 +191,33 @@ export default class App extends Component<Props, State> {
     loading = loading || images.some(image => image.loading);
 
     return (
-      <div id="app" class={style.app}>
-        {(leftImageBmp && rightImageBmp) ? (
-          <Output leftImg={leftImageBmp} rightImg={rightImageBmp} />
-        ) : (
-          <div class={style.welcome}>
-            <h1>Select an image</h1>
-            <input type="file" onChange={this.onFileChange} />
-          </div>
-        )}
-        {images.map((image, index) => (
-          <span class={index ? style.rightLabel : style.leftLabel}>{encoderMap[image.encoderState.type].label}</span>
-        ))}
-        {images.map((image, index) => (
-          <Options
-            class={index ? style.rightOptions : style.leftOptions}
-            encoderState={image.encoderState}
-            onTypeChange={this.onEncoderChange.bind(this, index)}
-            onOptionsChange={this.onOptionsChange.bind(this, index)}
-          />
-        ))}
-        {loading && <span style={{ position: 'fixed', top: 0, left: 0 }}>Loading...</span>}
-        {error && <span style={{ position: 'fixed', top: 0, left: 0 }}>Error: {error}</span>}
-      </div>
+      <file-drop accept="image/*" onfiledrop={this.onFileDrop}>
+        <div id="app" class={style.app}>
+          {(leftImageBmp && rightImageBmp) ? (
+            <Output leftImg={leftImageBmp} rightImg={rightImageBmp} />
+          ) : (
+              <div class={style.welcome}>
+                <h1>Select an image</h1>
+                <input type="file" onChange={this.onFileChange} />
+              </div>
+            )}
+          {images.map((image, index) => (
+            <span class={index ? style.rightLabel : style.leftLabel}>
+              {encoderMap[image.encoderState.type].label}
+            </span>
+          ))}
+          {images.map((image, index) => (
+            <Options
+              class={index ? style.rightOptions : style.leftOptions}
+              encoderState={image.encoderState}
+              onTypeChange={this.onEncoderChange.bind(this, index)}
+              onOptionsChange={this.onOptionsChange.bind(this, index)}
+            />
+          ))}
+          {loading && <span style={{ position: 'fixed', top: 0, left: 0 }}>Loading...</span>}
+          {error && <span style={{ position: 'fixed', top: 0, left: 0 }}>Error: {error}</span>}
+        </div>
+      </file-drop>
     );
   }
 }
--- a/codecs/mozjpeg_enc/package-lock.json
+++ b/codecs/mozjpeg_enc/package-lock.json
@@ -21,8 +21,8 @@
       "integrity": "sha1-u13KOCu5TwXhUZQ3PRb9O6HKEQ0=",
       "dev": true,
       "requires": {
-        "delegates": "1.0.0",
-        "readable-stream": "2.3.5"
+        "delegates": "^1.0.0",
+        "readable-stream": "^2.0.6"
       }
     },
     "balanced-match": {
@@ -43,7 +43,7 @@
       "integrity": "sha1-ysMo977kVzDUBLaSID/LWQ4XLV4=",
       "dev": true,
       "requires": {
-        "readable-stream": "2.3.5"
+        "readable-stream": "^2.0.5"
       }
     },
     "block-stream": {
@@ -52,7 +52,7 @@
       "integrity": "sha1-E+v+d4oDIFz+A3UUgeu0szAMEmo=",
       "dev": true,
       "requires": {
-        "inherits": "2.0.3"
+        "inherits": "~2.0.0"
       }
     },
     "brace-expansion": {
@@ -61,7 +61,7 @@
       "integrity": "sha512-iCuPHDFgrHX7H2vEI/5xpz07zSHB00TpugqhmYtVmMO6518mCuRMoOYFldEBl0g187ufozdaHgWKcYFb61qGiA==",
       "dev": true,
       "requires": {
-        "balanced-match": "1.0.0",
+        "balanced-match": "^1.0.0",
         "concat-map": "0.0.1"
       }
     },
@@ -72,8 +72,8 @@
       "dev": true,
       "requires": {
         "base64-js": "0.0.8",
-        "ieee754": "1.1.8",
-        "isarray": "1.0.0"
+        "ieee754": "^1.1.4",
+        "isarray": "^1.0.0"
       }
     },
     "buffer-crc32": {
@@ -88,10 +88,10 @@
       "integrity": "sha512-Cg8/ZSBEa8ZVY9HspcGUYaK63d/bN7rqS3CYCzEGUxuYv6UlmcjzDUz2fCFFHyTvUW5Pk0I+3hkA3iXlIj6guA==",
       "dev": true,
       "requires": {
-        "get-proxy": "2.1.0",
-        "isurl": "1.0.0",
-        "tunnel-agent": "0.6.0",
-        "url-to-options": "1.0.1"
+        "get-proxy": "^2.0.0",
+        "isurl": "^1.0.0-alpha5",
+        "tunnel-agent": "^0.6.0",
+        "url-to-options": "^1.0.1"
       }
     },
     "code-point-at": {
@@ -106,7 +106,7 @@
       "integrity": "sha1-Br42f+v9oMMwqh4qBy09yXYkJdQ=",
       "dev": true,
       "requires": {
-        "graceful-readlink": "1.0.1"
+        "graceful-readlink": ">= 1.0.0"
       }
     },
     "concat-map": {
@@ -121,8 +121,8 @@
       "integrity": "sha1-q6CXR9++TD5w52am5BWG4YWfxvI=",
       "dev": true,
       "requires": {
-        "ini": "1.3.5",
-        "proto-list": "1.2.4"
+        "ini": "^1.3.4",
+        "proto-list": "~1.2.1"
       }
     },
     "console-control-strings": {
@@ -158,14 +158,14 @@
       "integrity": "sha1-eu3YVCflqS2s/lVnSnxQXpbQH50=",
       "dev": true,
       "requires": {
-        "decompress-tar": "4.1.1",
-        "decompress-tarbz2": "4.1.1",
-        "decompress-targz": "4.1.1",
-        "decompress-unzip": "4.0.1",
-        "graceful-fs": "4.1.11",
-        "make-dir": "1.2.0",
-        "pify": "2.3.0",
-        "strip-dirs": "2.1.0"
+        "decompress-tar": "^4.0.0",
+        "decompress-tarbz2": "^4.0.0",
+        "decompress-targz": "^4.0.0",
+        "decompress-unzip": "^4.0.1",
+        "graceful-fs": "^4.1.10",
+        "make-dir": "^1.0.0",
+        "pify": "^2.3.0",
+        "strip-dirs": "^2.0.0"
       },
       "dependencies": {
         "pify": {
@@ -182,7 +182,7 @@
       "integrity": "sha1-gKTdMjdIOEv6JICDYirt7Jgq3/M=",
       "dev": true,
       "requires": {
-        "mimic-response": "1.0.0"
+        "mimic-response": "^1.0.0"
       }
     },
     "decompress-tar": {
@@ -191,9 +191,9 @@
       "integrity": "sha512-JdJMaCrGpB5fESVyxwpCx4Jdj2AagLmv3y58Qy4GE6HMVjWz1FeVQk1Ct4Kye7PftcdOo/7U7UKzYBJgqnGeUQ==",
       "dev": true,
       "requires": {
-        "file-type": "5.2.0",
-        "is-stream": "1.1.0",
-        "tar-stream": "1.5.5"
+        "file-type": "^5.2.0",
+        "is-stream": "^1.1.0",
+        "tar-stream": "^1.5.2"
       }
     },
     "decompress-tarbz2": {
@@ -202,11 +202,11 @@
       "integrity": "sha512-s88xLzf1r81ICXLAVQVzaN6ZmX4A6U4z2nMbOwobxkLoIIfjVMBg7TeguTUXkKeXni795B6y5rnvDw7rxhAq9A==",
       "dev": true,
       "requires": {
-        "decompress-tar": "4.1.1",
-        "file-type": "6.2.0",
-        "is-stream": "1.1.0",
-        "seek-bzip": "1.0.5",
-        "unbzip2-stream": "1.2.5"
+        "decompress-tar": "^4.1.0",
+        "file-type": "^6.1.0",
+        "is-stream": "^1.1.0",
+        "seek-bzip": "^1.0.5",
+        "unbzip2-stream": "^1.0.9"
       },
       "dependencies": {
         "file-type": {
@@ -223,9 +223,9 @@
       "integrity": "sha512-4z81Znfr6chWnRDNfFNqLwPvm4db3WuZkqV+UgXQzSngG3CEKdBkw5jrv3axjjL96glyiiKjsxJG3X6WBZwX3w==",
       "dev": true,
       "requires": {
-        "decompress-tar": "4.1.1",
-        "file-type": "5.2.0",
-        "is-stream": "1.1.0"
+        "decompress-tar": "^4.1.1",
+        "file-type": "^5.2.0",
+        "is-stream": "^1.1.0"
       }
     },
     "decompress-unzip": {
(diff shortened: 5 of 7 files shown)


PaulKinlan:
Just a quick update, don't review this just yet, there are a number of changes that I would like to make to improve the code.

PaulKinlan:
Ready for review ;)

codecs/mozjpeg_enc/example.html
```diff
@@ -27,6 +27,7 @@
       get_result_pointer: Module.cwrap('get_result_pointer', 'number', []),
       get_result_size: Module.cwrap('get_result_size', 'number', []),
     };
+    console.log('Version:', api.version().toString(16));
```
jakearchibald: Remove log. @surma is this from you?

surma:
This PR merges `master` into `options` as a side-effect which makes it very hard to review. Any reason why this PR is not against master?

codecs/mozjpeg_enc/example.html
surma: Yes, it’s just the example pages for the codecs to see that they are actually using the right version. This is on master already.

PaulKinlan:
Because it was suggested to fork from options branch because it had a number of things that I think I needed, also I couldn't get it to work (re: WASM issues) without having your master changes in.

PaulKinlan:
Pretty much all of the changes are isolated into src/components/app/*

src/components/app/custom-els/FileDrop/index.ts
```diff
@@ -0,0 +1,144 @@
+import * as styles from './styles.css';
+
+class DragState {
+
+  private element: FileDrop;
+  private hasDragData: boolean;
```
jakearchibald: I don't think we need this. Just call `getMatchingFiles` each time you need to know. If this becomes a performance issue (which I seriously doubt) we can look at a way to fix it.

src/components/app/custom-els/FileDrop/index.ts
```diff
@@ -0,0 +1,144 @@
...
+  bindEvents() {
+    this.element.addEventListener('dragover', this.onDragOver.bind(this));
+    this.element.addEventListener('drop', this.onDrop.bind(this));
+    this.element.addEventListener('dragenter', this.onDragEnter.bind(this));
+    this.element.addEventListener('dragend', this.onDragEnd.bind(this));
+    this.element.addEventListener('dragleave', this.onDragLeave.bind(this));
+  }
+
+  onDragEnter(e: DragEvent) {
+    this.dragNest++;
```
jakearchibald: After this, can't we just return if `dragNest` is > 1?

src/components/app/custom-els/FileDrop/index.ts
```diff
@@ -0,0 +1,144 @@
...
+      }
+    }
+    return;
+  }
+}
+
+/*
+  <file-drop
+    accepts='image/*'
+    onFileDrop='javascript: (ev) => console.log(ev.detail);'
```
jakearchibald: Pretty sure this won't work.

src/components/app/index.tsx
```diff
@@ -93,6 +98,19 @@ export default class App extends Component<Props, State> {
     const fileInput = event.target as HTMLInputElement;
     const sourceFile = fileInput.files && fileInput.files[0];
     if (!sourceFile) return;
+    await this.updateFile(sourceFile);
+  }
+
+  @bind
+  async onFileDrop(event: CustomEvent) {
+    const sourceFile = event.detail as File;
+    this.setState(initialState());
```
jakearchibald: I wouldn't expect dropping a file to change my code settings.

src/components/app/index.tsx
```diff
@@ -93,6 +98,19 @@ export default class App extends Component<Props, State> {
...
+
+    if (!sourceFile) return;
```
jakearchibald: Should this check appear before resetting state?

src/components/app/index.tsx
```diff
@@ -147,7 +165,9 @@ export default class App extends Component<Props, State> {
     const rightImg = images[1].data;
 
     return (
+      <file-drop accepts="image/*" onFileDrop={this.onFileDrop}>
```
jakearchibald: Indent everything inside this

jakearchibald:
@surma I changed the base of the PR. Easier to review now.

src/components/app/custom-els/FileDrop/index.ts
```diff
@@ -0,0 +1,144 @@
...
+    this.hasDragData = false;
+    this.dragNest = 0;
+  }
+
+  bindEvents() {
+    this.element.addEventListener('dragover', this.onDragOver.bind(this));
+    this.element.addEventListener('drop', this.onDrop.bind(this));
+    this.element.addEventListener('dragenter', this.onDragEnter.bind(this));
+    this.element.addEventListener('dragend', this.onDragEnd.bind(this));
+    this.element.addEventListener('dragleave', this.onDragLeave.bind(this));
```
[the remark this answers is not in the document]
PaulKinlan: TIL.

src/components/app/custom-els/FileDrop/index.ts
jakearchibald: It works in preact but only because it's using `addEventListener` under the hood.

src/components/app/index.tsx
PaulKinlan: So.... there were some issues. This reset's everything back to what the page was on first load, there's a heap of state wrapped up with images, codecs, and other changes that I didn't feel like they applied when I dropped a new image on to the canvas.

src/components/app/index.tsx
PaulKinlan: good catch.

src/components/app/custom-els/FileDrop/index.ts
```diff
@@ -0,0 +1,144 @@
...
+    this.hasDragData = false;
+    this.dragNest = 0;
+    this.element.classList.remove(styles.dropValid);
+    this.element.classList.remove(styles.dropInvalid);
+    this.element.classList.remove(styles.dropActive);
+  }
+
+  private convertMimeTypeToRegex(mimeType: string): string {
+    return mimeType.replace(/\*/, '[^/]+' ).replace(/\//g, '\\/');
+  }
```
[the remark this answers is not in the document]
PaulKinlan: sg.

src/components/app/custom-els/FileDrop/missing-types.d.ts
```diff
@@ -0,0 +1,10 @@
+declare namespace JSX {
+  interface IntrinsicElements {
+    'file-drop': FileDropAttributes;
+  }
+
+  interface FileDropAttributes extends HTMLAttributes {
+    accepts?: string;
+    onFileDrop?: ((this: HTMLElement, ev: CustomEvent) => any) | null;
```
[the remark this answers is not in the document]
PaulKinlan: It's 'implemented', in that it dispatches the event. I wasn't actually sure if I should make a onFileDrop explicitily... I probably should.

src/components/app/custom-els/FileDrop/index.ts
```diff
@@ -0,0 +1,144 @@
...
+
+  get accepts () {
+    return this._accepts;
+  }
+
+  set accepts (val: string) { 
+    this._accepts = val;
+
+    if (val) {
+      this.setAttribute(this._accepts, '');
```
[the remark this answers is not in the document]
PaulKinlan: urgh.

src/components/app/custom-els/FileDrop/index.ts
```diff
@@ -0,0 +1,144 @@
...
+  private _dragState: DragState;
+
+  constructor() {
+    super();
+    this._accepts = '';
+    this._dragState = new DragState(this);
+  }
+
+  connectedCallback() {
+    this._dragState.bindEvents();
```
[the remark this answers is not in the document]
PaulKinlan: /me reads spec. Thought connectedCallback was only once, and if it was moved between documents it was adoptedCallback.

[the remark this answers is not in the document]
PaulKinlan: It does make sense though, as I'm not waiting for anything inside the element other than itself.

src/components/app/custom-els/FileDrop/index.ts
```diff
@@ -0,0 +1,144 @@
...
+  }
+
+  get accepts () {
+    return this._accepts;
```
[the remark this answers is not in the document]
PaulKinlan: This is a good catch. adding observed attributes.

src/components/app/custom-els/FileDrop/index.ts
```diff
@@ -0,0 +1,144 @@
...
+    let mimeRegex = this.convertMimeTypeToRegex(accept);
+    let acceptsRegex = new RegExp(mimeRegex);
+
+    for(let i=0; i<list.length; i++) {
+      let file = list[i];
+      if(acceptsRegex.test(file.type)) {
+        return file;
+      }
+    }
+    return;
```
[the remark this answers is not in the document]
PaulKinlan: FileList doesn't have find on it, and I didn't want to `Array.from` because it wasn't well supported in Edge iirc or even Chrome used by search (iirc).

src/components/app/custom-els/FileDrop/index.ts
```diff
@@ -0,0 +1,144 @@
...
+    let mimeRegex = this.convertMimeTypeToRegex(accept);
+    let acceptsRegex = new RegExp(mimeRegex);
+
+    for(let i=0; i<list.length; i++) {
+      let item = list[i];
+      if(item.kind === 'file' && acceptsRegex.test(item.type)) {
+        return true;
+      }
+    }
+    return false;
```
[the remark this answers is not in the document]
PaulKinlan: FileList doesn't have find on it, and I didn't want to Array.from because it wasn't well supported in Edge iirc or even Chrome used by search (iirc).

src/components/app/custom-els/FileDrop/index.ts
```diff
@@ -0,0 +1,144 @@
...
+  private convertMimeTypeToRegex(mimeType: string): string {
+    return mimeType.replace(/\*/, '[^/]+' ).replace(/\//g, '\\/');
+  }
+
+  private hasMatchingFile(list: DataTransferItemList, accept: string): boolean {
+    let mimeRegex = this.convertMimeTypeToRegex(accept);
+    let acceptsRegex = new RegExp(mimeRegex);
+
+    for(let i=0; i<list.length; i++) {
+      let item = list[i];
```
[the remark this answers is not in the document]
PaulKinlan: I might have made an incorrect assumption that I couldn't do an for-of on a FileList... lemme revist that.

src/components/app/custom-els/FileDrop/index.ts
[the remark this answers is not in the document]
jakearchibald: I'm sure we're breaking Chrome search in other ways. `Array.from` is fine.

src/components/app/custom-els/FileDrop/index.ts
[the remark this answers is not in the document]
jakearchibald: Connected is every time the element is appended. Ugh, MDN's docs are bad here. I'll update them.

jakearchibald:
It'd be nice to expose enter/leave events too. But y'know, sensibly, so `enter` can only ever fire once in a row.

PaulKinlan:
What would you enter and leave events for? There's a set of classes that understand the higher-level state (of which I should create events for)....

Mind if we break these improvements to file-drop into another issue and I can work on adding them?

jakearchibald:
@PaulKinlan Say I wanted to do some fancy kind of animation in reaction to these events which can't simply be done in CSS.

PaulKinlan:
Understood. Added an issue for improving.

Also, I've updated the array.find pieces.

jakearchibald:
@PaulKinlan let me know when this is ready for another review

jakearchibald:
> Current known issue, the entire App is position absolute, so I've just made my <file-drop> default to display: block, width & height 100%... kinda doesn't fell right.

Don't worry about that. That'll change once we apply the design.

PaulKinlan:
@jakearchibald it's ready for review.

src/components/app/custom-els/FileDrop/index.ts
```diff
@@ -0,0 +1,141 @@
...
+function hasMatchingFile(list: DataTransferItemList, accept: string): boolean {
+  const mimeRegex = convertMimeTypeToRegex(accept);
+  const acceptsRegex = new RegExp(mimeRegex);
+
+  const firstFile = Array.from(list).find(item => item.kind === 'file' && acceptsRegex.test(item.type))
+
+  return !!firstFile;
+}
+
+function firstMatchingFile(list: FileList, accept: string): File | undefined {
```
jakearchibald: You could probably make `hasMatchingFile` and `firstMatchingFile` the same function that takes `FileList | DataTransferItemList`.

src/components/app/custom-els/FileDrop/index.ts
```diff
@@ -0,0 +1,141 @@
...
+    this._element.addEventListener('drop', this._onDrop);
+    this._element.addEventListener('dragenter', this._onDragEnter);
+    this._element.addEventListener('dragend', this._onDragEnd);
+    this._element.addEventListener('dragleave', this._onDragLeave);
+  }
+
+  @bind
+  private _onDragEnter(event: DragEvent) {
+    this._dragEnterCount++;
+    if(this._hasDragData == false) {
```
jakearchibald: Fix your ifs!

src/components/app/custom-els/FileDrop/index.ts
```diff
@@ -0,0 +1,141 @@
...
+  }
+
+  @bind
+  private _onDragEnd(event: DragEvent) {
+    this._reset();
+  }
+
+  @bind
+  private _onDragOver(event: DragEvent) {
+    if(!!this._hasDragData) {
```
jakearchibald: No need for `!!`.

src/components/app/custom-els/FileDrop/index.ts
```diff
@@ -0,0 +1,141 @@
...
+    return true;
+  }
+
+  @bind
+  private _onDrop(event: DragEvent) {
+    if(this._hasDragData) {
+      event.preventDefault();
+      let dragData = firstMatchingFile(event.dataTransfer.files, this._element.accept);
+      
+      this._element.dispatchEvent(new CustomEvent('filedrop', { detail: dragData }));
```
jakearchibald: Don't use `CustomEvent`.

src/components/app/custom-els/FileDrop/index.ts
```diff
@@ -0,0 +1,141 @@
...
+      return false;
+    }
+    return true;
+  }
+
+  @bind
+  private _onDrop(event: DragEvent) {
+    if(this._hasDragData) {
+      event.preventDefault();
+      let dragData = firstMatchingFile(event.dataTransfer.files, this._element.accept);
```
jakearchibald: `const`.

src/components/app/custom-els/FileDrop/index.ts
```diff
@@ -0,0 +1,141 @@
...
+
+  constructor(private readonly _element: FileDrop) {
+  }
+
+  bindEvents() {
+    this._element.addEventListener('dragover', this._onDragOver);
+    this._element.addEventListener('drop', this._onDrop);
+    this._element.addEventListener('dragenter', this._onDragEnter);
+    this._element.addEventListener('dragend', this._onDragEnd);
+    this._element.addEventListener('dragleave', this._onDragLeave);
```
jakearchibald: These should just be in the constructor.

src/components/app/custom-els/FileDrop/index.ts
```diff
@@ -0,0 +1,141 @@
...
+  constructor() {
+    super();
+    this._dragState = new DragState(this);
+    this._dragState.bindEvents();
+  }
+
+  static get observedAttributes() { return ["accept"]; }
+
+  attributeChangedCallback(name:string, oldValue:string, newValue: string) {
+    if(name === 'accept' && oldValue != newValue) {
```
jakearchibald: Better to make the attribute the source or truth.

src/components/app/custom-els/FileDrop/missing-types.d.ts
```diff
@@ -0,0 +1,10 @@
+declare namespace JSX {
+  interface IntrinsicElements {
+    'file-drop': FileDropAttributes;
+  }
+
+  interface FileDropAttributes extends HTMLAttributes {
+    accepts?: string;
+    onfiledrop?: ((this: HTMLElement, ev: CustomEvent) => any) | null;
```
jakearchibald: Not implemented.

src/components/app/custom-els/FileDrop/styles.css
```diff
@@ -0,0 +1,22 @@
...
+  transition: opacity 200ms ease-in-out, background-color 200ms;
+  opacity: 0.5;
+  background-color:green;
+}
+
+.drop-invalid {
+  transition: opacity 200ms ease-in-out, background-color 200ms;
+  opacity: 0.5;
+  background-color:red;
+}
```
jakearchibald: Same issues as before.

src/components/app/custom-els/FileDrop/index.ts
```diff
@@ -0,0 +1,141 @@
+import * as styles from './styles.css';
+import { bind } from '../../../../lib/util';
+
+
+function convertMimeTypeToRegex(mimeType: string): string {
+  return mimeType.replace(/\*/, '[^/]+' ).replace(/\//g, '\\/');
+}
```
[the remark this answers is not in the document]
PaulKinlan: I should be clearer about this function, it is converting valid mimetypes image/* via the accepts header to something that we can check on drag over. We don't get the file name but we do get the mimetypes on dragEnter/Over.

src/components/app/custom-els/FileDrop/index.ts
PaulKinlan: It really felt wasteful, and also I don't have the actual Data until we're about to drop the file.

src/components/app/custom-els/FileDrop/index.ts
```diff
@@ -0,0 +1,144 @@
...
+    }
+    return true;
+  }
+
+  onDrop(e: DragEvent) {
+    if(this.hasDragData) {
+      e.preventDefault();
+      let dragData = this.firstMatchingFile(e.dataTransfer.files, this.element.accepts);
+      
+      this.element.dispatchEvent(new CustomEvent('filedrop', { detail: dragData }));
```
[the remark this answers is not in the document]
PaulKinlan: re: CustomEvent, I thought we only just started supporting extending the Event and it wasn't support..

I was thinking of re-casting the main drag event, but it means that you get all dragged content, which can include a lot of other mimetypes that aren't files at all.

src/components/app/custom-els/FileDrop/index.ts
PaulKinlan: Yeah, I thought about that although the data structures are different and I had different logic...

src/components/app/custom-els/FileDrop/missing-types.d.ts
PaulKinlan: It's JSX hinting, the build fails without it.

src/components/app/custom-els/FileDrop/styles.css
PaulKinlan: What issue? I don't see it in the previous comments.

src/components/app/custom-els/FileDrop/index.ts
```diff
@@ -0,0 +1,144 @@
...
+    return mimeType.replace(/\*/, '[^/]+' ).replace(/\//g, '\\/');
+  }
+
+  private hasMatchingFile(list: DataTransferItemList, accept: string): boolean {
+    let mimeRegex = this.convertMimeTypeToRegex(accept);
+    let acceptsRegex = new RegExp(mimeRegex);
+
+    for(let i=0; i<list.length; i++) {
+      let item = list[i];
+      if(item.kind === 'file' && acceptsRegex.test(item.type)) {
```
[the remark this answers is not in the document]
PaulKinlan: just reading spec. TIL it can accept can accept extensions.

src/components/app/custom-els/FileDrop/index.ts
```diff
@@ -0,0 +1,144 @@
...
+    this.element.classList.remove(styles.dropValid);
+    this.element.classList.remove(styles.dropInvalid);
+    this.element.classList.remove(styles.dropActive);
+  }
+
+  private convertMimeTypeToRegex(mimeType: string): string {
+    return mimeType.replace(/\*/, '[^/]+' ).replace(/\//g, '\\/');
+  }
+
+  private hasMatchingFile(list: DataTransferItemList, accept: string): boolean {
```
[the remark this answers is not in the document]
PaulKinlan: So. Longer update.

I now realise that I can call getFile on a DataTransferItem, and I can use that reference up until the drop happens (where the actual file contents becomes available). This lets me check to see if we have the correct file mime type at least on drag start...

src/components/app/custom-els/FileDrop/index.ts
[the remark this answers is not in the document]
jakearchibald: I had fun figuring this out. Dunno if it's smarter than it's worth.

```ts
function getMatchingDragItems<T extends File | DataTransferItem>(items: ArrayLike<T>, acceptVal: string): T[] {
  // Split accepts values by ',' then by '/'. Trim everything & lowercase.
  const accepts = acceptVal.split(',').map(accept => {
    return accept.trim().split('/').map(part => part.trim().toLowerCase());
  }).filter(acceptParts => acceptParts.length === 2); // Filter invalid values

  return Array.from(items).filter(item => {
    // Filter out non-file DataTransferItems
    if (item instanceof DataTransferItem && item.kind !== 'file') return false;

    // 'Parse' the type.
    const [typeMain, typeSub] = item.type.split('/').map(s => s.trim().toLowerCase());

    for (const [acceptMain, acceptSub] of accepts) {
      // Look for an exact match, or a partial match if * is accepted, eg image/*.
      if (typeMain === acceptMain && (acceptSub === '*' || typeSub === acceptSub)) {
        return true;
      }
    }

    return false;
  });
}
```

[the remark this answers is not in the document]
jakearchibald: It takes `DataTransferItemList` or `FileList`, and an `accept` attribute value, and returns an array of `DataTransferItem` or `File`.

src/components/app/custom-els/FileDrop/index.ts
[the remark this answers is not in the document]
jakearchibald: I think it'd be much easier to just expose the 'detail' as a property on the element, then fire a generic event.

src/components/app/custom-els/FileDrop/index.ts
jakearchibald: https://github.com/GoogleChromeLabs/squoosh/pull/56#discussion_r197836134 might help

src/components/app/custom-els/FileDrop/missing-types.d.ts
jakearchibald: Ah, of course, I thought it was general typing for the custom event. Sorry! Ignore me.

src/components/app/custom-els/FileDrop/styles.css
jakearchibald: I also can't find it. Hmm.

Anyway, these are Squoosh styles. The component itself probably shouldn't have any styles. Move the squoosh styles out of custom-els and into `app/styles.scss`.

src/components/app/custom-els/FileDrop/index.ts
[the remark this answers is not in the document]
jakearchibald: Huh, TIL `getAsFile`. It just returns null for me, which makes sense from a user privacy perspective.

jakearchibald:
@PaulKinlan let me know when/if this is ready for another review.

src/components/app/custom-els/FileDrop/index.ts
```diff
@@ -0,0 +1,124 @@
+import * as styles from './styles.css';
+import { bind } from '../../../../lib/util';
+
+function convertMimeTypeToRegex(mimeType: string): string {
```
jakearchibald: This still isn't returning a regex

src/components/app/custom-els/FileDrop/index.ts
```diff
@@ -0,0 +1,124 @@
...
+}
+
+export class FileDropEvent extends Event {
+  private _data: FileDropEventInit;
+  constructor(typeArg: string, eventInitDict: FileDropEventInit) {
+    super(typeArg, eventInitDict);
+    this._data = eventInitDict;
+  }
+
+  get data (): FileDropEventInit {
```
jakearchibald: Instead of `data`, make it `file` which returns the file.

src/components/app/custom-els/FileDrop/index.ts
```diff
@@ -0,0 +1,124 @@
...
+    return this._data;
+  }
+}
```
jakearchibald: You also need:

```ts
interface HTMLElementEventMap {
  'filedrop': FileDropEvent;
}
```

This means TypeScript will know that listeners to "filedrop" will receive a `FileDropEvent`.

src/components/app/custom-els/FileDrop/index.ts
```diff
@@ -0,0 +1,124 @@
...
+    }
+    return true;
+  }
+
+  @bind
+  private _onDrop(event: DragEvent) {
+    const file = firstMatchingFile(event.dataTransfer.items, this.accept);
+    if (file) {
+      event.preventDefault();
+      const file = firstMatchingFile(event.dataTransfer.items, this.accept);
```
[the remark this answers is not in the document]
PaulKinlan: mate. not sure how this got in.... fixing it all now

src/components/app/custom-els/FileDrop/index.ts
PaulKinlan: hmm - I'm not sure how to reference this at all.... in missing types, or in the main file of the class.

src/components/app/custom-els/FileDrop/styles.css
PaulKinlan: done.

src/components/app/custom-els/FileDrop/index.ts
PaulKinlan: I think I worked it out.

jakearchibald:
@PaulKinlan btw, you didn't let me know

src/components/app/custom-els/FileDrop/styles.css
```diff
@@ -0,0 +1,7 @@
+file-drop {
+  display: block;
+  overflow: hidden;
+  touch-action: none;
+  height:100%;
+  width:100%;
+}
```
jakearchibald: These don't feel like styles everyone would want, so they shouldn't be here. Perhaps with the exception of `display: block`.

src/components/app/custom-els/FileDrop/index.ts
```diff
@@ -0,0 +1,115 @@
...
+}
+
+export class FileDropEvent extends Event {
+  private _data: FileDropEventInit;
+  constructor(typeArg: string, eventInitDict: FileDropEventInit) {
+    super(typeArg, eventInitDict);
+    this._data = eventInitDict;
+  }
+
+  get file (): File {
```
jakearchibald: Remove space before `()` throughout file.

src/components/app/custom-els/FileDrop/index.ts
```diff
@@ -0,0 +1,115 @@
...
+    this._dragEnterCount++;
+
+    // We don't have data, attempt to get it and if it matches, set the correct state.
+    const dragDataItem = firstMatchingItem(event.dataTransfer.items, this.accept);
+    if (dragDataItem) {
+      this.classList.add('drop-valid');
+    } else {
+      this.classList.add('drop-invalid');
+    }
+    return !!dragDataItem;
```
jakearchibald: Remove the return.

src/components/app/custom-els/FileDrop/index.ts
```diff
@@ -0,0 +1,115 @@
...
+    this._reset();
+  }
+
+  @bind
+  private _onDragOver(event: DragEvent) {
+    const dragDataItem = firstMatchingItem(event.dataTransfer.items, this.accept);
+    if (!dragDataItem) return true;
+
+    event.preventDefault();
+    return false;
```
jakearchibald: Remove the return.

src/components/app/custom-els/FileDrop/index.ts
```diff
@@ -0,0 +1,115 @@
...
+    } else {
+      this.classList.add('drop-invalid');
+    }
+    return !!dragDataItem;
+  }
+
+  @bind
+  private _onDragLeave() {
+    this._dragEnterCount--;
+    if (this._dragEnterCount >= 0) {
```
jakearchibald: This should be `=== 0`

src/components/app/custom-els/FileDrop/index.ts
```diff
@@ -0,0 +1,115 @@
...
+
+  @bind
+  private _onDragEnd() {
+    this._reset();
+  }
+
+  @bind
+  private _onDragOver(event: DragEvent) {
+    const dragDataItem = firstMatchingItem(event.dataTransfer.items, this.accept);
+    if (!dragDataItem) return true;
```
jakearchibald: Just `return`

src/components/app/custom-els/FileDrop/index.ts
```diff
@@ -0,0 +1,115 @@
...
+
+    event.preventDefault();
```
jakearchibald: Move this before the return.

src/components/app/custom-els/FileDrop/index.ts
```diff
@@ -0,0 +1,115 @@
...
+    return false;
+  }
+
+  @bind
+  private _onDrop(event: DragEvent) {
+    const dragDataItem = firstMatchingItem(event.dataTransfer.items, this.accept);
+    if (!dragDataItem) return;
+    const fileData = dragDataItem.getAsFile();
+    if (fileData) {
```
jakearchibald: Flip this to `if (!fileData) return;`

src/components/app/custom-els/FileDrop/index.ts
```diff
@@ -0,0 +1,115 @@
...
+      this.dispatchEvent(new FileDropEvent('filedrop', { file: fileData }));
+      this._reset();
+    }
+    event.preventDefault();
```
jakearchibald: Move to top.

src/components/app/custom-els/FileDrop/index.ts
```diff
@@ -0,0 +1,115 @@
...
+
+    event.preventDefault();
+    return false;
+  }
+
+  @bind
+  private _onDrop(event: DragEvent) {
+    const dragDataItem = firstMatchingItem(event.dataTransfer.items, this.accept);
+    if (!dragDataItem) return;
+    const fileData = dragDataItem.getAsFile();
```
jakearchibald: Rename `fileData` to `file`.

src/components/app/custom-els/FileDrop/index.ts
```diff
@@ -0,0 +1,115 @@
...
+    if (fileData) {
+      this.dispatchEvent(new FileDropEvent('filedrop', { file: fileData }));
```
jakearchibald: Then `{ file: fileData }` to `{ file }`

src/components/app/custom-els/FileDrop/index.ts
```diff
@@ -0,0 +1,115 @@
...
+      this._reset();
+    }
+    event.preventDefault();
+  }
+
+  private _reset() {
+    this._dragEnterCount = 0;
+    this.classList.remove('drop-valid');
+    this.classList.remove('drop-invalid');
+    this.classList.remove('drop-active');
```
jakearchibald: Remove `this.classList.remove('drop-active');`

src/components/app/index.tsx
```diff
@@ -3,6 +3,8 @@ import { bind, bitmapToImageData } from '../../lib/util';
 import * as style from './style.scss';
 import Output from '../output';
 import Options from '../options';
+import {FileDropEvent, FileDrop } from './custom-els/FileDrop';
```
jakearchibald: `import { FileDropEvent, FileDrop } from './custom-els/FileDrop';`

src/components/app/index.tsx
```diff
@@ -93,6 +99,18 @@ export default class App extends Component<Props, State> {
...
     const sourceFile = fileInput.files && fileInput.files[0];
     if (!sourceFile) return;
+    await this.updateFile(sourceFile);
+  }
+
+  @bind
+  async onFileDrop(event: FileDropEvent) {
+    const sourceFile: File = event.file;
+    if (!sourceFile) return;
+    this.setState(initialState());
```
jakearchibald: Create an issue to remove this line and fix the issues it causes.

jakearchibald:
@PaulKinlan Is this ready for another review?

src/components/app/custom-els/FileDrop/index.ts
```diff
@@ -0,0 +1,117 @@
+import { bind } from '../../../../lib/util';
+import './styles.css';
+
+function firstMatchingItem(list: DataTransferItemList, accept: string): DataTransferItem | undefined {
+  const mimeRegexPattern = accept.replace(/\*/, '[^/]+').replace(/\//g, '\\/');
+  const acceptsRegex = new RegExp(mimeRegexPattern);
```
jakearchibald: Change this to avoid regex.

src/components/app/custom-els/FileDrop/index.ts
```diff
@@ -0,0 +1,117 @@
...
+
+interface FileDropEventInit extends EventInit {
+  file: File;
+}
+
+export class FileDropEvent extends Event {
+  private _data: FileDropEventInit;
+  constructor(typeArg: string, eventInitDict: FileDropEventInit) {
+    super(typeArg, eventInitDict);
+    this._data = eventInitDict;
```
jakearchibald: Rename `_data` to `_file`.

src/components/app/custom-els/FileDrop/index.ts
```diff
@@ -0,0 +1,117 @@
...
+    return this.getAttribute('accept') || '';
+  }
+
+  set accept(val: string) {
+    this.setAttribute('accept', val);
+  }
+
+  @bind
+  private _onDragEnter(event: DragEvent) {
+    this._dragEnterCount++;
```
jakearchibald: After this, return if `_dragEnterCount > 1`.

src/components/app/custom-els/FileDrop/index.ts
```diff
@@ -0,0 +1,117 @@
...
+  @bind
+  private _onDragEnd() {
+    this._reset();
+  }
+
+  @bind
+  private _onDragOver(event: DragEvent) {
+    event.preventDefault();
+    const dragDataItem = firstMatchingItem(event.dataTransfer.items, this.accept);
+    if (!dragDataItem) return;
```
jakearchibald: These don't do anything. Just make it prevent default.

src/components/app/index.tsx
PaulKinlan: Issue raised. #63

src/components/app/index.tsx
PaulKinlan: done.

PR #65, opened 2018-06-29
Refactorings

jakearchibald:
This is some stuff I did as I was fixing the second-file-load issue.

Feel free to throw it back at me.

### DIFF
--- a/src/components/app/index.tsx
+++ b/src/components/app/index.tsx
@@ -35,6 +35,28 @@ interface State {
   error?: string;
 }
 
+async function updateCompressedImage(
+  source: SourceImage,
+  encodeData: EncoderState,
+): Promise<ImageBitmap> {
+  // Special case for identity
+  if (encodeData.type === identity.type) return source.bmp;
+
+  const compressedData = await (() => {
+    switch (encodeData.type) {
+      case mozJPEG.type: return mozJPEG.encode(source.data, encodeData.options);
+      default: throw Error(`Unexpected encoder name`);
+    }
+  })();
+
+  const blob = new Blob([compressedData], {
+    type: encoderMap[encodeData.type].mimeType,
+  });
+
+  const bitmap = await createImageBitmap(blob);
+  return bitmap;
+}
+
 export default class App extends Component<Props, State> {
   state: State = {
     loading: false,
@@ -43,15 +65,15 @@ export default class App extends Component<Props, State> {
         encoderState: { type: identity.type, options: identity.defaultOptions },
         loadingCounter: 0,
         loadedCounter: 0,
-        loading: false
+        loading: false,
       },
       {
         encoderState: { type: mozJPEG.type, options: mozJPEG.defaultOptions },
         loadingCounter: 0,
         loadedCounter: 0,
-        loading: false
-      }
-    ]
+        loading: false,
+      },
+    ],
   };
 
   constructor() {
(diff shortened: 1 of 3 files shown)


src/components/app/index.tsx
```diff
@@ -35,6 +35,28 @@ interface State {
...
+    }
+  })();
+
+  const blob = new Blob([compressedData], {
+    type: encoderMap[encodeData.type].mimeType,
+  });
+
+  const bitmap = await createImageBitmap(blob);
+  return bitmap;
+}
```
jakearchibald: This was only using state to set the error message, so I moved that to `updateImage`.

src/components/app/index.tsx
```diff
@@ -95,7 +117,7 @@ export default class App extends Component<Props, State> {
     const { source, images } = this.state;
 
     for (const [i, image] of images.entries()) {
-      if (source !== prevState.source || image !== prevState.images[i]) {
+      if (source !== prevState.source || image.encoderState !== prevState.images[i].encoderState) {
```
jakearchibald: The image only needs updated if the encoder settings have changed (or the source has changed)

src/components/app/index.tsx
```diff
@@ -133,57 +156,53 @@ export default class App extends Component<Props, State> {
...
 
-    image = this.state.images[index];
-    // If a later encode has landed before this one, return.
-    if (loadingCounter < image.loadedCounter) return;
-    image.bmp = result;
-    image.loading = image.loadingCounter !== loadingCounter;
-    image.loadedCounter = loadingCounter;
-    this.setState({ });
-  }
+    this.setState({ images });
```
jakearchibald: I'm trying to avoid editing state & instead setting new state.

src/components/output/index.tsx
```diff
@@ -39,13 +39,17 @@ export default class Output extends Component<Props, State> {
     }
   }
 
+  shouldComponentUpdate(nextProps: Props) {
+    return this.props.leftImg !== nextProps.leftImg || this.props.rightImg !== nextProps.rightImg;
+  }
```
jakearchibald: I did this to reduce calls to render() while I was debugging. Is it worth keeping around?

src/lib/util.ts
```diff
@@ -44,9 +44,9 @@ export async function bitmapToImageData(bitmap: ImageBitmap): Promise<ImageData>
 }
 
 /** Replace the contents of a canvas with the given bitmap */
-export function drawBitmapToCanvas (canvas: HTMLCanvasElement, img: ImageBitmap) {
+export function drawBitmapToCanvas(canvas: HTMLCanvasElement, bitmap: ImageBitmap) {
   const ctx = canvas.getContext('2d');
   if (!ctx) throw Error('Canvas not initialized');
   ctx.clearRect(0, 0, canvas.width, canvas.height);
-  ctx.drawImage(img, 0, 0);
+  ctx.drawImage(bitmap, 0, 0);
```
jakearchibald: Just making the naming consistent with the previous function in this file.

src/components/app/index.tsx
```diff
@@ -35,6 +35,28 @@ interface State {
   error?: string;
 }
 
+async function updateCompressedImage(
```
surma: Can this be called `compressImage`?

src/components/app/index.tsx
surma: Seems like your comment should be a code comment

src/components/app/index.tsx
developit: Was going to suggest that in the other PR, glad its added here.

PR #67, opened 2018-06-29
Tests (via Karmatic)

developit:
I took a second to check if @surma's testing PR could be done with Karmatic, and once I patched that lib to work with Webpack 4 it does!  This has the advantage of using webpack inside Karma, so there's no configuration needed (it inherits our webpack config).


package.json
```diff
@@ -8,7 +8,10 @@
     "build:codecs": "npm run build:mozjpeg_enc",
     "start": "webpack serve --host 0.0.0.0 --hot",
     "build": "webpack -p",
-    "lint": "eslint src"
+    "lint": "eslint src",
```
surma: Can you update this from `master` to use `tslint`?

package.json
```diff
@@ -56,14 +67,19 @@
...
     "file-loader": "^1.1.11",
     "html-webpack-plugin": "^3.0.6",
     "if-env": "^1.0.4",
+    "karmatic": "^1.1.7",
     "loader-utils": "^1.1.0",
     "mini-css-extract-plugin": "^0.3.0",
+    "mocha": "^5.2.0",
     "node-sass": "^4.7.2",
     "optimize-css-assets-webpack-plugin": "^4.0.0",
+    "prettier": "^1.12.1",
```
surma: Is this one from me? I think this can go 😅

surma:
Definitely want @kosamari and  @jakearchibald to take a look before merging, tho.

package.json
developit: must be, haha

package.json
developit: done!

developit:
Updated to fix issues & merge master.

We ended up with a package-lock in here - is that intentional?

surma:
I was surprised we don’t have a `package-lock.json` on master, honestly. For an app, we should always commit it. @jakearchibald wdyt?

developit:
If we commit the lockfile, I think it would be nice to have git set up to diff it without generating conflicts. Something like [npm-merge-driver](https://www.npmjs.com/package/npm-merge-driver) would work.

surma:
I’m gonna close this and get back to this if we get considerable community activity.

PR #69, opened 2018-06-29
Adds Download support. Fixes #47

### LINKED ISSUE #47: Download image link
For each compressed image, we need a button that downloads the compressed image.

PaulKinlan:
New version of downloadUrl that is a lot simpler now.

+ For images that are not encoded by Identity encoder, will create a download url
+ Uses data in the encoder to get file type (note: adds '' to identity encoder)
+ Hoists some of the logic in the updateImage function


developit:
I started to resolve the conflict but it's a tad thick, probably better to leave that for either of you.

jakearchibald:
I'll do it

PaulKinlan:
/me squints.... I'm presuming I missed a large refactor that landed on master...

PaulKinlan:
Looking at this, it looks pretty simple to fix. However, I will be on a flight and it might not have wifi.

jakearchibald:
I've sneaked a fix for a race condition in here. Technically it shouldn't be part of this PR, but feels like it'll cause the least amount of merge conflicts here.

jakearchibald:
This is ready for fresh review.

jakearchibald:
I could have sworn I merged this. Will do it manually.

PR #71, opened 2018-07-01
Adding native encoders

jakearchibald:
Adding the browser's JPEG & PNG encoders. Also added file size indicator.

Things are getting interesting now! This shows how much better MozJPEG is than Chrome's encoder. I also leaned that Chrome's PNG encoder is better than OSX's screenshot tool.

I'll make a PR for adding WebP too, but I'll need to figure out a way to feature test that.

### DIFF
--- a/src/components/app/index.tsx
+++ b/src/components/app/index.tsx
@@ -197,11 +197,13 @@ export default class App extends Component<Props, State> {
       return;
     }
 
+    const bmp = await createImageBitmap(result);
+
     images = this.state.images.slice() as [EncodedImage, EncodedImage];
 
     images[index] = {
       ...images[index],
-      bmp: await createImageBitmap(result),
+      bmp,
       size: result.size,
       loading: image.loadingCounter !== loadingCounter,
       loadedCounter: loadingCounter,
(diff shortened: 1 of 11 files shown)


src/components/app/index.tsx
```diff
@@ -1,4 +1,6 @@
 import { h, Component } from 'preact';
+import { partial } from 'filesize';
```
jakearchibald: I don't know why I can't just import the library. I suspect the types are broken. Will file an issue to investigate and fix.

src/lib/util.ts
```diff
@@ -50,3 +50,20 @@ export function drawBitmapToCanvas(canvas: HTMLCanvasElement, bitmap: ImageBitma
...
+export async function canvasEncode(data: ImageData, type: string, quality?: number) {
+  const canvas = document.createElement('canvas');
+  canvas.width = data.width;
+  canvas.height = data.height;
+  const ctx = canvas.getContext('2d');
+  if (!ctx) throw Error('Canvas not initialized');
+  ctx.putImageData(data, 0, 0);
+
+  const blob = await new Promise(resolve =>
+    // TypeScript doesn't allow the promise to resolve with null for some reason.
```
jakearchibald: I will investigate this and PR TypeScript. I think their types are just broken.

surma: I think TS type inference with the revealing constructor patterns is a bit lacking. If you do `new Promise<Blob | null>(...)` it will work with `null`.

jakearchibald:
Going to split this into two PRs. Stand by.

jakearchibald:
It's now split.

PR #89, opened 2018-07-04
Switch to .mjs for ES Modules.

### LINKED ISSUE #88: Use .mjs for JavaScript modules
Squoosh is mostly using TypeScript (`.ts`), but it also uses `.js` for https://github.com/GoogleChromeLabs/squoosh/blob/master/src/lib/fix-pmc.js despite it being a module.

Let's [follow our own advice](https://developers.google.com/web/fundamentals/primers/modules#mjs) and use `.mjs` to denote JS modules as opposed to classic scripts.

If this is a problem for our current toolchain, it would be good to identify the problem and raise it upstream.

developit:
Fixes #88.


webpack.config.js
```diff
@@ -37,7 +37,7 @@ module.exports = function (_, env) {
       globalObject: 'self'
     },
     resolve: {
-      extensions: ['.ts', '.tsx', '.js', '.jsx', '.scss', '.css'],
+      extensions: ['.ts', '.tsx', '.mjs', '.js', '.scss', '.css'],
```
surma: Just to triple-check: You removed `.jsx` intentionally from the list?

developit: yup.

PR #91, opened 2018-07-04
WASM WebpDecoder

### LINKED ISSUE #75: Integrate WebP
This is a little more involved than MozJPEG, as some browsers will need the decoder too.

* [x] WASM the decoder.
* [ ] Integrate the decoder & enable the encoder for all browsers.

When integrating the decoder, it should be done in a reusable way, so we can do a similar thing for FLIF & AV1.

surma:
This PR adds general decoder infrastructure, but also integrates the wasmified WebP decoder.

Works brilliantly in Chrome. Running into issues with FF, where `await new DecoderWorker()` doesn’t seem to return. @developit plz halp?

But the architecture is ready to review either way :)

Closes #75


src/codecs/webp/decoder.ts
```diff
@@ -0,0 +1,16 @@
+import { blobToArrayBuffer, imageDataToBitmap } from '../../lib/util';
+import DecoderWorker  from './DecoderWorker';
```
developit: the file needs to be called `Decoder.worker.ts` in order to be workerified

src/codecs/decoders.ts
```diff
@@ -0,0 +1,42 @@
...
+  supportedMimeTypes: string[];
+  supportedExtensions: string[];
+}
+
+// We load all decoders and filter out the unsupported ones.
+export const decodersPromise: Promise<Decoder[]> = Promise.all(
+  [
+    browserPNG,
+    browserJPEG,
+    wasmWebP,
```
surma: This is just so the wasm decoder takes precedence of the native one for testing purposes. Will remove before merging.

surma: Or rather: will move it back down before merging.

surma:
So the problem is with the identity decoder in FF. If I use any other encoder, it works fine. I am wondering if FF transfers the data buffer and neuters it where Chrome doesn’t.

surma:
@developit I’m assigning this to you as I have _no_ idea what exactly is going on. I have switched to FF stable and figured out that the exception is thrown when stepping from line 14437 to 14476, which is _really_ odd. My gut says that FF is GC’ing something that shouldn’t be GC’d, but that’s just a hunch.

screen shot 2018-07-05 at 11 47 59 am

Can you take a look? I am not sure if this in an FF bug, a preact bug or me being dumb.

developit:
Investigating again.

developit:
I burned a few hours on this and I have nothing. We're seeing this bug:
https://bugzilla.mozilla.org/show_bug.cgi?id=1439217

I've attempted a myriad of workarounds and nothing works.

src/codecs/browser-jpeg/decoder.ts
```diff
@@ -0,0 +1,15 @@
...
+export async function decode(file: File): Promise<ImageBitmap> {
+  return fileToBitmap(file);
+}
+
+// tslint:disable-next-line:max-line-length It’s a data URL. Whatcha gonna do?
+const jpegFile = 'data:image/jpeg;base64,/9j/4AAQSkZJRgABAQAAAQABAAD/2wBDAAMCAgICAgMCAgIDAwMDBAYEBAQEBAgGBgUGCQgKCgkICQkKDA8MCgsOCwkJDRENDg8QEBEQCgwSExIQEw8QEBD/2wBDAQMDAwQDBAgEBAgQCwkLEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBD/wAARCAABAAEDASIAAhEBAxEB/8QAFQABAQAAAAAAAAAAAAAAAAAAAAn/xAAUEAEAAAAAAAAAAAAAAAAAAAAA/8QAFAEBAAAAAAAAAAAAAAAAAAAAAP/EABQRAQAAAAAAAAAAAAAAAAAAAAD/2gAMAwEAAhEDEQA/AJVAA//Z';
+
+export function isSupported(): Promise<boolean> {
+  return canDecode(jpegFile);
+}
```
jakearchibald: Part of me likes the robustness here, but I think it's overkill. If our target browsers are all capable of decoding a particular format, I don't think we need a feature test for it.

I haven't looked at the rest of the PR yet, so this may be nonsense, but I figured the pattern would be:

1. Look at the first n bytes of the file. (I think this is the best method rather than looking at type or extension. It's certainly what the browser does).
1. If it's a type that isn't universally supported by our target browsers, then:
    1. If this browser doesn't support decoding natively, then return the result of decoding it using the relevant WASM decoder.
1. Return the result of giving it to the native decoder.

src/codecs/browser-webp/decoder.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export const name = 'Browser WebP Decoder';
+export const supportedExtensions = ['webp'];
+export const supportedMimeTypes = ['image/webp'];
+export async function decode(file: File): Promise<ImageBitmap> {
+  return fileToBitmap(file);
+}
+
+// tslint:disable-next-line:max-line-length It’s a data URL. Whatcha gonna do?
+const webpFile = 'data:image/webp;base64,UklGRkAAAABXRUJQVlA4WAoAAAAQAAAAAAAAAAAAQUxQSAIAAAAAAFZQOCAYAAAAMAEAnQEqAQABAAFAJiWkAANwAP79NmgA';
```
jakearchibald: This one from Modernizr seems smaller: `'data:image/webp;base64,UklGRh4AAABXRUJQVlA4TBEAAAAvAAAAAAfQ//73v/+BiOh/AAA='`.

https://github.com/Modernizr/Modernizr/blob/400db4043c22af98d46e1d2b9cbc5cb062791192/feature-detects/img/webp.js#L52

src/codecs/browser-jpeg/decoder.ts
jakearchibald: In terms of the sniffing, all WebPs start with bytes `RIFF` followed by 4 bytes that can be anything, then bytes `WEBPVP8 ` (including the space at the end).

This is what Chrome uses: https://cs.chromium.org/chromium/src/net/base/mime_sniffer.cc?gsn=mask&l=155

src/codecs/webp/Decoder.worker.ts
```diff
@@ -0,0 +1,84 @@
...
+        locateFile(url: string): string {
+          // Redirect the request for the wasm binary to whatever webpack gave us.
+          if (url.endsWith('.wasm')) {
+            return wasmBinaryUrl;
+          }
+          return url;
+        },
+        onRuntimeInitialized() {
+          // An Emscripten is a then-able that, for some reason, `then()`s itself,
+          // causing an infite loop when you wrap it in a real promise. Deleten the `then`
```
jakearchibald: "Deleting"?

src/codecs/webp/decoder.ts
```diff
@@ -0,0 +1,16 @@
...
+export const supportedExtensions = ['webp'];
+export const supportedMimeTypes = ['image/webp'];
+export async function decode(file: File): Promise<ImageBitmap> {
+  const decoder = await new DecoderWorker();
+  const imageData = await decoder.decode(await blobToArrayBuffer(file));
+  return imageDataToBitmap(imageData);
+}
+
+export async function isSupported(): Promise<boolean> {
+  // TODO(@surma): Should we do wasm detection here or something?
```
jakearchibald: Nah. WASM is part of our baseline. If we want to feature-detect this, it'll be right at the start of the app where we may choose to display a helpful message to unsupported browsers.

src/lib/util.ts
```diff
@@ -64,3 +79,26 @@ export async function canvasEncode(data: ImageData, type: string, quality?: numb
...
+    const img = document.createElement('img');
+    img.src = data;
+    img.onload = _ => resolve(true);
+    img.onerror = _ => resolve(false);
+  });
+}
+
+export function fileToBitmap(file: File): Promise<ImageBitmap> {
+  return createImageBitmap(file);
+}
```
jakearchibald: This seems weird, but I think it's worth keeping as I don't think `createImageBitmap` is supported on our target browsers. So, we'll eventually expand this method to fall back to img + canvas or whatever.

src/lib/util.ts
```diff
@@ -64,3 +79,26 @@ export async function canvasEncode(data: ImageData, type: string, quality?: numb
   if (!blob) throw Error('Encoding failed');
   return blob;
 }
+
+export function canDecode(data: string): Promise<boolean> {
```
jakearchibald: Nit: `canDecodeImage`?

src/components/app/index.tsx
```diff
@@ -233,7 +249,7 @@ export default class App extends Component<Props, State> {
       loadedCounter: loadingCounter,
     };
 
-    this.setState({ images, error: '' });
+    this.setState({ images });
```
jakearchibald: Why not clearing the previous error?

src/components/app/index.tsx
```diff
@@ -174,7 +178,12 @@ export default class App extends Component<Props, State> {
   async updateFile(file: File) {
     this.setState({ loading: true });
     try {
-      const bmp = await createImageBitmap(file);
+      const decoder = await findDecoder(file);
+      if(!decoder) {
+        throw new Error('Can’t find a decoder for the given file');
+      }
+      console.log(`Decoding using ${decoder.name}`);
```
jakearchibald: Remove in final PR.

src/codecs/browser-jpeg/decoder.ts
developit: That sounds like a good idea - I don't think Firefox even reports a mimetype for webp files.

jakearchibald: Seems like this logic should sit in `fileToBitmap`.

src/lib/util.ts
surma: That was the point behind this function

src/components/app/index.tsx
surma: I think this got merged into master in the meantime... I don’t remember removing it myself

jakearchibald: It's still there on master

surma:
@jakearchibald @developit PTAL

src/codecs/decoders.ts
```diff
@@ -0,0 +1,58 @@
...
+  'image/png',
+  'image/gif',
+];
+
+export async function decodeFile(file: File): Promise<ImageBitmap> {
+  const mimeType = await sniffMimeType(file);
+  if (!mimeType) {
+    throw new Error('Could not determine mime type');
+  }
+  if (nativelySupportedMimeTypes.includes(mimeType)) {
```
jakearchibald: I still think we're doing this backwards. Instead of whitelisting supported types, we should have a list of types we have wasm decoders for. If the incoming image is one of those types, throw it at that decoder (which may just throw it at the browser's decoder), otherwise, throw it at the browser's decoder.

That means we can support types only supported by particular browsers without effort. Right now this PR prevents loading jpeg2000 even though at least Safari supports that format.

Happy to take a stab at it if I'm explaining badly.

src/codecs/browser-webp/decoder.ts
```diff
@@ -0,0 +1,18 @@
+import { canDecodeImage, createImageBitmapPolyfill } from '../../lib/util';
+
+export const name = 'Browser WebP Decoder';
```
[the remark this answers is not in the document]
surma: I think we should keep it exported for easier debugging (doesn’t cost anything right?).

src/codecs/browser-webp/decoder.ts
```diff
@@ -0,0 +1,18 @@
...
+export const supportedMimeTypes = ['image/webp'];
```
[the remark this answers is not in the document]
surma: It’s used outside this module for the sniffing to find the appropriate codec

src/lib/util.ts
```diff
@@ -80,3 +95,51 @@ export async function canvasEncode(data: ImageData, type: string, quality?: numb
...
+  [/^II*/, 'image/tiff'],
+  [/^MM\x00*/, 'image/tiff'],
+  [/^RIFF....WEBPVP8 /, 'image/webp'],
+]);
+
+export async function sniffMimeType(blob: Blob): Promise<string | undefined> {
+  const firstChunk = await blobToArrayBuffer(blob.slice(0, 1024));
+  const firstChunkString = Array.from(
+    new Uint8Array(firstChunk)).map(v => String.fromCodePoint(v),
+  ).join('');
```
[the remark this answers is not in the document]
surma: My bad

src/lib/util.ts
```diff
@@ -80,3 +95,51 @@ export async function canvasEncode(data: ImageData, type: string, quality?: numb
...
+const magicNumberToMimeType = new Map<RegExp, string>([
+  [/^%PDF-/, 'application/pdf'],
+  [/^GIF87a/, 'image/gif'],
+  [/^GIF89a/, 'image/gif'],
+  [/^\x89PNG\x0D\x0A\x1A\x0A/, 'image/png'],
+  [/^\xFF\xD8\xFF/, 'image/jpeg'],
+  [/^BM/, 'image/bmp'],
+  [/^I I/, 'image/tiff'],
+  [/^II*/, 'image/tiff'],
+  [/^MM\x00*/, 'image/tiff'],
```
mathiasbynens: Fun fact: `\x00` could be just `\0` — it is _not_ an octal escape sequence in JS!

surma:
@jakearchibald PTAL

src/codecs/browser-webp/decoder.ts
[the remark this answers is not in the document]
jakearchibald: Where?

src/codecs/browser-webp/decoder.ts
```diff
@@ -0,0 +1,18 @@
...
+export async function decode(file: File): Promise<ImageBitmap> {
```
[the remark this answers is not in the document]
jakearchibald: This is still an issue.

src/codecs/decoders.ts
```diff
@@ -0,0 +1,47 @@
...
+    if (await decoder.isSupported()) {
+      return decoder;
+    }
+    return null;
+  }),
+// TypeScript is not smart enough to realized that I’m filtering all the falsy
+// values here.
+).then(list => list.filter(item => !!item)) as any as Promise<Decoder[]>;
+
+export async function findDecodersByMimeType(mimeType: string): Promise<Decoder[]> {
```
jakearchibald: This is exported, but I can't see it used anywhere.

src/codecs/webp/decoder.ts
```diff
@@ -0,0 +1,18 @@
+import { blobToArrayBuffer, imageDataToBitmap } from '../../lib/util';
+import DecoderWorker  from './Decoder.worker';
+
+export const name = 'WASM WebP Decoder';
+export const supportedMimeTypes = ['image/webp'];
```
jakearchibald: Exported, but doesn't look like it's used elsewhere.

src/codecs/webp/decoder.ts
```diff
@@ -0,0 +1,18 @@
...
+export async function decode(file: File): Promise<ImageBitmap> {
```
jakearchibald: Blob rather than File?

src/lib/util.ts
```diff
@@ -80,3 +95,53 @@ export async function canvasEncode(data: ImageData, type: string, quality?: numb
...
+  [/^RIFF....WEBPVP8 /, 'image/webp'],
+]);
+
+export async function sniffMimeType(blob: Blob): Promise<string | ''> {
```
jakearchibald: `''` is also a string right?

src/codecs/browser-webp/decoder.ts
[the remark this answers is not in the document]
surma: Ah no, never mind, it’s not used outside anymore :D

src/codecs/decoders.ts
surma: Yeah, exported it for completeness sake. Is exporting considered expensive?

src/lib/util.ts
surma: ...

src/codecs/decoders.ts
jakearchibald: I think it prevents the name being mangled, but it also increases complexity. If I need to make changes to something that's exported, I need to consider its usage across the project. If it's scoped to a single module, I only need to look at that one file.

Same argument for public/private. If it weren't a burden, we'd export everything and make everything public.

src/lib/util.ts
jakearchibald: My point is that `string | ''` is redundant compared to just `string`.

src/codecs/decoders.ts
surma: Mangling is a good argument. Will un-export

src/lib/util.ts
surma: I know. I just don’t like being exposed for the idiot that I am

jakearchibald: 😀 it's one of those mistakes that looks stupid, but given the evolution of the code you can see how it happened. Someone once pulled me up on some code like this:

```ts
const result: boolean = whatever();
if (result) {
  return true;
} else {
  return false;
}
```

"Why not just return the result of `whatever()`" haha yeah oops, I *was* doing stuff in that `if` until a refactor.

surma:
@jakearchibald PTAL

src/codecs/decoders.ts
jakearchibald: This is still exported.

src/lib/util.ts
```diff
@@ -80,3 +95,53 @@ export async function canvasEncode(data: ImageData, type: string, quality?: numb
...
+  [/^RIFF....WEBPVP8 /, 'image/webp'],
+]);
+
+export async function sniffMimeType(blob: Blob): Promise<string> {
+  const firstChunk = await blobToArrayBuffer(blob.slice(0, 16));
+  const firstChunkString =
+    Array.from(new Uint8Array(firstChunk))
+      .map(v => String.fromCodePoint(v))
+      .join('');
+  for (const [detector, mimeType] of magicNumberToMimeType.entries()) {
```
jakearchibald: Nit: `.entries()` isn't needed, it's the default iterator.

surma:
🎉 🎉 🎉 🎉 🎉 🎉 🎉

jakearchibald:
I learned being a thorough dickhead from Anne in spec reviews 😀

PR #93, opened 2018-07-04
show decode errors reliably

developit:


src/components/app/index.tsx
```diff
@@ -153,7 +153,7 @@ export default class App extends Component<Props, State> {
       // changed.
       if (source !== prevState.source || image.encoderState !== prevImage.encoderState) {
         if (prevImage.downloadUrl) URL.revokeObjectURL(prevImage.downloadUrl);
-        this.updateImage(i);
+        this.updateImage(i).catch(console.error);
```
surma: We should write this out. No reason to be smart here (also I think some browsers require a `bind`)

developit: what do you mean? I can factor the error into a method, but this lifecycle method isn't async so I didn't want to use await.

surma: ha no I just meant `.catch(err => console.error(err)`. I’m gonna push a commit.

PR #94, opened 2018-07-04
Fix lint issues resulting from switching to airbnb

developit:


surma:
@developit Can you check if you have a `~/tslint.json`? Ran into a similar issue with Paul Lewis and that’s what was the cause for him.

developit:
Alright, I've committed the fix for the lint scripts. I had to disable the `"import-name"` rule, since it was complaining about ridiculous things - saying all 3 of these imports should be named "Options":

```js
import MozJpegEncoderOptions from '../../codecs/mozjpeg/options';
import BrowserJPEGEncoderOptions from '../../codecs/browser-jpeg/options';
import BrowserWebPEncoderOptions from '../../codecs/browser-webp/options';
```

PR #95, opened 2018-07-04
Add multi-panel component

kosamari:
This PR is the same thing for #24 but against master.

All of code review feedback from #24 should have been addressed, and typescript errors resolved.

### DIFF
--- a/src/components/output/custom-els/MultiPanel/index.ts
+++ b/src/components/output/custom-els/MultiPanel/index.ts
@@ -2,7 +2,7 @@ import './styles.css';
 
 /**
  * A multi-panel view that the user can add any number of 'panels'.
- * 'a panel' consists of two elements. even index element begomes heading,
+ * 'a panel' consists of two elements. Even index element becomes heading,
  * and odd index element becomes the expandable content.
  */
 export default class MultiPanel extends Element {
@@ -24,13 +24,13 @@ export default class MultiPanel extends Element {
   }
 
   // Click event handler
-  _onClick(event: Event) {
+  private _onClick(event: MouseEvent) {
     const el: Element = event.target as HTMLElement;
     this._expand(this._getClosestHeading(el) as HTMLElement);
   }
 
   // KeyDown event handler
-  _onKeyDown(event: Event) {
+  private _onKeyDown(event: Event) {
     const selectedEl = event.target as HTMLElement;
     const keyboardEvent = event as KeyboardEvent;
     // if keydown event is not on heading element, ignore
@@ -81,7 +81,7 @@ export default class MultiPanel extends Element {
     }
   }
 
-  _expand (heading: HTMLElement) {
+  private _expand (heading: HTMLElement) {
 
     if (heading === undefined) {
       return;
@@ -107,7 +107,7 @@ export default class MultiPanel extends Element {
 
   // children of multi-panel should always be even nuber (heading/content pair)
   // if children are odd numbers, add a div at the end to prevent potential error.
-  _childrenChange () {
+  private _childrenChange () {
     if (this.children.length % 2 !== 0) {
       console.error(`detected odd number of elements inside multi-panel,
       please make sure you have heading/content pair`);
@@ -148,7 +148,7 @@ export default class MultiPanel extends Element {
     }
   }
 
-  _getClosestHeading (el: Element) {
+  private _getClosestHeading (el: Element) {
     const closestEl = el.closest('multi-panel > *');
     if (closestEl && closestEl.classList.contains('panel-heading')) {
       return closestEl as HTMLElement;
@@ -157,7 +157,7 @@ export default class MultiPanel extends Element {
   }
 
   // returns headding that is before currently selected one.
-  _prevHeading () {
+  private _prevHeading () {
     // activeElement would be the currently selected headding
     // 2 elemements before that would be the previouse heading unless it is the first element.
     if (this.firstElementChild === document.activeElement) {
@@ -173,7 +173,7 @@ export default class MultiPanel extends Element {
   }
 
   // returns headding that is after currently selected one.
-  _nextHeading () {
+  private _nextHeading () {
     // activeElement would be the currently selected headding
     // 2 elemements after that would be the next heading.
     const nextContent = document.activeElement.nextElementSibling;
@@ -184,13 +184,13 @@ export default class MultiPanel extends Element {
   }
 
   // returns first heading in multi-panel.
-  _firstHeading () {
+  private _firstHeading () {
     // first element is always first headding
     return this.firstElementChild as HTMLElement;
   }
 
   // returns last heading in multi-panel.
-  _lastHeading () {
+  private _lastHeading () {
     // if the last element is headding, return last element
     const lastEl = this.lastElementChild as HTMLElement;
     if (lastEl && lastEl.classList.contains('panel-heading')) {
--- a/src/components/output/custom-els/MultiPanel/index.ts
+++ b/src/components/output/custom-els/MultiPanel/index.ts
@@ -33,7 +33,7 @@ export default class MultiPanel extends HTMLElement {
 
   // Click event handler
   private _onClick(event: MouseEvent) {
-    const el: Element = event.target as Element;
+    const el = event.target as Element;
     const heading = getClosestHeading(el);
     if (!heading) return;
     this._expand(heading);
@@ -43,7 +43,6 @@ export default class MultiPanel extends HTMLElement {
   private _onKeyDown(event: KeyboardEvent) {
     const selectedEl = document.activeElement;
     const heading = getClosestHeading(selectedEl);
-
     // if keydown event is not on heading element, ignore
     if (!heading) return;
 
@@ -70,6 +69,7 @@ export default class MultiPanel extends HTMLElement {
         newHeading = this._lastHeading();
         break;
 
+      // this has 3 cases listed to support IEs and FF before 37
       case 'Enter':
       case ' ':
       case 'Spacebar':
--- a/src/components/output/custom-els/MultiPanel/index.ts
+++ b/src/components/output/custom-els/MultiPanel/index.ts
@@ -33,7 +33,7 @@ export default class MultiPanel extends Element {
   private _onKeyDown(event: Event) {
     const selectedEl = document.activeElement;
     const keyboardEvent = event as KeyboardEvent;
-    const closestHeading = this._getClosestHeading(selectedEl)
+    const closestHeading = this._getClosestHeading(selectedEl);
     // if keydown event is not on heading element, ignore
     if (!closestHeading) {
       return;
@@ -84,17 +84,13 @@ export default class MultiPanel extends Element {
 
   private _expand(heading: HTMLElement) {
 
-    if (heading === undefined) {
-      return;
-    }
+    if (!heading) return;
 
     const content = heading.nextElementSibling;
 
     // heading element should always have nextElementSibling (checked on _childrenChange)
     // but in case it is null, return.
-    if (content === null) {
-      return;
-    }
+    if (!content) return;
 
     // toggle expanded and aria-expanded attributes
     if (content.hasAttribute('expanded')) {
@@ -170,7 +166,6 @@ export default class MultiPanel extends Element {
     if (previousContent) {
       return previousContent.previousElementSibling as HTMLElement;
     }
-    return;
   }
 
   // returns heading that is after currently selected one.
@@ -181,7 +176,6 @@ export default class MultiPanel extends Element {
     if (nextContent) {
       return nextContent.nextElementSibling as HTMLElement;
     }
-    return;
   }
 
   // returns first heading in multi-panel.
@@ -202,7 +196,6 @@ export default class MultiPanel extends Element {
     if (lastContent) {
       return lastContent.previousElementSibling as HTMLElement;
     }
-    return;
   }
 }
 
--- a/src/components/output/custom-els/MultiPanel/index.ts
+++ b/src/components/output/custom-els/MultiPanel/index.ts
@@ -11,8 +11,8 @@ export default class MultiPanel extends Element {
     super();
 
     // add EventListners
-    this.addEventListener('click', this._onClick, false);
-    this.addEventListener('keydown', this._onKeyDown, false);
+    this.addEventListener('click', this._onClick);
+    this.addEventListener('keydown', this._onKeyDown);
 
     // Watch for children changes.
     new MutationObserver(() => this._childrenChange())
@@ -89,13 +89,13 @@ export default class MultiPanel extends Element {
 
     const content = heading.nextElementSibling;
 
-    // heading elment should always have nextElementSibling (checked on _childrenChange)
+    // heading element should always have nextElementSibling (checked on _childrenChange)
     // but in case it is null, return.
     if (content === null) {
       return;
     }
 
-    // toggle expanded and aria-expanded attoributes
+    // toggle expanded and aria-expanded attributes
     if (content.hasAttribute('expanded')) {
       content.removeAttribute('expanded');
       content.setAttribute('aria-expanded', 'false');
@@ -105,7 +105,7 @@ export default class MultiPanel extends Element {
     }
   }
 
-  // children of multi-panel should always be even nuber (heading/content pair)
+  // children of multi-panel should always be even number (heading/content pair)
   // if children are odd numbers, add a div at the end to prevent potential error.
   private _childrenChange() {
     if (this.children.length % 2 !== 0) {
@@ -159,15 +159,15 @@ export default class MultiPanel extends Element {
   // returns heading that is before currently selected one.
   private _prevHeading() {
     // activeElement would be the currently selected heading
-    // 2 elemements before that would be the previouse heading unless it is the first element.
+    // 2 elemements before that would be the previous heading unless it is the first element.
     if (this.firstElementChild === document.activeElement) {
       return this.firstElementChild as HTMLElement;
     }
-    // previouse Element of active Element is previouse Content,
-    // previouse Element of previouse Content  is previouseHeading
-    const previouseContent = document.activeElement.previousElementSibling;
-    if (previouseContent) {
-      return previouseContent.previousElementSibling as HTMLElement;
+    // previous Element of active Element is previous Content,
+    // previous Element of previous Content  is previousHeading
+    const previousContent = document.activeElement.previousElementSibling;
+    if (previousContent) {
+      return previousContent.previousElementSibling as HTMLElement;
     }
     return;
   }
--- a/src/components/output/custom-els/MultiPanel/index.ts
+++ b/src/components/output/custom-els/MultiPanel/index.ts
@@ -7,7 +7,7 @@ import './styles.css';
  */
 export default class MultiPanel extends Element {
 
-  constructor () {
+  constructor() {
     super();
 
     // add EventListners
@@ -19,7 +19,7 @@ export default class MultiPanel extends Element {
       .observe(this, { childList: true });
   }
 
-  connectedCallback () {
+  connectedCallback() {
     this._childrenChange();
   }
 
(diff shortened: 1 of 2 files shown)


kosamari:
cc : @developit If you have moment, would be great to get review... I'm still not sure about how to best write typescript.

developit:
need a day on this one but I shall review!

src/components/output/custom-els/MultiPanel/styles.css
```diff
@@ -0,0 +1,11 @@
+multi-panel > .panel-heading {
```
developit: Think we could go with sass here? I believe the other web components are using it already.

src/components/output/custom-els/MultiPanel/index.ts
```diff
@@ -0,0 +1,208 @@
...
+  _lastHeading () {
+    // if the last element is headding, return last element
+    const lastEl = this.lastElementChild as HTMLElement;
+    if (lastEl && lastEl.classList.contains('panel-heading')) {
+      return lastEl;
+    }
+    // otherwise return 2nd from the last
+    const lastContent = this.lastElementChild;
+    if (lastContent) {
+      return lastContent.previousElementSibling as HTMLElement;
```
developit: If we type the method return (`private _lastHeading(): HTMLElement {`), are we able to drop the `as` casting here? Or was TypeScript saying `previousElementSibling` was some other type?

src/components/output/custom-els/MultiPanel/index.ts
```diff
@@ -0,0 +1,208 @@
...
+      children[0].setAttribute('tabindex', '0');
+    }
+  }
+
+  _getClosestHeading (el: Element) {
+    const closestEl = el.closest('multi-panel > *');
+    if (closestEl && closestEl.classList.contains('panel-heading')) {
+      return closestEl as HTMLElement;
+    }
+    return undefined;
```
developit: The return here seems unnecessary (not sure if it's a TS thing?)

src/components/output/custom-els/MultiPanel/index.ts
```diff
@@ -0,0 +1,208 @@
...
+    if (content.hasAttribute('expanded')) {
+      content.removeAttribute('expanded');
+      content.setAttribute('aria-expanded', 'false');
+    } else {
+      content.setAttribute('expanded', '');
+      content.setAttribute('aria-expanded', 'true');
+    }
+  }
+
+  // children of multi-panel should always be even nuber (heading/content pair)
```
developit: number

src/components/output/custom-els/MultiPanel/index.ts
```diff
@@ -0,0 +1,208 @@
...
+  // if children are odd numbers, add a div at the end to prevent potential error.
+  _childrenChange () {
+    if (this.children.length % 2 !== 0) {
+      console.error(`detected odd number of elements inside multi-panel,
+      please make sure you have heading/content pair`);
+    }
+
+    const children : Element[] = Array.from(this.children);
```
developit: Wondering if this loop could use the tree walking approach used elsewhere in this component? It would also make the child count checking easy:

```js
let heading = this.firstElementChild;

while (heading) {
  const content = heading.nextElementSibling;
  if (!content) {
    return console.error('<multi-panel> requires an even number of element children.');
  }

  // L122 - 141 here

  heading = content.nextElementSibling;
}
```

developit:
all my comments are optional stuff and I'm good if we merge.

src/components/output/custom-els/MultiPanel/index.ts
```diff
@@ -0,0 +1,208 @@
...
+    selectedEl.setAttribute('tabindex', '-1');
+    if (newHeading) {
+      newHeading.setAttribute('tabindex', '0');
+      newHeading.focus();
+    }
+  }
+
+  _expand (heading: HTMLElement) {
+
+    if (heading === undefined) {
```
jakearchibald: `if (!heading) return;` should be enough.

src/components/output/custom-els/MultiPanel/index.ts
```diff
@@ -0,0 +1,208 @@
...
+      return;
+    }
+
+    const content = heading.nextElementSibling;
+
+    // heading elment should always have nextElementSibling (checked on _childrenChange)
+    // but in case it is null, return.
+    if (content === null) {
```
jakearchibald: `if (!content) return;` should be enough.

src/components/output/custom-els/MultiPanel/index.ts
```diff
@@ -0,0 +1,208 @@
...
+      const randomId = Math.random().toString(36).substr(2, 9);
+
+      // if panel-heading/panel-content class is already assigned to elements
+      // (= new child was added earlier), skip the ID/tabindex setting activities
+      // and just flag tabindexAlreadySet
+      if (heading.classList.contains('panel-heading')
+        && content.classList.contains('panel-content')) {
+        tabindexAlreadySet = true;
+        continue;
+      }
```
jakearchibald: Assuming that the tab index is set because some classes are set feels risky. Would it work if you were checking to see if tab index was set?

src/components/output/custom-els/MultiPanel/index.ts
```diff
@@ -0,0 +1,208 @@
...
+    }
+  }
+
+  _getClosestHeading (el: Element) {
+    const closestEl = el.closest('multi-panel > *');
+    if (closestEl && closestEl.classList.contains('panel-heading')) {
+      return closestEl as HTMLElement;
+    }
+    return undefined;
+  }
```
jakearchibald: This function doesn't use any state from the class (no `this.`), so it could just be a function.

src/components/output/custom-els/MultiPanel/index.ts
```diff
@@ -0,0 +1,208 @@
...
+    if (this.firstElementChild === document.activeElement) {
+      return this.firstElementChild as HTMLElement;
+    }
+    // previouse Element of active Element is previouse Content,
+    // previouse Element of previouse Content  is previouseHeadding
+    const previouseContent = document.activeElement.previousElementSibling;
+    if (previouseContent) {
+      return previouseContent.previousElementSibling as HTMLElement;
+    }
+    return;
```
jakearchibald: Don't think you need this return.

src/components/output/custom-els/MultiPanel/index.ts
```diff
@@ -0,0 +1,208 @@
...
+
+  // returns headding that is after currently selected one.
+  _nextHeading () {
+    // activeElement would be the currently selected headding
+    // 2 elemements after that would be the next heading.
+    const nextContent = document.activeElement.nextElementSibling;
+    if (nextContent) {
+      return nextContent.nextElementSibling as HTMLElement;
+    }
+    return;
```
jakearchibald: Don't think this return is needed.

src/components/output/custom-els/MultiPanel/styles.css
```diff
@@ -0,0 +1,11 @@
...
+  background:gray;
+}
+multi-panel > .panel-content {
+  height:0px;
+  overflow:scroll;
+  transition: height 1s;
+}
+multi-panel > .panel-content[expanded] {
+  height:30px;
```
jakearchibald: Why 30?

src/components/output/custom-els/MultiPanel/index.ts
```diff
@@ -0,0 +1,208 @@
...
+
+  _expand (heading: HTMLElement) {
+
+    if (heading === undefined) {
+      return;
+    }
+
+    const content = heading.nextElementSibling;
+
+    // heading elment should always have nextElementSibling (checked on _childrenChange)
```
jakearchibald: `elment` => `element`

src/components/output/custom-els/MultiPanel/index.ts
```diff
@@ -0,0 +1,208 @@
...
+    // but in case it is null, return.
+    if (content === null) {
+      return;
+    }
+
+    // toggle expanded and aria-expanded attoributes
```
jakearchibald: `attoributes` => `attributes`

src/components/output/custom-els/MultiPanel/index.ts
```diff
@@ -0,0 +1,208 @@
...
+
+  _getClosestHeading (el: Element) {
+    const closestEl = el.closest('multi-panel > *');
+    if (closestEl && closestEl.classList.contains('panel-heading')) {
+      return closestEl as HTMLElement;
+    }
+    return undefined;
+  }
+
+  // returns headding that is before currently selected one.
```
jakearchibald: `headding` => `heading` (and other places in this function)

src/components/output/custom-els/MultiPanel/index.ts
```diff
@@ -0,0 +1,208 @@
...
+  _prevHeading () {
+    // activeElement would be the currently selected headding
+    // 2 elemements before that would be the previouse heading unless it is the first element.
+    if (this.firstElementChild === document.activeElement) {
+      return this.firstElementChild as HTMLElement;
+    }
+    // previouse Element of active Element is previouse Content,
+    // previouse Element of previouse Content  is previouseHeadding
+    const previouseContent = document.activeElement.previousElementSibling;
```
jakearchibald: `previouseContent` => `previousContent`

src/components/output/custom-els/MultiPanel/index.ts
```diff
@@ -0,0 +1,208 @@
...
+  }
+
+  // returns headding that is before currently selected one.
+  _prevHeading () {
+    // activeElement would be the currently selected headding
+    // 2 elemements before that would be the previouse heading unless it is the first element.
+    if (this.firstElementChild === document.activeElement) {
+      return this.firstElementChild as HTMLElement;
+    }
+    // previouse Element of active Element is previouse Content,
```
jakearchibald: `previouse` => `previous`

src/components/output/custom-els/MultiPanel/index.ts
```diff
@@ -0,0 +1,208 @@
...
+  // returns headding that is after currently selected one.
+  _nextHeading () {
+    // activeElement would be the currently selected headding
+    // 2 elemements after that would be the next heading.
+    const nextContent = document.activeElement.nextElementSibling;
+    if (nextContent) {
+      return nextContent.nextElementSibling as HTMLElement;
+    }
+    return;
+  }
```
jakearchibald: I'd usually ask for this to be a function (as it doesn't use internal state), but it makes sense to leave it so it's symmetrical with `_prevHeading`.

src/components/output/custom-els/MultiPanel/index.ts
kosamari: yeah it will yell at `not assignable`

src/components/output/custom-els/MultiPanel/index.ts
```diff
@@ -0,0 +1,208 @@
...
+    new MutationObserver(() => this._childrenChange())
+      .observe(this, { childList: true });
+  }
+
+  connectedCallback () {
+    this._childrenChange();
+  }
+
+  // Click event handler
+  _onClick(event: Event) {
```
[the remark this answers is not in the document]
kosamari: when I change this to MouseEvent, then 
L14 `this.addEventListener('click', this._onClick, false);` will error for being not assignable.... :/

src/components/output/custom-els/MultiPanel/styles.css
kosamari: hmmm,  I simply followed PinchZoom and TwoUp which are css.

src/components/output/custom-els/MultiPanel/index.ts
```diff
@@ -0,0 +1,208 @@
...
+        newHeading = this._firstHeading();
+        break;
+
+      case 'End':
+        newHeading = this._lastHeading();
+        break;
+
+      case 'Enter':
+      case ' ':
+      case 'Spacebar':
```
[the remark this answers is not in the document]
kosamari: This is for  IEs and FF before 37.... _just in case_

src/components/output/custom-els/MultiPanel/index.ts
```diff
@@ -0,0 +1,208 @@
...
+
+    event.preventDefault();
+    selectedEl.setAttribute('tabindex', '-1');
+    if (newHeading) {
+      newHeading.setAttribute('tabindex', '0');
+      newHeading.focus();
+    }
+  }
+
+  _expand (heading: HTMLElement) {
```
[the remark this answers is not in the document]
kosamari: yeah, linter didn't yell at me for this. Fixed throughout the code.

src/components/output/custom-els/MultiPanel/index.ts
kosamari: explicitly undefined for https://github.com/GoogleChromeLabs/squoosh/pull/95/files#diff-c97772baa7c313b09c4dac57ba10c96cR37
```if (!this._getClosestHeading(selectedEl)) {
      return;
    }
```

I just wanted to say I'm returning falsey value.

src/components/output/custom-els/MultiPanel/index.ts
kosamari: .... I get your point but then `_prevHeading()` `_firstHeading()` `_lastHeading()` are all in there as method because they use `this`.... and I kind of want to keep them all in one place... is that bad...

src/components/output/custom-els/MultiPanel/index.ts
kosamari: oh wait, you answered my question on previous comment ! :)

src/components/output/custom-els/MultiPanel/styles.css
kosamari: no real reason for 30px. 
This will be the size of the expanded panel which is application design dependent. I was thinking user can override this ? it just need some kind of value for height and this is "placeholder default"

src/components/output/custom-els/MultiPanel/index.ts
kosamari: never mind this was resolved! typescript 🤷‍♀️

kosamari:
Okay, one more try. 
- Reflecting on help in team chat. I _think_ typecasting is cleaner now. 
- `_childrenChange()` has changed quite bit based on bug I run into
  - now use while loop as per @developit 's suggestion
  - it also do clean up of classname and attirubutes
    - if an element is inserted to the top or middle of `<multi-panel>` children,  it will pair heading/content from the top again.

src/components/output/custom-els/MultiPanel/index.ts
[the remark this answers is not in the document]
jakearchibald: Good catch. Can you add a comment about that?

src/components/output/custom-els/MultiPanel/index.ts
```diff
@@ -0,0 +1,220 @@
+import './styles.css';
+
+/**
+ * A multi-panel view that the user can add any number of 'panels'.
+ * 'a panel' consists of two elements. Even index element becomes heading,
+ * and odd index element becomes the expandable content.
+ */
```
jakearchibald: Move this comment above the class - that means things like VSCode will show it along with the type hinting.

src/components/output/custom-els/MultiPanel/index.ts
```diff
@@ -0,0 +1,220 @@
...
+      if (!content) {
+        console.error('<multi-panel> requires an even number of element children.');
+        heading = null;
+        continue;
+      }
+
+      // When odd number of elements were inserted in the middle,
+      // what was heading before may become content after the insertion.
+      // Remove classes and attributes to prepare for this change.
+      if (heading.classList.contains('panel-content')) {
```
surma: What is the `if` for? You can just call `remove()`, right?

src/components/output/custom-els/MultiPanel/index.ts
```diff
@@ -0,0 +1,220 @@
...
+    const selectedEl = document.activeElement;
+    const heading = getClosestHeading(selectedEl);
+
+    // if keydown event is not on heading element, ignore
+    if (!heading) return;
+
+    // don’t handle modifier shortcuts used by assistive technology.
+    if (event.altKey) return;
+
+    let newHeading;
```
jakearchibald: Give this type `HTMLElement | undefined`, just to avoid it having any.

I thought the linter would pick this up since it's part of strict mode, but it isn't. @surma

src/components/output/custom-els/MultiPanel/index.ts
```diff
@@ -0,0 +1,220 @@
...
+        this._expand(heading);
+        break;
+
+      // Any other key press is ignored and passed back to the browser.
+      default:
+        return;
+    }
+
+    event.preventDefault();
+    selectedEl.setAttribute('tabindex', '-1');
```
jakearchibald: If `newHeading` is undefined, doesn't this mean we lose the ability to focus any heading, or have I misread things?

src/components/output/custom-els/MultiPanel/index.ts
```diff
@@ -0,0 +1,220 @@
...
+      if (heading.classList.contains('panel-content')) {
+        heading.classList.remove('panel-content');
+      }
+      if (content.classList.contains('panel-heading')) {
+        content.classList.remove('panel-heading');
+      }
+      if (heading.hasAttribute('expanded') && heading.hasAttribute('aria-expanded')) {
+        heading.removeAttribute('expanded');
+        heading.removeAttribute('aria-expanded');
+      }
```
jakearchibald: If I have an expanded section open, but add another section, it'll close the one I had open. That doesn't seem ideal, but I guess it could be filed as a separate issue.

src/components/output/custom-els/MultiPanel/index.ts
kosamari: heh yeah can just be removed regardless of the check. It was more like... I liked seeing all wrapped in if statements. Fixed :)

src/components/output/custom-els/MultiPanel/index.ts
kosamari: New heading is always set if it made it thus far in the code. On switch statement part, it is wither setting new hading or hitting space bar to open. 

This is wrapped in if statement to make TypeScript happy. yes, it _could_ be undefined but almost certainly not because any keyboard input that does not set new heading is caught in `default` and returned.

kosamari: oh wait, hmmmmm you are right in case of spacebar, it'll take the focus away I guess. moving that line inside if statement.

src/components/output/custom-els/MultiPanel/index.ts
kosamari: I'd like to take this to separate issue. Put some real UI elements inside the panel and see how it should behave along side testing how we should move focus.

kosamari:
HOPEFULLY this is the last one for this PR.
I'm going to create some issue for further consideration.

src/components/output/custom-els/MultiPanel/index.ts
```diff
@@ -0,0 +1,218 @@
...
+  private _onClick(event: MouseEvent) {
+    const el: Element = event.target as Element;
+    const heading = getClosestHeading(el);
+    if (!heading) return;
+    this._expand(heading);
+  }
+
+  // KeyDown event handler
+  private _onKeyDown(event: KeyboardEvent) {
+    const selectedEl = document.activeElement;
```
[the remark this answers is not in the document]
kosamari: added the logic for this in L50:51

src/components/output/custom-els/MultiPanel/styles.css
```diff
@@ -0,0 +1,11 @@
+multi-panel > .panel-heading {
+  background:gray;
+}
+multi-panel > .panel-content {
+  height:0px;
+  overflow:scroll;
+  transition: height 1s;
+}
+multi-panel > .panel-content[expanded] {
+  height:auto;
```
kosamari: I like this approach because it is only attribute toggle & not editing CSS from JavaScript.

However, this means user of this component should only set height of contents _inside_ the `.panel-content` element and not  `.panel-content` itself. 

thoughts ?

```html
<div>header</div>
<div class="panel-content">
  <div id='container'>set height however you like</div>
</div>
```

jakearchibald:
Merrrrrged! 🎉

PR #96, opened 2018-07-10
Fix errors arising from strict .mjs handling

developit:


webpack.config.js
```diff
@@ -49,7 +49,38 @@ module.exports = function (_, env) {
       }
     },
     module: {
+      // Disable the default JavaScript handling:
+      defaultRules: [],
       rules: [
+        {
+          oneOf: [
```
tomayac: @developit Love your spelling of one-off here. Making clear it's really just a one-of.

developit: webpack gets the credit for this one, I'm just writing to their API :P

developit:
A note: we can just remove this if preact ships its `module` export with the `.esm` extension. I've opened a PR there: developit/preact#1165

mathiasbynens:
> A note: we can just remove this if preact ships its module export with the `.esm` extension.

Did you mean `.mjs`?

developit:
yes. haha.

Issue #97, opened 2018-07-10
Let user's paste content into squoosh


PaulKinlan:
Handle onpaste, if there is an image on the clipboard that we can read then replace the current contents of the app.

Open question: Should we alert the user so that they know something destructive is about to happen (unlike drag there is UI visible interaction)

PR #98, opened 2018-07-10
Adds support for onpaste #97

PaulKinlan:
+ file-drop listens for onpaste, if there is data and it matches the
   type it will use that and raise a custom event
+ Adds a new event on file drop 'onfilepaste'
+ App listens for this event and will map to onFilePaste.

Later update.
We no longer user onfilepaste, instead we route through onfiledrop and distinguish via `action`

### DIFF
--- a/src/components/App/custom-els/FileDrop/index.ts
+++ b/src/components/App/custom-els/FileDrop/index.ts
@@ -32,6 +32,7 @@ function getFileData(data: DataTransfer, accept: string): File | null {
 }
 
 interface FileDropEventInit extends EventInit {
+  action: string;
   file: File;
 }
 
@@ -43,11 +44,17 @@ function fixExtendedEvent(instance: Event, type: Function) {
 }
 
 export class FileDropEvent extends Event {
+  private _action: string;
   private _file: File;
   constructor(typeArg: string, eventInitDict: FileDropEventInit) {
     super(typeArg, eventInitDict);
     fixExtendedEvent(this, FileDropEvent);
     this._file = eventInitDict.file;
+    this._action = eventInitDict.action;
+  }
+
+  get action(): string {
+    return this._action;
   }
 
   get file(): File {
@@ -118,19 +125,20 @@ export class FileDrop extends HTMLElement {
   private _onDrop(event: DragEvent) {
     event.preventDefault();
     this._reset();
-
+    const action = 'drop';
     const file = getFileData(event.dataTransfer, this.accept);
     if (file === null) return;
 
-    this.dispatchEvent(new FileDropEvent('filedrop', { file }));
+    this.dispatchEvent(new FileDropEvent('filedrop', { action, file }));
   }
 
   @bind
   private _onPaste(event: ClipboardEvent) {
+    const action = 'paste';
     const file = getFileData(event.clipboardData, this.accept);
     if (file === null) return;
 
-    this.dispatchEvent(new FileDropEvent('filepaste', { file }));
+    this.dispatchEvent(new FileDropEvent('filedrop', { action, file }));
   }
 
   private _reset() {
--- a/src/components/App/custom-els/FileDrop/index.ts
+++ b/src/components/App/custom-els/FileDrop/index.ts
@@ -24,6 +24,13 @@ function firstMatchingItem(list: DataTransferItemList, acceptVal: string): DataT
   });
 }
 
+function getFileData(data: DataTransfer, accept: string): File | null {
+  const dragDataItem = firstMatchingItem(data.items, accept);
+  if (!dragDataItem) return null;
+
+  return dragDataItem.getAsFile();
+}
+
 interface FileDropEventInit extends EventInit {
   file: File;
 }
@@ -112,7 +119,7 @@ export class FileDrop extends HTMLElement {
     event.preventDefault();
     this._reset();
 
-    const file = this._getFileData(event.dataTransfer);
+    const file = getFileData(event.dataTransfer, this.accept);
     if (file === null) return;
 
     this.dispatchEvent(new FileDropEvent('filedrop', { file }));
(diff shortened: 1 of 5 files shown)


src/components/App/custom-els/FileDrop/index.ts
```diff
@@ -110,15 +111,28 @@ export class FileDrop extends HTMLElement {
...
 
+  @bind
+  private _onPaste(event: ClipboardEvent) {
+    const file = this._getFileData(event.clipboardData);
+    if (file === null) return;
+
+    this.dispatchEvent(new FileDropEvent('filepaste', { file }));
+  }
+
+  private _getFileData(data: DataTransfer): File | null {
```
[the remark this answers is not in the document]
PaulKinlan: wrt to interactions: if the UI element can't handle 'onpaste' it should be bubbling up to the file-drop element, if the element does handle `onpaste` then I would expect it to stop propagation and look to see if it can read the data.  I'll do some research to see what is happening with those elements.

wrt to focusable: that's actually a good point, we should probably make the other elements have the ability to focus, tapping outside of this edit box on github to the main page will remove focus of this element, and I think users would expect the same.

wrt to events: my thinking was that if you don't want to handle the event, you can just ignore the `onfilepaste` vs the event will fire and you have to check the event type... that being said, if you look at the internal logic, it's exactly the same so it might imply externally it would be seen the same.

PaulKinlan:
wrt to interactions: if the UI element can't handle 'onpaste' it should be bubbling up to the file-drop element, if the element does handle onpaste then I would expect it to stop propagation and look to see if it can read the data. I'll do some research to see what is happening with those elements.

wrt to focusable: that's actually a good point, we should probably make the other elements have the ability to focus, tapping outside of this edit box on github to the main page will remove focus of this element, and I think users would expect the same.

wrt to events: my thinking was that if you don't want to handle the event, you can just ignore the onfilepaste vs the event will fire and you have to check the event type... that being said, if you look at the internal logic, it's exactly the same so it might imply externally it would be seen the same.

jakearchibald:
> you can just ignore the onfilepaste vs the event will fire and you have to check the event type

It comes down to what we want to be the default. If we want paste to "just work" in places that file drop would, then the developer should have to write extra code to prevent it just working. Right now the developer has to write extra code to make paste work as well as drop.

PaulKinlan:
Yep - I'm easy either way. I think it's a neat side-effect. I'll change it up.

PaulKinlan:
I'll add a 'type' or 'name' to the event so you can disambiguate.

PaulKinlan:
(it won't be `type` ;)

src/components/App/custom-els/FileDrop/index.ts
```diff
@@ -8,7 +8,7 @@ function firstMatchingItem(list: DataTransferItemList, acceptVal: string): DataT
     return accept.trim().split('/').map(part => part.trim());
   }).filter(acceptParts => acceptParts.length === 2); // Filter invalid values
 
-  return Array.from(list).find((item) => {
+  return Array.from<DataTransferItem>(list).find((item) => {
```
jakearchibald: I'm not seeing the issue related to this.

PaulKinlan: My linter in Code keeps highlighting at as red and warning, it's not happening in precheck.. I can remove, but it was mega annoying.

PaulKinlan: I can add an issue to track it if needs be, but I was clearing up a warning on my code instance.

jakearchibald:
@PaulKinlan let me know when this is ready for another review.

PaulKinlan:
Updated: focus - clicking off the element back on to the canvas area will enable `onpaste` events to trigger again (on Chrome OS at least)

PaulKinlan:
Should be ready - unless we want `<input>` elements to propagate.

src/components/App/custom-els/FileDrop/index.ts
jakearchibald: I don't think this is being flagged by one of the project's linters or typescript. I'm not seeing it here.

src/components/App/custom-els/FileDrop/index.ts
```diff
@@ -24,7 +24,15 @@ function firstMatchingItem(list: DataTransferItemList, acceptVal: string): DataT
...
 
+function getFileData(data: DataTransfer, accept: string): File | null {
+  const dragDataItem = firstMatchingItem(data.items, accept);
+  if (!dragDataItem) return null;
+
+  return dragDataItem.getAsFile();
+}
+
 interface FileDropEventInit extends EventInit {
+  action: string;
```
jakearchibald: Make this `'drop' | 'paste'`.

src/components/App/custom-els/FileDrop/index.ts
```diff
@@ -24,7 +24,15 @@ function firstMatchingItem(list: DataTransferItemList, acceptVal: string): DataT
   });
 }
 
+function getFileData(data: DataTransfer, accept: string): File | null {
```
jakearchibald: `firstMatchingItem` returns `undefined` rather than `null`. Make them consistent? Either seems fine.

src/components/App/custom-els/FileDrop/index.ts
```diff
@@ -36,11 +44,17 @@ function fixExtendedEvent(instance: Event, type: Function) {
 }
 
 export class FileDropEvent extends Event {
+  private _action: string;
```
jakearchibald: Make this `'drop' | 'paste'`.

src/components/App/custom-els/FileDrop/index.ts
```diff
@@ -36,11 +44,17 @@ function fixExtendedEvent(instance: Event, type: Function) {
...
   private _file: File;
   constructor(typeArg: string, eventInitDict: FileDropEventInit) {
     super(typeArg, eventInitDict);
     fixExtendedEvent(this, FileDropEvent);
     this._file = eventInitDict.file;
+    this._action = eventInitDict.action;
+  }
+
+  get action(): string {
```
jakearchibald: You can omit the type signature here I think. It's simply enough that typescript will infer it.

src/components/App/custom-els/FileDrop/index.ts
jakearchibald: Actually, you should probably abstract this.

```ts
type FileDropAction = 'drop' | 'paste';
```

jakearchibald:
btw I filed https://bugs.chromium.org/p/chromium/issues/detail?id=863849 to see if we can make it easier to copy images.

PaulKinlan:
REady for another review

PR #99, opened 2018-07-11
Snackbar

developit:
A custom element that exposes a `showSnackbar()` method and handles the message stacking business. Here's what it looks like:

snackbar


src/lib/SnackBar/index.ts
```diff
@@ -0,0 +1,98 @@
...
+    snackbar.show();
+  }
+
+  handleSnackBarRemoved () {
+    this._snackbars = this._snackbars.slice(1);
+    this._processStack();
+  }
+}
+
+customElements.define('snack-bar', SnackBarElement);
```
jakearchibald: Our other custom elements are in a custom elements folder. Do the same for consistency?

src/components/App/index.tsx
```diff
@@ -104,6 +105,8 @@ export default class App extends Component<Props, State> {
     ],
   };
 
+  private snackbar?: SnackBarElement = undefined;
```
jakearchibald: You don't need the `= undefined` fwiw.

src/lib/SnackBar/index.ts
```diff
@@ -0,0 +1,98 @@
...
+
+  _processStack () {
+    if (this._snackbars.length === 0) return;
+    const snackbar = this._snackbars[0];
+    snackbar.onhide = this.handleSnackBarRemoved.bind(this);
+    snackbar.show();
+  }
+
+  handleSnackBarRemoved () {
+    this._snackbars = this._snackbars.slice(1);
```
surma: `shift()`?

src/lib/SnackBar/index.ts
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  remove () {
+    this.cancelTimer();
+    this.parent.removeChild(this.element);
+    if (this.onhide) this.onhide();
+  }
+}
+
+export default class SnackBarElement extends HTMLElement {
```
surma: Kinda wondering... why is this a Custom Element? Seems like it should mostly be a module with a `showSnackbar()` function?

src/components/App/index.tsx
```diff
@@ -233,10 +235,15 @@ export default class App extends Component<Props, State> {
       loadedCounter: loadingCounter,
     };
 
-    this.setState({ images, error: '' });
+    this.setState({ images });
+  }
+
+  showError (error: Error | String) {
```
jakearchibald: Are we providing this function with error objects anywhere? If not, we don't need to accept it.

Also, `string` rather than `String`.

src/components/App/index.tsx
```diff
@@ -269,7 +276,7 @@ export default class App extends Component<Props, State> {
             />
           ))}
           {anyLoading && <span style={{ position: 'fixed', top: 0, left: 0 }}>Loading...</span>}
-          {error && <span style={{ position: 'fixed', top: 0, left: 0 }}>Error: {error}</span>}
+          <snack-bar ref={(c) => { this.snackbar = c as SnackBarElement; }} />
```
jakearchibald: Didn't you create a helper for this?

src/lib/SnackBar/index.ts
jakearchibald: Yeah, as it stands it doesn't make sense as a custom element. It would make more sense if the snack bar items appeared within the custom element, but if that isn't useful a simple module/class could work.

src/lib/SnackBar/index.ts
```diff
@@ -0,0 +1,98 @@
...
+  private _snackbars: SnackBar[] = [];
```
jakearchibald: Right now a snackbar contains snackbars, but they're different things. Either `Snackbar` needs to become something like `SnackbarItem` or `SnackBarElement` needs to become `AdmiralSnackbar` or something.

src/lib/SnackBar/index.ts
```diff
@@ -0,0 +1,98 @@
...
+    this.element.removeAttribute('aria-hidden');
+    this.closeTimer = setTimeout(this.hide.bind(this), this.options.timeout || DEFAULT_TIMEOUT);
+  }
+
+  hide () {
+    if (!this.showing) return;
+    this.showing = false;
+    this.cancelTimer();
+    this.element.addEventListener('animationend', this.remove.bind(this));
+    this.element.addEventListener('transitionend', this.remove.bind(this));
```
jakearchibald: Why both `animationend` and `transitionend`?

src/lib/SnackBar/index.ts
```diff
@@ -0,0 +1,98 @@
...
+    this.element.setAttribute('aria-hidden', 'true');
```
jakearchibald: Do you need to set this since you're removing the element?

src/lib/SnackBar/index.ts
```diff
@@ -0,0 +1,98 @@
...
+  message: string;
+  timeout?: number;
+  actionText?: string;
+  actionHandler?: () => boolean | null;
+}
+
+class SnackBar {
+  on?: () => void;
+  onhide?: () => void;
+  private options: SnackBarOptions;
```
jakearchibald: Start private properties `_`.

src/lib/SnackBar/index.ts
```diff
@@ -0,0 +1,98 @@
...
+
+export interface SnackBarOptions {
+  message: string;
+  timeout?: number;
+  actionText?: string;
+  actionHandler?: () => boolean | null;
+}
+
+class SnackBar {
+  on?: () => void;
```
jakearchibald: Should this be optional given it has a default value?

src/lib/SnackBar/index.ts
```diff
@@ -0,0 +1,98 @@
...
+  private closeTimer?: number;
+
+  constructor (options: SnackBarOptions, parent: Element) {
+    this.options = options;
+    this.parent = parent;
+
+    this.element.className = 'snackbar';
+    this.element.setAttribute('aria-live', 'assertive');
+    this.element.setAttribute('aria-atomic', 'true');
+    this.element.setAttribute('aria-hidden', 'true');
```
jakearchibald: I don't think we need this, since the presence in the dom is controlling visibility.

src/lib/SnackBar/index.ts
```diff
@@ -0,0 +1,98 @@
...
+
+    this.text.className = 'snackbar--text';
+    this.text.textContent = options.message;
+    this.element.appendChild(this.text);
+
+    if (options.actionText) {
+      this.button.className = 'snackbar--button';
+      this.button.textContent = options.actionText || '';
```
jakearchibald: `options.actionText` cannot be falsey at this point.

src/lib/SnackBar/index.ts
```diff
@@ -0,0 +1,98 @@
...
+      this.button.addEventListener('click', (event) => {
+        if (options.actionHandler && options.actionHandler() === false) return;
+        this.hide();
```
jakearchibald: Since hide is async, it's possible to click this button multiple times. Do we want that?

src/lib/SnackBar/index.ts
```diff
@@ -0,0 +1,98 @@
...
+    this.cancelTimer();
+    this.element.addEventListener('animationend', this.remove.bind(this));
+    this.element.addEventListener('transitionend', this.remove.bind(this));
+    this.element.setAttribute('aria-hidden', 'true');
+  }
+
+  remove () {
+    this.cancelTimer();
+    this.parent.removeChild(this.element);
+    if (this.onhide) this.onhide();
```
jakearchibald: Seems like `onhide` is more like `onremove`.

src/lib/SnackBar/index.ts
developit: I'd originally had the CE provide a root. Do we think that might be useful? If so it would justify the CE.

src/lib/SnackBar/index.ts
developit: Yes, but I don't think it makes sense to nest the element dirs inside the component dirs since they're n:n

src/components/App/index.tsx
developit: without it, typescript complains that the property is uninitialized. It has to be defined either here or in the constructor, and there's no useful value to set it to in the constructor

src/components/App/index.tsx
developit: I had thought so but I likely changed it after.

src/lib/SnackBar/index.ts
developit: I think I was relying on referential inequality before, nothing is now though.

src/lib/SnackBar/index.ts
developit: The removal happens at the end of the animation, which is actually a CSS animation triggered by this attribute.

src/lib/SnackBar/index.ts
developit: I'm not sure what the `on?` was from, that seemed like a typo. `onhide` is an optional hook, it is only invoked if it exists.

src/lib/SnackBar/index.ts
developit: It's used to control the animation.

src/lib/SnackBar/index.ts
developit: There's a boolean `_showing` flag checked in hide() to prevent this. I've added the same guard to actionHandler invocation though.

src/lib/SnackBar/index.ts
developit: indeed, I'll fix the name

src/lib/SnackBar/index.ts
developit: I hadn't written the CSS yet. Removed.

src/lib/SnackBar/index.ts
developit: It seems like I should rename the CE to `<snack-bar-manager>` or something along those lines. The name should imply that it's the root in which snackbars are displayed.

src/components/App/index.tsx
developit: separate PR.

developit:
Comments address except for the CE's naming - I'd suggest `<snack-bar-manager>`... or maybe `SnackBar` and `Snack`? I'm twisting that metaphor though..

src/components/App/index.tsx
```diff
@@ -246,10 +248,15 @@ export default class App extends Component<Props, State> {
       loadedCounter: loadingCounter,
     };
 
-    this.setState({ images, error: '' });
+    this.setState({ images });
+  }
+
+  showError (error: string) {
+    if (!this.snackbar) return;
```
jakearchibald: Nit: Throw instead? This shouldn't happen, right?

src/lib/SnackBar/missing-types.d.ts
```diff
@@ -0,0 +1,13 @@
+import { SnackBarOptions } from '.';
+
+declare global {
+  namespace JSX {
+    interface IntrinsicElements {
+      'snack-bar': SnackBarAttributes;
+    }
+
+    interface SnackBarAttributes extends HTMLAttributes {
+      showSnackbar?: (options: SnackBarOptions) => void;
```
jakearchibald: This doesn't look like an attribute, or am I missing something?

src/lib/SnackBar/index.ts
```diff
@@ -0,0 +1,96 @@
...
+  }
+}
+
+export default class SnackBarElement extends HTMLElement {
+  private _snackbars: SnackBar[] = [];
+
+  showSnackbar (options: SnackBarOptions) {
+    const snackbar = new SnackBar(options, this);
+    this._snackbars.push(snackbar);
+    this._processStack();
```
jakearchibald: It seems a bit indirect that the child manages the parent. Wouldn't it make more sense for the `SnackBarElement` to manage the adding/removing of its content?

In that case, the `SnackBar` class doesn't need any methods, it can just be a props-like object, and `SnackBarElement` would create elements for them as needed.

@surma second opinion?

jakearchibald:
Also, what do you think about `showSnackbar` returning a promise that resolves with the action? It only ever happens once right, so it feels more like a promise than a callback.

src/components/App/index.tsx
developit: I guess it could happen if an error occurred prior to rendering, but it doesn't matter to me

developit:
@jakearchibald Alright I took a crack at addressing those things.  I didn't end up gutting `Snack{}`, but did remove the parent reference.

src/lib/SnackBar/index.ts
```diff
@@ -0,0 +1,97 @@
...
+export default class SnackBarElement extends HTMLElement {
+  private _snackbars: Snack[] = [];
+
+  showSnackbar (options: SnackOptions) {
+    const snack = new Snack(options);
+    this._snackbars.push(snack);
+    this._processStack();
+  }
+
+  private _processStack () {
```
surma: async?

src/lib/SnackBar/index.ts
```diff
@@ -0,0 +1,97 @@
...
+  }
+}
+
+export default class SnackBarElement extends HTMLElement {
+  private _snackbars: Snack[] = [];
+
+  showSnackbar (options: SnackOptions) {
+    const snack = new Snack(options);
+    this._snackbars.push(snack);
+    this._processStack();
```
surma: Not 100% sure, but shouldn’t we make sure that the stack isn’t already being processed from a previous call to `showSnackbar`?

developit: ah yup, good catch.

src/lib/SnackBar/index.ts
developit: indeed.

src/lib/SnackBar/missing-types.d.ts
```diff
@@ -0,0 +1,13 @@
+import { SnackOptions } from '.';
+
+declare global {
+  namespace JSX {
+    interface IntrinsicElements {
+      'snack-bar': SnackBarAttributes;
+    }
+
+    interface SnackBarAttributes extends HTMLAttributes {
+      showSnackbar?: (options: SnackOptions) => void;
```
jakearchibald: This isn't an attribute, is it? I don't think it shows up on the JSX at all right?

src/lib/SnackBar/index.ts
jakearchibald: +1

PR #100, opened 2018-07-16
Vertical two-up

jakearchibald:
The designs show the two-up working vertically on mobile, so I implemented that.

### DIFF
--- a/src/components/Output/custom-els/TwoUp/index.ts
+++ b/src/components/Output/custom-els/TwoUp/index.ts
@@ -2,14 +2,14 @@ import * as styles from './styles.css';
 import { PointerTracker, Pointer } from '../../../../lib/PointerTracker';
 
 const legacyClipCompatAttr = 'legacy-clip-compat';
-const verticalAttr = 'vertical';
+const orientationAttr = 'orientation';
 
 /**
  * A split view that the user can adjust. The first child becomes
  * the left-hand side, and the second child becomes the right-hand side.
  */
 export default class TwoUp extends HTMLElement {
-  static get observedAttributes() { return [verticalAttr]; }
+  static get observedAttributes() { return [orientationAttr]; }
 
   private readonly _handle = document.createElement('div');
   /**
@@ -62,7 +62,7 @@ export default class TwoUp extends HTMLElement {
   }
 
   attributeChangedCallback(name: string) {
-    if (name === verticalAttr) {
+    if (name === orientationAttr) {
       this._resetPosition();
     }
   }
@@ -71,7 +71,7 @@ export default class TwoUp extends HTMLElement {
     // Set the initial position of the handle.
     requestAnimationFrame(() => {
       const bounds = this.getBoundingClientRect();
-      this._position = (this.vertical ? bounds.height : bounds.width) / 2;
+      this._position = (this.orientation === 'vertical' ? bounds.height : bounds.width) / 2;
       this._setPosition();
     });
   }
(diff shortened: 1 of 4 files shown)


src/components/Output/custom-els/TwoUp/index.ts
```diff
@@ -1,14 +1,15 @@
...
+const legacyClipCompatAttr = 'legacy-clip-compat';
+const verticalAttr = 'vertical';
 
 /**
  * A split view that the user can adjust. The first child becomes
  * the left-hand side, and the second child becomes the right-hand side.
  */
 export default class TwoUp extends HTMLElement {
-  static get observedAttributes () { return [legacyClipCompat]; }
+  static get observedAttributes() { return [verticalAttr]; }
```
jakearchibald: I had `legacyClipCompat` here previously, but it was a mistake.

src/components/Output/custom-els/TwoUp/index.ts
```diff
@@ -52,39 +53,64 @@ export default class TwoUp extends HTMLElement {
...
+    });
+  }
+
   /**
    * If true, this element works in browsers that don't support clip-path (Edge).
    * However, this means you'll have to set the height of this element manually.
    */
-  get noClipPathCompat () {
-    return this.hasAttribute(legacyClipCompat);
+  get legacyClipCompat() {
```
jakearchibald: I think I renamed the attr but forgot to rename the prop.

src/components/Output/custom-els/TwoUp/index.ts
```diff
@@ -52,39 +53,64 @@ export default class TwoUp extends HTMLElement {
...
+      this.removeAttribute(legacyClipCompatAttr);
+    }
+  }
+
+  /**
+   * Split vertically rather than horizontally.
+   */
+  get vertical() {
+    return this.hasAttribute(verticalAttr);
+  }
```
jakearchibald: There's repetition here, but I'm not sure if there's a typescript-friendly way to solve it.

I'd usually transform the attr names to property names then generate the getters/setters automatically.

src/components/Output/custom-els/TwoUp/index.ts
```diff
@@ -1,14 +1,15 @@
 import * as styles from './styles.css';
 import { PointerTracker, Pointer } from '../../../../lib/PointerTracker';
 
-const legacyClipCompat = 'legacy-clip-compat';
+const legacyClipCompatAttr = 'legacy-clip-compat';
+const verticalAttr = 'vertical';
```
PaulKinlan: Is this meant to be like the disabled attribute, where by presence indicates the state? I was wondering why this wasn't just an `orientation` attribute with a type that has a default of horizontal, but the option of 'vertical'

src/components/Output/custom-els/TwoUp/index.ts
surma: We could write a decorator for tying attrs with props. Not sure if it’s worth it (yet).

jakearchibald: If we feel we need it, we should talk to the polymer folks. Pretty sure they'll have thought about this before.

jakearchibald:
@PaulKinlan @surma Ready for further review.

I've changed this to be an `orientation` property/attribute with `"vertical"` and `"horizontal"` values.

Setting `el.orientation` just sets the attribute.

Getting `el.orientation` will always return `"vertical"` or `"horizontal"`. This seems to be how `input.type` behaves.

PR #103, opened 2018-07-17
Basic webp encoding integration

jakearchibald:
Right now `webp/Encoder.worker.ts` is a copy/paste job that's crying out for abstraction, but I feel that might be wasted work as we start adding options to encoders.

I've started playing with C++ and emscripten bindings https://github.com/GoogleChromeLabs/squoosh/compare/webp-encode-integrate...webp-encode-options - it feels like this might be a better way forward, especially as we need to pass option structs into the encoder (and perhaps the other way, for getting default options).

I don't have strong feelings when it comes to merging this with the copy-paste job, and filing a v1 issue to fix that across codecs, or blocking this PR on it.

### DIFF
--- a/codecs/webp_enc/webp_enc.d.ts
+++ b/codecs/webp_enc/webp_enc.d.ts
@@ -1,0 +1,1 @@
+export default function(opts: EmscriptenWasm.ModuleOpts): EmscriptenWasm.Module;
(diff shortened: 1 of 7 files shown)


jakearchibald:
Btw, our wasm encoder is much slower than the built in browser one. I guess that's due to a lack of threads and SIMD. Might be a good module to track the performance of WASM/emscripten over time.

PR #110, opened 2018-07-20
WebP encode options

jakearchibald:
I've integrated the options for WebP! There was some flailing here, so reviewing is much appreciated.

screen shot 2018-07-20 at 16 15 36


src/codecs/webp/Encoder.worker.ts
```diff
@@ -1,4 +1,4 @@
-import webp_enc from '../../../codecs/webp_enc/webp_enc';
+import webp_enc, { WebPModule } from '../../../codecs/webp_enc/webp_enc';
```
surma: WOAH TIL

src/codecs/webp/encoder.ts
```diff
@@ -1,13 +1,76 @@
...
 
-export interface EncodeOptions { quality: number; }
+export enum WebPImageHint {
+  WEBP_HINT_DEFAULT, // default preset.
+  WEBP_HINT_PICTURE, // digital picture, like portrait, inner shot
+  WEBP_HINT_PHOTO,   // outdoor photograph, with natural lighting
+  WEBP_HINT_GRAPH,   // Discrete tone image (graph, map-tile etc).
+}
+
+export interface EncodeOptions {
```
surma: If at all possible, we should link to a resource somewhere explaining what these options are (maybe even from the UI?)

src/codecs/webp/options.tsx
```diff
@@ -1,3 +1,317 @@
...
+        Number((form.quality as HTMLInputElement).value),
+      // In lossless mode, the method is derived from the preset.
+      method: lossless ?
+        losslessPresets[Number(losslessPresetInput.value)][0] :
+        Number((form.method_input as HTMLInputElement).value),
+      image_hint: (form.image_hint as HTMLInputElement).checked ?
+        WebPImageHint.WEBP_HINT_GRAPH :
+        WebPImageHint.WEBP_HINT_DEFAULT,
+      // .checked
+      exact: Number((form.exact as HTMLInputElement).checked),
```
surma: I know it would be a 1-line function, but to reduce visual noise may be add a `isChecked()` function?

jakearchibald: Like this?

```ts
function fieldCheckedAsNumber(field: any): number {
  return Number((field as HTMLInputElement).checked);
}
```

src/codecs/webp/encoder.ts
jakearchibald: I'll add a comment saying which files have details. There's the header which has some info, and the help output which has slightly different info.

Related: we should find a way to add info into the UI. I guess it's a design job.

jakearchibald: Doh, you said UI too. Agreed. I think this is important. Want me to add details for this PR or create an issue.

I'll detail which files the headers are in as part of this PR.

src/codecs/webp/options.tsx
jakearchibald: I guess you want a similar function for `.value`?

jakearchibald: It abstracts away the awkward casting, so sgtm.

src/codecs/webp/encoder.ts
jakearchibald: Filed https://github.com/GoogleChromeLabs/squoosh/issues/111

surma:
🎉

PR #115, opened 2018-07-30
Implement image quantization via libimagequant

surma:
This PR adds a wasm’d version of [libimagequant](https://github.com/ImageOptim/libimagequant), the low-level library that libpngquant uses.

**DON’T MERGE YET**


surma:
@jakearchibald @developit I am a bit lost as to how to integrate the quantizer options in the UI. `<Options>` seems to be currently specific to `EncoderState`, so should I create a new component á la `<Options>` for the quantizer? What would y’all prefer?

jakearchibald:
If it's just a number, I think each of the WebP and PNG options could just add a range to their options. If there are multiple options we could create a component for the quantisation options for the WebP and PNG options to include.

surma:
It’s 2 numbers currently (`maxNumColors` and `dither`). I thought it’d be preferrable to avoid entangling quantization and whatever encoder your choose. Yes, quantization won’t do a lot of good with JPEG, but that could be an interesting discover to make for the user themselves. 

My idea was to add a quantizer section at the top of the options, no matter which encoder you take with an “enable” checkbox and the 2 options above. WDYT?

developit:
@surma that UI is what I was picturing.

jakearchibald:
I guess the question is: do we want to provide palette reduction to encoders that don't typically benefit from it?

src/components/App/index.tsx
```diff
@@ -59,23 +60,28 @@ const filesize = partial({});
...
   source: SourceImage,
+  quantizerOptions: quantizer.QuantizeOptions | null,
   encodeData: EncoderState,
 ): Promise<File> {
   // Special case for identity
   if (encodeData.type === identity.type) return source.file;
 
+  let sourceData = source.data;
+  if (quantizerOptions) {
+    sourceData = await quantizer.quantize(sourceData, quantizerOptions);
```
developit: Thought: would it be easier to mix the quantizer options into EncoderState and then peel them off here?

developit:
@jakearchibald maybe that could be shown as more of a hint in the UI rather than enabling/disabling quant?

src/components/App/index.tsx
surma: In terms of naming that would make sense. I might give that version a spin tomorrow.

surma:
> do we want to provide palette reduction to encoders that don't typically benefit from it?

I’d vote for yes. That’s a really nice lesson to learn for the user.

jakearchibald:
Ok, so we're treating quantizing as a modification separate to compression.

It feels like it should go above the codec selection on each side, and be completely separate to the codec options.

Well, not quite completely, as it shouldn't be available if "original image" is selected.

I guess we could add other things in there later like "resize".

jakearchibald:
Like @developit, I thought we'd mix this into encoder state, but if we're creating the concept of "preprocessors" then it feels like it should be a new property of `EncodedImage`.

The preprocessor would have its own "options" object, an options component, and a function that takes an image & options and spits out the transformed image. Kinda similar to the encoder but it's just returning RGBA data.

This will be useful in future as we can cache this step, so if an encoder-related setting changes, we don't need to reprocess the image.

A quick-win "preprocessor" we can also add is "background colour", which is useful for removing alpha from an image in a controlled way.

surma:
Alright peeps, here we go:
squoosh quant

(Ironically, the palette/dithering of the GIF kinda hides the effect 🙄)

@jakearchibald @developit PTAL

src/components/Options/index.tsx
```diff
@@ -61,25 +67,68 @@ export default class Options extends Component<Props, State> {
...
+    const el = event.currentTarget as HTMLInputElement;
+
+    const preprocessorState = this.props.preprocessorState;
+    const preprocessor = el.name.split('.')[0] as keyof typeof preprocessorState;
+    preprocessorState[preprocessor].enabled = el.checked;
+    this.props.onPreprocessorOptionsChange(preprocessorState);
+  }
+
+  @bind
+  onQuantizerOptionsChange(opts: QuantizeOptions) {
```
surma: There’s probably a nice way of writing a generic `onPreprocessorOptionsChange` handler, but I’d liek to address that in a separate PR.

jakearchibald:
I love this. Will review.

src/codecs/imagequant/quantizer.ts
```diff
@@ -0,0 +1,17 @@
+import QuantizerWorker from './Quantizer.worker';
+
+export const name = 'Image Quanitzer';
+export async function quantize(data: ImageData, opts: QuantizeOptions): Promise<ImageData> {
+  const quantizer = await new QuantizerWorker();
```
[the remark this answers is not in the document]
surma: AFAIU, `comlink-loader` puts all of these things in a single worker. cc @developit to confirm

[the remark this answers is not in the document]
jakearchibald: Whoa, so this is in the same worker as our WebP encoder? Isn't that bad for loading?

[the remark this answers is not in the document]
surma: What kind of loading? THe code is still loaded lazily, but into the same worker.

surma:
@PaulKinlan @jakearchibald PTAL

src/components/Options/index.tsx
surma: #118

Issue #121, opened 2018-08-02
Enabling quantization on one side enables it on the other


jakearchibald:
It shouldn't.

Issue #120, opened 2018-08-01
Preprocessing and "original image" don't mix


jakearchibald:
They appear to work (takes up processing time) but the output doesn't change.

We need to either apply the preprocessing to the output, which may be weird since it's no longer the "original image", or hide the preprocessing options.

My vote goes for hiding.

PR #122, opened 2018-08-02
Quant fixes

jakearchibald:
Fixes https://github.com/GoogleChromeLabs/squoosh/issues/121 and https://github.com/GoogleChromeLabs/squoosh/issues/120.

### DIFF
--- a/src/components/Options/index.tsx
+++ b/src/components/Options/index.tsx
@@ -125,14 +125,12 @@ export default class Options extends Component<Props, State> {
           />
           Enable
         </label>
-        {preprocessorState.quantizer.enabled ? (
+        {preprocessorState.quantizer.enabled &&
           <QuantizerOptionsComponent
             options={preprocessorState.quantizer}
             onChange={this.onQuantizerOptionsChange}
           />
-        ) : (
-          <div/>
-        )}
+        }
         <hr/>
         <label>
           Mode:
--- a/src/components/Options/index.tsx
+++ b/src/components/Options/index.tsx
@@ -81,18 +81,15 @@ export default class Options extends Component<Props, State> {
   onPreprocessorEnabledChange(event: Event) {
     const el = event.currentTarget as HTMLInputElement;
 
-    const newPreprocessorState = {
+    const preprocessor = el.name.split('.')[0] as keyof PreprocessorState;
+
+    this.props.onPreprocessorOptionsChange({
       ...this.props.preprocessorState,
-    };
-
-    const preprocessor = el.name.split('.')[0] as keyof typeof newPreprocessorState;
-
-    newPreprocessorState[preprocessor] = {
-      ...newPreprocessorState[preprocessor],
-      enabled: el.checked,
-    };
-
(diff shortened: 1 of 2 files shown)


src/components/Options/index.tsx
```diff
@@ -107,25 +112,27 @@ export default class Options extends Component<Props, State> {
...
-          Enable
-        </label>
-        {preprocessorState.quantizer.enabled ? (
-          <QuantizerOptionsComponent
-            options={preprocessorState.quantizer}
-            onChange={this.onQuantizerOptionsChange}
-          />
-        ) : (
-          <div/>
+        {encoderState.type !== 'identity' && (
```
developit: Not saying it needs to be in this PR, but do you think it'd be worth having a property on each encoder indicating whether we should show/allow quantization as a preprocessing step?

src/components/Options/index.tsx
```diff
@@ -81,10 +81,15 @@ export default class Options extends Component<Props, State> {
   onPreprocessorEnabledChange(event: Event) {
     const el = event.currentTarget as HTMLInputElement;
 
-    const preprocessorState = this.props.preprocessorState;
-    const preprocessor = el.name.split('.')[0] as keyof typeof preprocessorState;
-    preprocessorState[preprocessor].enabled = el.checked;
-    this.props.onPreprocessorOptionsChange(preprocessorState);
+    const preprocessor = el.name.split('.')[0] as keyof PreprocessorState;
+
+    this.props.onPreprocessorOptionsChange({
```
developit: Forgot to mention - I wrote a utility for doing the immutable deep set stuff, but then found a good one already available. Think we should pull it in? It's 180b and we have this pattern in a couple spots:

https://github.com/fwilkerson/clean-set

Seems like it might help with the typings as well.

src/components/Options/index.tsx
```diff
@@ -107,25 +112,27 @@ export default class Options extends Component<Props, State> {
...
-            onChange={this.onQuantizerOptionsChange}
-          />
-        ) : (
-          <div/>
+        {encoderState.type !== 'identity' && (
+          <div>
+            <p>Quantization</p>
+            <label>
+              <input
+                name="quantizer.enable"
```
developit: `enabled`?

src/components/Options/index.tsx
jakearchibald: When wouldn't we allow it?

src/components/Options/index.tsx
jakearchibald: This looks really good. Yeah, we should use it.

In this case, we should probably just be getting all the data from the form fresh each time (this is how I do it with the webp encoder).

src/components/Options/index.tsx
jakearchibald: I only changed the indenting as part of this PR, but happy to rename it.

surma:
Thanks for fixing stuff!

jakearchibald:
I'll pick up @developit's comments in another PR

PR #123, opened 2018-08-06
Easter egg

jakearchibald:
This is silly, but I like it.

We could also allow saving to the ZX video memory format (SCR), but it'd have to be the correct resolution. That could come later I guess.

The performance of this is pretty bad given the number of times quantize is called, but hey it's just a silly thing, so maybe it doesn't matter.


src/lib/util.ts
```diff
@@ -159,3 +159,24 @@ export function inputFieldValueAsNumber(field: any): number {
 export function inputFieldCheckedAsNumber(field: any): number {
   return Number((field as HTMLInputElement).checked);
 }
+
+/**
+ * Creates a promise that resolves when the user types the konami code.
+ */
+export function konami(): Promise<void> {
+  return new Promise((resolve) => {
+    const expectedPattern = '38384040373937396665';
```
mathiasbynens: `// ↑	↑	↓ ↓ ← → ← → B A`

jakearchibald: I was a little conflicted in terms of comments. As in, would it make it too obvious? But I guess it isn't supposed to be that secret. I'll add the comment.

jakearchibald:
If the ZX had a 4k mode:

mvimg_20180504_185133

codecs/imagequant/imagequant.c
```diff
@@ -50,6 +52,169 @@ void quantize(uint8_t* image_buffer, int image_width, int image_height, int num_
...
+/**
+ * The ZX has one bit per pixel, but can assign two colours to an 8x8 block. The two colours must
+ * both be 'regular' or 'bright'. Black exists as both regular and bright.
+ */
+EMSCRIPTEN_KEEPALIVE
+void zx_quantize(uint8_t* image_buffer, int image_width, int image_height, float dithering) {
+  int size = image_width * image_height;
+  int bytes_per_pixel = 4;
+  result = (int) malloc(size * bytes_per_pixel);
+  uint8_t* block = malloc(8 * 8 * bytes_per_pixel);
```
surma: You could move this array to the stack.

```c
uint8_t block[8 * 8 * 4];
```

And no `free()` call at the end.

codecs/imagequant/imagequant.c
```diff
@@ -50,6 +52,169 @@ void quantize(uint8_t* image_buffer, int image_width, int image_height, int num_
...
+
+      // Get the three most popular colours for the block.
+      int first_color_index = 0;
+      int second_color_index = 0;
+      int third_color_index = 0;
+      int highest_popularity = -1;
+      int second_highest_popularity = -1;
+      int third_highest_popularity = -1;
+
+      for (int color_index = 0; color_index < 15; color_index++) {
```
surma: I think there’s a more elegant way to do this block, but also it’s an easter egg. I might grab this in my spare time if I fee like it 😅

jakearchibald: Interested to see the better way of doing this. My unfamiliarity with C was getting in the way.

src/lib/util.ts
```diff
@@ -159,3 +159,25 @@ export function inputFieldValueAsNumber(field: any): number {
...
+ */
+export function konami(): Promise<void> {
+  return new Promise((resolve) => {
+    // Keycodes for: ↑ ↑ ↓ ↓ ← → ← → B A
+    const expectedPattern = '38384040373937396665';
+    let rollingPattern = '';
+
+    const listener = (event: KeyboardEvent) => {
+      rollingPattern += event.keyCode;
+      rollingPattern = rollingPattern.slice(0, expectedPattern.length);
```
remy: This is wrong. It'll only look at the first 12 key presses in the "rolling pattern", thus ignoring everything after the 12th keypress.

I believe it should be `rollingPattern = rollingPattern.slice(-expectedPattern.length);` - so you get the trailing 12 keypresses.

surma: 🔥 🔥 🔥

jakearchibald: lol I can't code

jakearchibald: (that explains a couple of issues I saw that I thought were focus issues)

jakearchibald: https://github.com/GoogleChromeLabs/squoosh/commit/602d5140f9e5d26f34b696852d01b6d0088e1ad7 fixed! Thanks for spotting that.

PR #124, opened 2018-08-06
Adding clean-set

jakearchibald:
**DO NOT MERGE**

I had a go at adding `cleanSet`, but looking at the diff, it doesn't seem like the thing we want.

There may be cases where we do want it, but it seems like we more frequently want something like `cleanMerge`:

```js
const newObj = cleanMerge(obj, path, objToMerge);
// Eg:
const obj = {
  foo: { enabled: true, val: 123, bar: 'hello' },
};
const newObj = cleanMerge(obj, 'foo', { enabled: false, val: 321 });
console.log(newObj.foo.enabled); // false
console.log(newObj.foo.val); // 321
console.log(newObj.foo.bar); // 'hello'
```

So in the diff:

```js
images = cleanSet(images, `${index}.file`, file as any);
images = cleanSet(images, `${index}.bmp`, bmp as any);
images = cleanSet(images, `${index}.downloadUrl`, URL.createObjectURL(file) as any);
images = cleanSet(
  images,
  `${index}.loading`,
  (images[index].loadingCounter !== loadingCounter) as any,
);
images = cleanSet(images, `${index}.loadedCounter`, loadingCounter as any);
```

Would be:

```js
images = cleanMerge(images, index, {
  file,
  bmp,
  downloadUrl: URL.createObjectURL(file),
  loading: images[index].loadingCounter !== loadingCounter,
});
```

The `cleanMerge` implementation would be something like:

```ts
export function cleanMerge<A>(source: A, keys: string | string[], toMerge: any[] | object): A {
  if (typeof keys === 'string') keys = keys.split('.');

  let last = copy(source);
  let mergedObject = last;

  const lastIndex = keys.length - 1;
  for (const [i, key] of keys.entries()) {
    if (lastIndex) {
      last[key] = Object.assign(copy(last[key]), toMerge);
    }
    else {
      last = last[key] = copy(last[key]);
    }
  }

  return mergedObject;
}

function copy<A = any[] | object>(source: A): A {
  if (Array.isArray(source)) return [...source] as any;
  return {...(source as any)};
}
```

(although I haven't tested it)

### DIFF
--- a/src/components/App/index.tsx
+++ b/src/components/App/index.tsx
@@ -1,5 +1,6 @@
 import { h, Component } from 'preact';
 import { partial } from 'filesize';
+import cleanSet from 'clean-set';
 
 import { bind, bitmapToImageData } from '../../lib/util';
 import * as style from './style.scss';
@@ -151,9 +152,6 @@ export default class App extends Component<Props, State> {
     type: EncoderType,
     options?: EncoderOptions,
   ): void {
-    const images = this.state.images.slice() as [EncodedImage, EncodedImage];
-    const oldImage = images[index];
-
     // Some type cheating here.
     // encoderMap[type].defaultOptions is always safe.
     // options should always be correct for the type, but TypeScript isn't smart enough.
@@ -162,11 +160,9 @@ export default class App extends Component<Props, State> {
       options: options || encoderMap[type].defaultOptions,
     } as EncoderState;
 
-    images[index] = {
-      ...oldImage,
-      encoderState,
-      preprocessorState,
-    };
+    let { images } = this.state;
+    images = cleanSet(images, `${index}.encoderState`, encoderState as any);
+    images = cleanSet(images, `${index}.preprocessorState`, preprocessorState as any);
 
     this.setState({ images });
   }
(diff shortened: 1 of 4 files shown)


developit:
@jakearchibald I like that syntax better, it's more reminiscent of the `immutability-helper` / mondodb $update syntax.  Are you going to "update this PR in-place" (haha) with cleanMerge?

jakearchibald:
I'll update this PR. Good point about making it a separate module.

jakearchibald:
The problem with any solution in this shape is it bypasses type checking.

```js
cleanMerge(foo, 'bar.baz', { yo: true });
```

There's no way to statically determine that `foo.bar.baz.yo` must allow a boolean type.

jakearchibald:
I've updated this PR with `cleanSet` and `cleanMerge`.

My main problem is the type checking you lose vs what we're doing currently. What do you think, big deal or no problem?

developit:
Mentioned in chat, but I think if we can pick a half-decent way to enforce types at the call site, the keypath stuff seems like an acceptable risk.

src/components/App/index.tsx
```diff
@@ -162,12 +160,7 @@ export default class App extends Component<Props, State> {
       options: options || encoderMap[type].defaultOptions,
     } as EncoderState;
 
-    images[index] = {
-      ...oldImage,
-      encoderState,
-      preprocessorState,
-    };
-
+    const images = cleanMerge(this.state.images, '' + index, { encoderState, preprocessorState });
```
developit:
```js
const images = cleanMerge(
    this.state.images,
    '' + index,
    { encoderState, preprocessorState } as Partial<EncodedImage>
);
```

developit:
Still want to do the `Partial<EncodedImage>` cast?

jakearchibald:
Does the cast do anything? It doesn't result in any checks, unless I'm missing something

PR #125, opened 2018-08-06
Avoid preprocessing images that have already been preprocessed.

jakearchibald:
The ZX quantiser is slow, so we *need this*.

Happy to delay merging this until https://github.com/GoogleChromeLabs/squoosh/pull/124 is resolved, since it touches a lot of the same stuff.

### DIFF
--- a/src/components/App/index.tsx
+++ b/src/components/App/index.tsx
@@ -66,6 +66,10 @@ interface State {
   error?: string;
 }
 
+interface UpdateImageOptions {
+  skipPreprocessing?: boolean;
+}
+
 const filesize = partial({});
 
 async function preprocessImage(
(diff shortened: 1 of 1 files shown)


src/components/App/index.tsx
```diff
@@ -150,57 +148,43 @@ export default class App extends Component<Props, State> {
...
-    // encoderMap[type].defaultOptions is always safe.
-    // options should always be correct for the type, but TypeScript isn't smart enough.
-    const encoderState: EncoderState = {
-      type,
-      options: options || encoderMap[type].defaultOptions,
-    } as EncoderState;
-
-    const images = cleanMerge(this.state.images, index, { encoderState, preprocessorState });
-    this.setState({ images });
-  }
```
jakearchibald: It seems better to update the relevant parts of state in the individual change events.

src/components/App/index.tsx
```diff
@@ -273,7 +265,7 @@ export default class App extends Component<Props, State> {
 
     let bmp;
     try {
-      bmp = await createImageBitmap(file);
+      bmp = await decodeImage(file);
```
jakearchibald: @surma I think this should happen, right?

jakearchibald:
@developit @kosamari @surma This is ready for review now

src/components/App/index.tsx
```diff
@@ -240,7 +224,8 @@ export default class App extends Component<Props, State> {
     }
   }
 
-  async updateImage(index: number): Promise<void> {
+  async updateImage(index: number, options: UpdateImageOptions = {}): Promise<void> {
+    const { skipPreprocessing = false } = options;
```
surma: Syntactic nit: Not a fan of this visually. Nbd, especially since it’s only one option, but I prefer

```js
options = {...defaultUpdateImageOptions, ...options};
```

WDYT?

src/components/App/index.tsx
surma: +1

src/components/App/index.tsx
jakearchibald: Doesn't that just introduce more scrolling to find out what the defaults are?

jakearchibald: (fwiw our lint rules don't allow argument overwriting)

surma: I feel like we had this conversation before. It probably does, but I find this incredibly noise, especially with more options. As I said, nbd.

jakearchibald: Last time we had this discussion we settled on the pattern above 😄. But yeah, it was a compromise. I initially put the destructuring in the args, which is kinda ugly, and until it pops up as part of autocomplete, I'm not bothered.

PR #134, opened 2018-08-09
Adding changed points and coalescing to pointer tracker. Fixes #133.

### LINKED ISSUE #133: PointerTracker: Expose changed pointers and getCoalescedEvents
I ticked the tyres of PointerTracker by making a little painting 'app' https://codepen.io/jaffathecake/pen/pZqzLw?editors=0010.

I'm overdrawing because I'm doing stuff with pointers that haven't moved, and I have to jump through hoops to get to `getCoalescedEvents`.

I think this could be fixed by adding `changedPointers`

```js
const tracker = new PointerTracker(el, {
  move(previous, changed, event) {
    // …
  }
});
```

Where `changed` is an array of pointers that moved. To get the previous pointer, the developer would need to do `previousPointer = previous.find(p => p.id = changed[0].id)`, but that seems ok.

Then, pointer instances can have `getCoalesced` or similar, which returns an array of pointers. This would work no matter what kind of event triggered it (returning an array of one pointer). I find "coalesced" tough to spell, so it might be worth thinking of a different name.

jakearchibald:
Low priority, as squoosh doesn't need this.

### DIFF
--- a/src/lib/PointerTracker/index.ts
+++ b/src/lib/PointerTracker/index.ts
@@ -12,17 +12,20 @@ export class Pointer {
   clientY: number;
   /** ID for this pointer */
   id: number = -1;
+  /** The platform object used to create this Pointer */
+  nativePointer: Touch | PointerEvent | MouseEvent;
 
-  constructor (private _nativePointer: Touch | PointerEvent | MouseEvent) {
-    this.pageX = _nativePointer.pageX;
-    this.pageY = _nativePointer.pageY;
-    this.clientX = _nativePointer.clientX;
-    this.clientY = _nativePointer.clientY;
+  constructor (nativePointer: Touch | PointerEvent | MouseEvent) {
+    this.nativePointer = nativePointer;
+    this.pageX = nativePointer.pageX;
+    this.pageY = nativePointer.pageY;
+    this.clientX = nativePointer.clientX;
+    this.clientY = nativePointer.clientY;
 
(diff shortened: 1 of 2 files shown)


src/lib/PointerTracker/missing-types.d.ts
```diff
@@ -4,3 +4,7 @@ interface Window {
   PointerEvent: typeof PointerEvent;
   Touch: typeof Touch;
 }
+
+interface PointerEvent {
+  getCoalescedEvents(): PointerEvent[];
```
surma: Eugh, sad they didn’t include that in TypeScript v3 :(

src/lib/PointerTracker/index.ts
```diff
@@ -172,19 +187,18 @@ export class PointerTracker {
...
       const index = this.currentPointers.findIndex(p => p.id === pointer.id);
-      if (index === -1) continue;
-      shouldCallback = true;
+      if (index === -1) continue; // Not a pointer we're tracking
+      trackedChangedPointers.push(pointer);
       this.currentPointers[index] = pointer;
     }
 
-    if (!shouldCallback) return;
+    if (!trackedChangedPointers[0]) return;
```
surma: `trackedChangedPointers.length <= 0`?

jakearchibald: I wouldn't use `<` unless it could be less than 1, but happy to go for `=== 0`.

PR #136, opened 2018-08-10
Add a `<range-input>` WC.

developit:
Note: the 2 flipped ranges used in WebP options are problematic. Perhaps I should add `<range-input reversed>`?

### DIFF
--- a/src/codecs/webp/options.tsx
+++ b/src/codecs/webp/options.tsx
@@ -74,9 +74,8 @@ export default class WebPEncoderOptions extends Component<Props, {}> {
       <div>
         <label>
           Effort:
-          <input
+          <range-input
             name="lossless_preset"
-            type="range"
             min="0"
             max="9"
             value={'' + determineLosslessQuality(options.quality)}
@@ -85,10 +84,9 @@ export default class WebPEncoderOptions extends Component<Props, {}> {
         </label>
         <label>
           Slight loss:
-          <input
+          <range-input
             class={styles.flipRange}
             name="near_lossless"
-            type="range"
             min="0"
             max="100"
             value={'' + options.near_lossless}
@@ -119,9 +117,8 @@ export default class WebPEncoderOptions extends Component<Props, {}> {
       <div>
         <label>
           Effort:
-          <input
+          <range-input
             name="method_input"
-            type="range"
             min="0"
             max="6"
             value={'' + options.method}
@@ -130,9 +127,8 @@ export default class WebPEncoderOptions extends Component<Props, {}> {
         </label>
         <label>
           Quality:
-          <input
+          <range-input
             name="quality"
-            type="range"
             min="0"
             max="100"
             step="0.01"
@@ -151,9 +147,8 @@ export default class WebPEncoderOptions extends Component<Props, {}> {
         </label>
         <label>
           Alpha quality:
-          <input
+          <range-input
             name="alpha_quality"
-            type="range"
             min="0"
             max="100"
             value={'' + options.alpha_quality}
@@ -162,9 +157,8 @@ export default class WebPEncoderOptions extends Component<Props, {}> {
         </label>
         <label>
           Alpha filter quality:
-          <input
+          <range-input
             name="alpha_filtering"
-            type="range"
             min="0"
             max="2"
             value={'' + options.alpha_filtering}
@@ -173,9 +167,8 @@ export default class WebPEncoderOptions extends Component<Props, {}> {
         </label>
         <label>
           Spacial noise shaping:
-          <input
+          <range-input
             name="sns_strength"
-            type="range"
             min="0"
             max="100"
             value={'' + options.sns_strength}
@@ -193,9 +186,8 @@ export default class WebPEncoderOptions extends Component<Props, {}> {
         </label>
         <label>
           Filter strength:
-          <input
+          <range-input
             name="filter_strength"
-            type="range"
             min="0"
             max="100"
             disabled={!!options.autofilter}
@@ -214,10 +206,9 @@ export default class WebPEncoderOptions extends Component<Props, {}> {
         </label>
         <label>
           Filter sharpness:
-          <input
+          <range-input
             class={styles.flipRange}
             name="filter_sharpness"
-            type="range"
             min="0"
             max="7"
             value={'' + options.filter_sharpness}
@@ -235,9 +226,8 @@ export default class WebPEncoderOptions extends Component<Props, {}> {
         </label>
         <label>
           Passes:
-          <input
+          <range-input
             name="pass"
-            type="range"
             min="1"
             max="10"
             value={'' + options.pass}
@@ -258,9 +248,8 @@ export default class WebPEncoderOptions extends Component<Props, {}> {
         </label>
         <label>
           Segments:
-          <input
+          <range-input
             name="segments"
-            type="range"
             min="1"
             max="4"
             value={'' + options.segments}
@@ -269,9 +258,8 @@ export default class WebPEncoderOptions extends Component<Props, {}> {
         </label>
         <label>
           Partitions:
-          <input
+          <range-input
             name="partitions"
-            type="range"
             min="0"
             max="3"
             value={'' + options.partitions}
--- a/src/components/Output/custom-els/RangeInput/index.ts
+++ b/src/components/Output/custom-els/RangeInput/index.ts
@@ -27,7 +27,7 @@ class RangeInputElement extends HTMLElement {
   private _input = document.createElement('input');
   private _valueDisplayWrapper = document.createElement('div');
   private _valueDisplay = document.createElement('span');
-  private _precision = 0;
+  private _precision?: number;
 
   constructor() {
     super();
@@ -60,11 +60,11 @@ class RangeInputElement extends HTMLElement {
   }
 
   set valueDisplayPrecision(precision: string) {
-    this._precision = parseInt(precision, 10) || 0;
+    this._precision = precision === '' ? undefined : parseInt(precision, 10) || 0;
   }
 
   get valueDisplayPrecision() {
-    return '' + this._precision;
+    return this._precision == null ? '' : '' + this._precision;
   }
 
   @bind
@@ -84,10 +84,10 @@ class RangeInputElement extends HTMLElement {
     const percent = 100 * (value - min) / (max - min);
     this.style.setProperty('--value-percent', percent + '%');
     let displayValue = '' + value;
-    if (this._precision) {
-      displayValue = parseFloat(displayValue).toPrecision(this._precision);
-    } else {
+    if (this._precision === 0) {
       displayValue = '' + Math.round(parseFloat(displayValue));
+    } else if (this._precision !== undefined) {
+      displayValue = parseFloat(displayValue).toPrecision(this._precision);
     }
     this._valueDisplay.textContent = '' + displayValue;
   }
--- a/src/components/Output/custom-els/RangeInput/index.ts
+++ b/src/components/Output/custom-els/RangeInput/index.ts
@@ -1,13 +1,13 @@
 import './styles.css';
 
-const EVENTS = [
+const RETARGETED_EVENTS = [
   'input',
   'change',
   'focus',
   'blur',
 ];
 
-const PROPERTIES = [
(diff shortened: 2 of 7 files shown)


surma:
How about we give `range-input` a `left` and `right` instead of `min` and `max`?

jakearchibald:
`start` and `end` (in case we ever add a vertical option).

Yeah, the flipping is something the component should do with it's internal range input, rather than flipping the `<range-input>` itself.

developit:
@jakearchibald @surma the implication being that `min<max` but `start>end` is fine?

jakearchibald:
Yep!

src/components/Output/custom-els/RangeInput/index.ts
```diff
@@ -0,0 +1,94 @@
...
+      this.appendChild(this._input);
+      this._valueDisplayWrapper.appendChild(this._valueDisplay);
+      this.appendChild(this._valueDisplayWrapper);
+    }
+  }
+
+  private _handleEvent(event: Event) {
+    this._update();
+    event.stopImmediatePropagation();
+    const retargetted = new (event.constructor as typeof Event)(event.type, event);
```
surma: Does `... as Constructor<Event>` work?

surma: But also, is this redirection necessary?

developit: it's the same approach Jake used elsewhere, was the only thing I could get to work.  I could just use the Event constructor maybe? But there are multiple event types.

src/components/Output/custom-els/RangeInput/index.ts
```diff
@@ -0,0 +1,94 @@
...
+class RangeInputElement extends HTMLElement {
+  private _input = document.createElement('input');
+  private _valueDisplayWrapper = document.createElement('aside');
+  private _valueDisplay = document.createElement('span');
+
+  constructor() {
+    super();
+    this._input.type = 'range';
+    this._update = this._update.bind(this);
+    this._handleEvent = this._handleEvent.bind(this);
```
jakearchibald: Use `@bind` for these two.

src/components/Output/custom-els/RangeInput/index.ts
```diff
@@ -0,0 +1,94 @@
+import './styles.css';
+
+const EVENTS = [
+  'input',
+  'change',
+  'focus',
+  'blur',
+];
```
jakearchibald: 🚲 `RETARGETED_EVENTS`?

src/components/Output/custom-els/RangeInput/index.ts
```diff
@@ -0,0 +1,94 @@
...
+
+const PROPERTIES = [
+  'name',
+  'value',
+  'min',
+  'max',
+  'step',
+];
```
jakearchibald: 🚲 `REFLECTED_PROPERTIES`?

src/components/Output/custom-els/RangeInput/index.ts
```diff
@@ -0,0 +1,94 @@
...
+
+interface RangeInputElement {
+  value: string;
+  min?: string;
+  max?: string;
+  step?: string;
```
jakearchibald: None of `min`, `max`, or `step` are optional.

src/components/Output/custom-els/RangeInput/index.ts
```diff
@@ -0,0 +1,94 @@
...
+    }
+    if (precision) {
+      displayValue = parseFloat(displayValue).toPrecision(precision);
+    }
+    this._valueDisplay.textContent = '' + displayValue;
+  }
+
+  attributeChangedCallback (name: string, value: string) {
+    if (name === 'value') this.value = value;
+  }
```
jakearchibald: Does this work? Doesn't look like you've defined "value" as an attribute you're interested in.

src/components/Output/custom-els/RangeInput/index.ts
```diff
@@ -0,0 +1,94 @@
...
+  value: string;
+  min?: string;
+  max?: string;
+  step?: string;
+  precision?: number | string;
+}
+
+class RangeInputElement extends HTMLElement {
+  private _input = document.createElement('input');
+  private _valueDisplayWrapper = document.createElement('aside');
```
jakearchibald: This doesn't feel like something a screen reader should be interested in, so just go with div/span and aria-hidden. The value is just noise compared to what the input already provides.

src/components/Output/custom-els/RangeInput/index.ts
```diff
@@ -0,0 +1,94 @@
...
+  'max',
+  'step',
+];
+
+interface RangeInputElement {
+  value: string;
+  min?: string;
+  max?: string;
+  step?: string;
+  precision?: number | string;
```
jakearchibald: `number | string` - pick one to be compatible with the rest of the interface.

src/components/Output/custom-els/RangeInput/index.ts
jakearchibald: `new Event` is fine here since all the events being retargeted are just regular events.

src/components/Output/custom-els/RangeInput/index.ts
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  private _update() {
+    const value = parseFloat(this.value || '0');
+    const min = parseFloat(this._input.min || '0');
+    const max = parseFloat(this._input.max || '100');
+    const percent = 100 * (value - min) / (max - min);
+    this.style.setProperty('--value-percent', percent + '%');
+    let displayValue = '' + value;
+    let precision = this.precision;
```
jakearchibald: `precision` should be renamed to reflect that it's only related to display.

src/components/Output/custom-els/RangeInput/styles.css
```diff
@@ -0,0 +1,102 @@
...
+range-input input::-webkit-slider-runnable-track,
+range-input input::-moz-range-track,
+range-input input::-ms-track {
+  appearance: none;
+  -ms-appearance: none;
+  -moz-appearance: none;
+  -webkit-appearance: none;
+}
+
+range-input:before {
```
jakearchibald: Nit: `::` for pseudo-elements.

src/components/Output/custom-els/RangeInput/styles.css
```diff
@@ -0,0 +1,102 @@
...
+  left: var(--value-percent, 0%);
+  bottom: 3px;
+  width: 32px;
+  height: 62px;
+  text-align: center;
+  padding: 8px 3px 0;
+  margin: 0 0 0 -16px;
+  filter: drop-shadow(0 1px 3px rgba(0,0,0,0.3));
+  transform-origin: 50% 90%;
+  opacity: 0;
```
jakearchibald: I'm getting a bit of jank when the bubble first appears. I wonder if an opacity of 0.0001 and `will-change: transform` will improve responsiveness here.

developit:
@jakearchibald  regarding the text field, I'd prefer to keep that separate from this component since it's hierarchically at the same level as a wrapping `<label>`. Should be easy to drive it using props+events, and that way we can have what I think seemed to be the implied "click to edit" text mode on that field, which wouldn't make much sense in a generic slider component.

src/components/Output/custom-els/RangeInput/index.ts
developit: tis now a div

src/components/Output/custom-els/RangeInput/index.ts
developit: it's now `valueDisplayPrecision`.

src/components/Output/custom-els/RangeInput/styles.css
developit: good call - I was seeing that but wasn't sure if it was related.

developit: that seems to have addressed it for me.

src/components/Output/custom-els/RangeInput/index.ts
developit: They have default values, but I agree it's likely pointless to support that.

src/components/Output/custom-els/RangeInput/styles.css
developit: ... actually it seems like removing the focus ring addressed the issue. odd.

src/components/Output/custom-els/RangeInput/index.ts
developit: not sure what you mean, though this is now just a reflection of the input's value attribute now.

src/components/Output/custom-els/RangeInput/index.ts
developit: didn't think we were using that in the web components. added.

src/components/Output/custom-els/RangeInput/index.ts
jakearchibald: But it only fires for things listed in `observedAttributes` right?

src/components/Output/custom-els/RangeInput/styles.css
jakearchibald: Hah, I guess that's a heavy paint operation?

src/codecs/webp/options.tsx
```diff
@@ -193,9 +186,8 @@ export default class WebPEncoderOptions extends Component<Props, {}> {
         </label>
         <label>
           Filter strength:
-          <input
+          <range-input
             name="filter_strength"
-            type="range"
             min="0"
             max="100"
             disabled={!!options.autofilter}
```
jakearchibald: This is broken - filed https://github.com/GoogleChromeLabs/squoosh/issues/142

src/components/Output/custom-els/RangeInput/index.ts
```diff
@@ -0,0 +1,107 @@
...
+
+  @bind
+  private _update() {
+    this._reflectAttributes();
+    const value = parseFloat(this.value || '0');
+    const min = parseFloat(this._input.min || '0');
+    const max = parseFloat(this._input.max || '100');
+    const percent = 100 * (value - min) / (max - min);
+    this.style.setProperty('--value-percent', percent + '%');
+    let displayValue = '' + value;
```
jakearchibald: This seems to go from string to number to string back to number.

jakearchibald:
Also, feels like `valueDisplayPrecision` should have a corresponding attribute. Maybe `label-precision` and `labelPrecision` (or is that confusing with `<label>`)?

developit:
Update: still fighting with TypeScript passively on this one. Hopefully will have a proper update tomorrow.

jakearchibald:
Which parts are causing issues?

src/codecs/webp/options.tsx
developit: fixed.
screen shot 2018-09-17 at 9 43 50 pm

developit:
ok. I *think* I have something working like you described. TBH I hate it, but this is where I was able to get to without resorting to `as any` in a myriad of places.

src/components/Output/index.tsx
```diff
@@ -2,6 +2,7 @@ import { h, Component } from 'preact';
 import PinchZoom from './custom-els/PinchZoom';
 import './custom-els/PinchZoom';
 import './custom-els/TwoUp';
+import './custom-els/RangeInput';
```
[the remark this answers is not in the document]
developit: Would you prefer it be imported from the options that require range inputs? I didn't want to since we arbitrarily nested these CE's within directories specific to each component, seemed like reaching across a boundary.

src/components/Output/custom-els/RangeInput/index.ts
```diff
@@ -0,0 +1,148 @@
+import { bind } from '../../../../lib/util';
+import './styles.css';
+
+const RETARGETED_EVENTS = [
+  'input',
+  'change',
```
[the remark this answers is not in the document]
developit: Yes. In this case the retargeting is pointless, but it invokes `this._update()` on input and change.

jakearchibald:
I agree it's reaching across a boundary, but I think the answer is to move
the custom element to a higher level.

jakearchibald:
Merging #171

PR #139, opened 2018-08-16
Pinch zoom features

### LINKED ISSUE #129: Expose a way to set scale/transform while preserving perspective/pan
Right now invoking `PinchZoom.setTransform({ scale })` uses a transform origin of `0,0`.  It might be nice to expose either a way to scale around a given center, or assume a default of the center of the `PinchZoom` region.

jakearchibald:
Do not merge, as this is not based on master, but please review!

Fixes #129.

### DIFF
--- a/src/components/Output/custom-els/PinchZoom/index.ts
+++ b/src/components/Output/custom-els/PinchZoom/index.ts
@@ -6,7 +6,14 @@ interface Point {
   clientY: number;
 }
 
-interface ApplyChangeOpts {
+interface ChangeOptions {
+  /**
+   * Fire a 'change' event if values are different to current values
+   */
+  allowChangeEvent?: boolean;
+}
+
+interface ApplyChangeOpts extends ChangeOptions {
   panX?: number;
   panY?: number;
   scaleDiff?: number;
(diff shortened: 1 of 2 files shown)


src/components/Output/custom-els/PinchZoom/index.ts
```diff
@@ -103,6 +117,59 @@ export default class PinchZoom extends HTMLElement {
...
+  scaleTo(scale: number, opts: ScaleToOpts = {}) {
+    let {
+      originX = 0,
+      originY = 0,
+    } = opts;
+
+    const {
+      relativeTo = 'content',
+      allowChangeEvent = false,
+    } = opts;
```
jakearchibald: Are these good defaults?

jakearchibald:
I'd like to add the ability to animate between values (maybe an enhancement for browsers that support `el.animate`).

src/components/Output/index.tsx
```diff
@@ -78,16 +77,16 @@ export default class Output extends Component<Props, State> {
 
   @bind
   zoomIn() {
-    this.setState({
-      scale: Math.min(this.state.scale * 1.25, 100),
-    });
+    if (!this.pinchZoomLeft) throw Error('Missing pinch-zoom element');
+
+    this.pinchZoomLeft.scaleTo(Math.min(this.state.scale * 1.25, 100), scaleToOpts);
```
developit: Worth removing the min/max here to let pinch-zoom decide?

developit:
LGTM. I do like the idea of animating using `.animate()` in browsers where it's available. Perhaps we could add an `animate?:bool` to the options?

src/components/Output/index.tsx
jakearchibald: Agreed. Will do that.

jakearchibald:
@developit I've hard coded a min scale as it fixes the bug we have now. Making it configurable isn't completely straight forward, so I've made an issue for it https://github.com/GoogleChromeLabs/squoosh/issues/149.

jakearchibald:
Shut up @googlebot

PR #140, opened 2018-08-17
Mozjpeg opts

### LINKED ISSUE #83: Add more MozJPEG options
We need to expose additional MozJPEG options:

1. [ ] Via WASm.
1. [ ] Via UI.

jakearchibald:
Fixes #83.

I added everything I felt was kinda useful.

I'd like to enable arithmetic coding at some point too, but we'd need to use the mozjpeg decoder for that. No browser supports arithmetic coded JPEGs. For now, the WASM is capable of it, but the UI doesn't expose it. Happy to remove it from the wasm if that's bad. Otherwise I'll create an issue to solve the decoding part.


codecs/mozjpeg_enc/mozjpeg_enc.cpp
```diff
@@ -109,18 +118,48 @@ void encode(uint8_t* image_buffer, int image_width, int image_height, int qualit
...
+  cinfo.optimize_coding = opts.optimize_coding ? TRUE : FALSE;
+
+  if (opts.arithmetic) {
+    cinfo.arith_code = TRUE;
+    cinfo.optimize_coding = FALSE;
+  }
+
+  cinfo.smoothing_factor = opts.smoothing;
+
+  jpeg_c_set_bool_param(&cinfo, JBOOLEAN_USE_SCANS_IN_TRELLIS, opts.trellis_multipass ? TRUE : FALSE);
```
surma: Are the ternaries necessary here?

jakearchibald: Good question. I assumed not since C doesn't have a real boolean type right? I'll test anyway.

surma: I mean, you are using `bool` in the struct definition, so I’d assume the compiler does the right thing ™️  here

jakearchibald: Yes it does.

Issue #146, opened 2018-08-21
Copy to other side


jakearchibald:
Take the settings & result from one side, and copy it to the other.

Ideally this should be instant, as we've already got the result.

jakearchibald:
If we have some kind of encodeds cache, we might get "instant" for free.

jakearchibald:
Done!

PR #148, opened 2018-08-22
Adding encoding results cache

### LINKED ISSUE #87: Cache n previous outputs
I don't know if this should be V1 or not…

When toggling an options switch, it'd be nice if it didn't have to re-encode. To achieve this, we could cache the previous n encodes, and if the settings match something we already have, just pull the item out of the cache.

The size of the cache may be dependent on image size.

jakearchibald:
Fixes #87.

I was chatting to @developit about #146 - copy to other side, and he pointed out that implementing would be a lot easier if we already had a results cache, so I went ahead and made one.

Do not merge, but please review - this is based on the UI branch as I wanted to use the shallow compare util that @developit wrote.

### DIFF
--- a/src/components/App/index.tsx
+++ b/src/components/App/index.tsx
@@ -6,6 +6,7 @@ import Output from '../Output';
 import Options from '../Options';
 import { FileDropEvent } from './custom-els/FileDrop';
 import './custom-els/FileDrop';
+import ResultCache from './result-cache';
 
 import * as quantizer from '../../codecs/imagequant/quantizer';
 import * as optiPNG from '../../codecs/optipng/encoder';
@@ -39,7 +40,7 @@ import { cleanMerge, cleanSet } from '../../lib/clean-modify';
 
 type Orientation = 'horizontal' | 'vertical';
 
-interface SourceImage {
+export interface SourceImage {
   file: File;
   bmp: ImageBitmap;
   data: ImageData;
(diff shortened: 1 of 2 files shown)


src/components/App/index.tsx
```diff
@@ -252,19 +254,39 @@ export default class App extends Component<Props, State> {
...
+    const cacheResult = this.encodeCache.match(source, image.preprocessorState, image.encoderState);
+
+    if (cacheResult) {
+      ({ file, preprocessed, bmp } = cacheResult);
+    } else {
+      try {
+        // Special case for identity
+        if (image.encoderState.type === identity.type) {
+          file = source.file;
+          bmp = source.bmp;
```
jakearchibald: I snuck a little fix in here. Previously we were passing `file` to the decoder in the identity case, meaning we were decoded something we already had decoded.

src/components/App/result-cache.ts
```diff
@@ -0,0 +1,68 @@
...
+  file: File;
+}
+
+interface CacheEntry extends CacheResult {
+  preprocessorState: PreprocessorState;
+  encoderState: EncoderState;
+  source: SourceImage;
+}
+
+const SIZE = 5;
```
jakearchibald: Given how big images can be, I went for a small number here. Feels like we needed at least 4 to allow toggling an option on each side.

src/components/App/index.tsx
```diff
@@ -136,7 +137,8 @@ export default class App extends Component<Props, State> {
     orientation: this.widthQuery.matches ? 'horizontal' : 'vertical',
   };
 
-  private snackbar?: SnackBarElement;
+  snackbar?: SnackBarElement;
```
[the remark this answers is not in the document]
developit: Yes

PR #150, opened 2018-08-22
Button to copy settings to the other side

### LINKED ISSUE #146: Copy to other side

jakearchibald:
Fixes #146, but won't merge yet, since it's based off a non-master branch.

Since I had to touch the CSS my OCD kinda kicked in and I flattened it (we don't really need the cascade with CSS modules). The diff might be easier to look at with ignoring whitespace.


jakearchibald:
@developit You were right, having a results cache made this easy.

developit:
LGTM, sorry for the slow review there.

Issue #131, opened 2018-08-08
Resize preprocessor


jakearchibald:
Not for V1, but resizing is an important part of image optimisation, and there are multiple techniques available.

We should do bilinear, bicubic, lanczos, and one or two of the pixel-art scalers https://en.wikipedia.org/wiki/Pixel-art_scaling_algorithms.

For the comparison to work, we'd set the width/height of the image to be equal to the original, so the left/right appear the same size.

jakearchibald:
https://github.com/ImageMagick/ImageMagick/blob/master/MagickCore/resize.c - this looks like a good place to start. `ResizeImage()`

jakearchibald:
Some pretty amazing docs too https://www.imagemagick.org/Usage/filter/.

jakearchibald:
I've added browser resizing, but we should add third-party resizing at some point.

kosamari:
Closing this issue for now since V1 shipped with resize.
Adding line item for V2 for third-party resizing

surma:
We wanna tackle this again, so reopening.

Also found this: 

https://github.com/PistonDevelopers/resize (Rust, Lanczos3 & Mitchell)

jakearchibald:
Ohh that's very handy. Lanczos3 & Mitchell are the ones I wanted too.

surma:
I got that Rust library working in wasm. Will PR once I have done the rest of the plumbing. 



(Yes, I got bored.)

surma:
Jake and I are working on this on the `rust-resize` branch.

jakearchibald:
Closed in https://github.com/GoogleChromeLabs/squoosh/pull/498

PR #152, opened 2018-08-23
Adding resize preprocessor

jakearchibald:
Part of #131, although at some point I'd like us to support non-browser resize methods, but this solves a lot of the use cases.

I decided that resizing is a pretty important part of image compression, especially if you're grabbing something taken by a phone.

Anyway, here it is.

It isn't based on master, so don't merge yet, but I'm interested in reviews, including reviews of the interaction.

### DIFF
--- a/src/codecs/resize/options.tsx
+++ b/src/codecs/resize/options.tsx
@@ -1,0 +1,128 @@
+import { h, Component } from 'preact';
+import linkState from 'linkstate';
+import { bind, inputFieldValueAsNumber } from '../../lib/util';
+import { ResizeOptions } from './resize';
+
+interface Props {
+  options: ResizeOptions;
+  aspect: number;
+  onChange(newOptions: ResizeOptions): void;
+}
+
+interface State {
+  maintainAspect: boolean;
+}
+
+export default class QuantizerOptions extends Component<Props, State> {
+  state: State = {
+    maintainAspect: true,
+  };
+
+  form: HTMLFormElement = undefined as any as HTMLFormElement;
+
+  reportOptions() {
+    const width = this.form.width as HTMLInputElement;
+    const height = this.form.height as HTMLInputElement;
+
+    if (!width.checkValidity() || !height.checkValidity()) return;
+
+    const options: ResizeOptions = {
+      width: inputFieldValueAsNumber(width),
+      height: inputFieldValueAsNumber(height),
+      method: this.form.resizeMethod.value,
+      fitMethod: this.form.fitMethod.value,
+    };
+    this.props.onChange(options);
+  }
+
+  @bind
+  onChange(event: Event) {
+    this.reportOptions();
+  }
+
+  componentDidUpdate(prevProps: Props, prevState: State) {
+    if (!prevState.maintainAspect && this.state.maintainAspect) {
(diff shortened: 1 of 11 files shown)


src/codecs/resize/options.tsx
```diff
@@ -0,0 +1,128 @@
...
+interface State {
+  maintainAspect: boolean;
+}
+
+export default class QuantizerOptions extends Component<Props, State> {
+  state: State = {
+    maintainAspect: true,
+  };
+
+  form: HTMLFormElement = undefined as any as HTMLFormElement;
```
jakearchibald: Interested in thought on this. Technically `form` isn't defined until first mount, but it's a pain adding throws to every method.

Is this a reasonable cheat, or would it be better to create a decorator that throws if form is missing?

jakearchibald: If we decide this is ok, I should add a comment explaining it.

src/codecs/resize/options.tsx
```diff
@@ -0,0 +1,128 @@
...
+  }
+
+  @bind
+  onChange(event: Event) {
+    this.reportOptions();
+  }
+
+  componentDidUpdate(prevProps: Props, prevState: State) {
+    if (!prevState.maintainAspect && this.state.maintainAspect) {
+      this.form.height.value = Math.round(Number(this.form.width.value) / this.props.aspect);
```
surma: Just wondering: Anything we should do to make sure `aspect > 0`?

src/codecs/resize/resize.ts
```diff
@@ -0,0 +1,46 @@
...
+    if (endAspect > currentAspect) {
+      sh = opts.height / (opts.width / data.width);
+      sy = (data.height - sh) / 2;
+    } else {
+      sw = opts.width / (opts.height / data.height);
+      sx = (data.width - sw) / 2;
+    }
+  }
+
+  const bmp = await createImageBitmap(data, sx, sy, sw, sh, {
```
surma: Can you use `createImageBitmapPolyfill`? You might have to augment it a bit, but we will have to polyfill it anyway.

src/codecs/resize/resize.ts
```diff
@@ -0,0 +1,46 @@
+import { bitmapToImageData } from '../../lib/util';
+
+type CreateImageBitmapResize = 'pixelated' | 'low' | 'medium' | 'high';
+
+export async function quantize(data: ImageData, opts: ResizeOptions): Promise<ImageData> {
```
surma: Rename to `resize`?

src/codecs/resize/options.tsx
jakearchibald: I'd expect an aspect of 0 to cause a decode error much sooner I think

src/codecs/resize/options.tsx
```diff
@@ -0,0 +1,128 @@
...
+  options: ResizeOptions;
+  aspect: number;
+  onChange(newOptions: ResizeOptions): void;
+}
+
+interface State {
+  maintainAspect: boolean;
+}
+
+export default class QuantizerOptions extends Component<Props, State> {
```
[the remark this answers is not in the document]
jakearchibald: Oh ffs. Too many silly mistakes in this PR. Soz.

PR #153, opened 2018-08-23
Ohh look new features

jakearchibald:
Don't merge this. It's only here so netlify gives me a URL.


jakearchibald:
The URL is https://deploy-preview-153--squoosh.netlify.com/

PR #155, opened 2018-08-28
fix broken file renaming for file names with dot

### LINKED ISSUE #154: File renaming broken if name includes .
Seems like we trim after the first `.`. It should be the last.

sendilkumarn:
fixes #154

### DIFF
--- a/src/components/App/index.tsx
+++ b/src/components/App/index.tsx
@@ -107,7 +107,7 @@ async function compressImage(
 
   return new File(
     [compressedData],
-    sourceFilename.replace(/\..+$/, '.' + encoder.extension),
+    sourceFilename.replace(/.[^.]*$/, `.${encoder.extension}`),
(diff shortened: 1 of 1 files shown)


src/components/App/index.tsx
```diff
@@ -107,7 +107,7 @@ async function compressImage(
 
   return new File(
     [compressedData],
-    sourceFilename.replace(/\..+$/, '.' + encoder.extension),
+    sourceFilename.replace(/.([^.]*)$/, `.${encoder.extension}`),
```
mathiasbynens: Since we don't need to capture the group here, let's avoid capturing: `/.(?:[^.]*)$/`

In fact, we don't need the group at all: `/.[^.]*$/`

sendilkumarn: Good catch @mathiasbynens I am sending in a fix now

PR #158, opened 2018-09-04
Give the result cache some LRU love

developit:


src/components/App/result-cache.ts
```diff
@@ -55,7 +58,15 @@ export default class ResultCache {
...
 
-    if (matchingEntry) {
+    if (matchingIndex !== -1) {
+      const matchingEntry = this._entries[matchingIndex];
+
+      if (matchingIndex !== 0) {
+        // Move the matched result to 1st position (LRU)
+        this._entries[matchingIndex] = this._entries[0];
+        this._entries[0] = matchingEntry;
+      }
```
jakearchibald: Is this how LRU usually works? It's not something I've implemented before, but I didn't think it'd swap with `[0]`, I thought it'd shift everything along. Eg:

```js
this._entries.splice(matchingIndex, 1);
this._entries.unshift(matchingEntry);
```

developit: Ah true - I should combine this with the existing append method. Good catch.

PR #165, opened 2018-09-07
Result cache LRU

jakearchibald:
Resurrecting this after Github closed it and blamed me 😄.

I've fixed the move-to-start bit, but also gone for array methods rather than moving things around manually, as it feels more idiomatic.


jakearchibald:
Shut up @googlebot

PR #168, opened 2018-09-17
Reposition the TwoUp handle on resize

### LINKED ISSUE #105: Two-up: react to element resizing
Right now you can move the divide to the far right, make the window smaller, and you lose the divider.

On resize, we should move the divide so it's positioned at the same percentage of the parent, so if it was in the middle before resizing, it ends still in the middle.

Resize observer helps a lot here. For other browsers we need to use an iframe (eww) or maybe falling back to `window.onresize` is good enough.

devnook:
Fixes #105 
Repositions the handle for TwoUp when the element resizes.
I have 2 questions:

1. Do we have constraints on Browser compatibility? Should I include a polyfill for ResizeObserver? 
2. Making TS happy with ResizeObserver was a bit elaborate and I'm a newbie in TS - please advise if there's a better approach!

### DIFF
--- a/src/components/Output/custom-els/TwoUp/index.ts
+++ b/src/components/Output/custom-els/TwoUp/index.ts
@@ -19,6 +19,10 @@ export default class TwoUp extends HTMLElement {
    */
   private _position = 0;
   /**
+   * The position of the split in %.
+   */
+  private _relativePosition = 0.5;
+  /**
    * The value of _position when the pointer went down.
    */
   private _positionOnPointerStart = 0;
@@ -85,7 +89,8 @@ export default class TwoUp extends HTMLElement {
     // Set the initial position of the handle.
     requestAnimationFrame(() => {
       const bounds = this.getBoundingClientRect();
-      this._position = (this.orientation === 'vertical' ? bounds.height : bounds.width) / 2;
+      this._position = (this.orientation === 'vertical' ?
+          bounds.height : bounds.width) * this._relativePosition;
       this._setPosition();
     });
   }
@@ -146,6 +151,7 @@ export default class TwoUp extends HTMLElement {
 
     // Clamp position to element bounds.
     this._position = Math.max(0, Math.min(this._position, bounds[dimensionAxis]));
+    this._relativePosition = this._position / bounds[dimensionAxis];
     this._setPosition();
   }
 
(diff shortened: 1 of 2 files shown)


jakearchibald:
> Do we have constraints on Browser compatibility? Should I include a polyfill for ResizeObserver?

Nah, just fall back to `window.onresize`.

src/components/Output/custom-els/TwoUp/missing-types.d.ts
```diff
@@ -19,3 +19,36 @@ declare namespace JSX {
...
+  readonly y: number;
+  readonly width: number;
+  readonly height: number;
+  readonly top: number;
+  readonly right: number;
+  readonly bottom: number;
+  readonly left: number;
+}
+
+declare global {
```
jakearchibald: I don't think this is needed. Everything in here can be moved outside of it.

src/components/Output/custom-els/TwoUp/index.ts
```diff
@@ -37,6 +37,14 @@ export default class TwoUp extends HTMLElement {
     new MutationObserver(() => this._childrenChange())
       .observe(this, { childList: true });
 
+    // Watch for element size changes.
+    if (window.hasOwnProperty('ResizeObserver')) {
```
jakearchibald: This could just be `if ('ResizeObserver' in window)`

devnook:
PTAL

src/components/Output/custom-els/TwoUp/index.ts
```diff
@@ -77,7 +89,8 @@ export default class TwoUp extends HTMLElement {
     // Set the initial position of the handle.
     requestAnimationFrame(() => {
       const bounds = this.getBoundingClientRect();
-      this._position = (this.orientation === 'vertical' ? bounds.height : bounds.width) / 2;
+      this._position = (this.orientation === 'vertical' ?
+          bounds.height : bounds.width) * this._relativePosition;
```
jakearchibald: Really minor minor nit: The wrapping of `(` and `)` is making me twitch. Can you switch it for something like:

```ts
this._position = (this.orientation === 'vertical' ? bounds.height : bounds.width)
  * this._relativePosition;
```

Or maybe duplicate the pattern below:

```ts
const dimensionAxis = this.orientation === 'vertical' ? 'height' : 'width';
this._position = bounds[dimensionAxis] * this._relativePosition;
```

jakearchibald:
🎉

PR #170, opened 2018-09-18
Center the image canvas in PinchZoom

### LINKED ISSUE #163: Centre image on load
Right now, if you load a small image, it appears in the top left. Ideally it should appear in the middle of the viewport.

devnook:
Fixes #163


jakearchibald:
Shut up @googlebot

jakearchibald:
Phew. I hate matrix maths. There might be a more direct way of doing this, but this works.

jakearchibald:
@devnook thanks for making this a CSS solution - it also plays well with resizing. For some reason I didn't think this would do the right thing if the image was bigger than the canvas, but it works great.

PR #171, opened 2018-09-18
Taking a swing at the range input

jakearchibald:
I'm very jetlagged so maybe this is simpler for very bad reasons.

### DIFF
--- a/src/components/Output/custom-els/RangeInput/index.ts
+++ b/src/components/Output/custom-els/RangeInput/index.ts
@@ -1,0 +1,115 @@
+import { bind } from '../../../../lib/util';
+import './styles.css';
+
+const RETARGETED_EVENTS = ['focus', 'blur'];
+const UPDATE_EVENTS = ['input', 'change'];
+const REFLECTED_PROPERTIES = ['name', 'min', 'max', 'step', 'value', 'disabled'];
+const REFLECTED_ATTRIBUTES = ['name', 'min', 'max', 'step', 'value', 'disabled'];
+
+class RangeInputElement extends HTMLElement {
+  private _input = document.createElement('input');
+  private _valueDisplayWrapper = document.createElement('div');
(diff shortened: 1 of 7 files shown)


src/components/Output/custom-els/RangeInput/index.ts
```diff
@@ -89,60 +46,66 @@ class RangeInputElement extends HTMLElement {
...
+    get() {
+      return this._input[prop];
+    },
+    set(val) {
+      this._input[prop] = val;
+      this._reflectAttributes();
+      this._update();
+    },
+  });
+}
```
jakearchibald: There's some type mess here, but it feels pretty well contained, and doesn't lose types from the outside (like cleanMerge does)

developit: I'd take type mess over the verbosity I ended up having to avoid it.

jakearchibald:
I've gone ahead and implemented start/end. @developit are you happy merging this into your branch (after review of course)?

src/components/Output/custom-els/RangeInput/index.ts
```diff
@@ -88,61 +47,118 @@ class RangeInputElement extends HTMLElement {
...
 
+  private _startGtEnd(): boolean {
+    const startNum = parseFloat(this.start);
+    const endNum = parseFloat(this.end);
+    return startNum > endNum;
+  }
+
   private _reflectAttributes() {
-    for (const attributeName in REFLECTED_ATTRIBUTES) {
+    if (this.name === 'near_lossless') debugger;
```
surma: Remove? ^^

jakearchibald:
@surma @developit ok, I ditched the start/end attributes and replaced them with value flipping in Preact. Seems to work much better.

jakearchibald:
Shut up @googlebot

PR #175, opened 2018-09-26
Creating fallbacks for all ImageBitmap usage & refactoring.

jakearchibald:
This turned out to be a bigger job than I thought. The resizing I was using is behind a flag in Chrome, so I switched it out for a simpler method.

While I was at it, I created fallbacks for all `createImageBitmap` usage, and removed our usage of `ImageBitmap` and replaced it with the more universally-supported `ImageData`. We were doing a lot of conversion between the two types anyway, so switching to just one makes more sense.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,32 +5,21 @@
   "requires": true,
   "dependencies": {
     "@fimbul/bifrost": {
-      "version": "0.9.0",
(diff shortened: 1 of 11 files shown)


jakearchibald:
I was worried that `ImageData` would be more blocking or memory intensive… but you can synchronously write them to a canvas so they can't be that bad.

Makes me wonder why they're different things, though.

PR #187, opened 2018-10-10
Handle vectors

jakearchibald:
**DO NOT MERGE but please review** (not based on master)

No longer fails when loading SVG.

Also, I made the resize preprocessor vector-aware, so it'll do the resizing in SVG if the source is SVG.

### DIFF
--- a/src/codecs/resize/resize.ts
+++ b/src/codecs/resize/resize.ts
@@ -5,12 +5,12 @@ function getCoverOffsets(sw: number, sh: number, dw: number, dh: number) {
   const endAspect = dw / dh;
 
   if (endAspect > currentAspect) {
-    const newSh = dh / (dw / sw);
+    const newSh = sw / endAspect;
     const newSy = (sh - newSh) / 2;
     return { sw, sh: newSh, sx: 0, sy: newSy };
   }
 
-  const newSw = dw / (dh / sh);
+  const newSw = sh * endAspect;
   const newSx = (sw - newSw) / 2;
   return { sh, sw: newSw, sx: newSx, sy: 0 };
 }
--- a/src/codecs/resize/resize.ts
+++ b/src/codecs/resize/resize.ts
@@ -1,21 +1,28 @@
-import { nativeResize, NativeResizeMethod } from '../../lib/util';
+import { nativeResize, NativeResizeMethod, drawableToImageData } from '../../lib/util';
 
-export function resize(data: ImageData, opts: ResizeOptions): ImageData {
+function getCoverOffsets(sw: number, sh: number, dw: number, dh: number) {
+  const currentAspect = sw / sh;
+  const endAspect = dw / dh;
+
+  if (endAspect > currentAspect) {
+    const newSh = dh / (dw / sw);
+    const newSy = (sh - newSh) / 2;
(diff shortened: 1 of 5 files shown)


surma:
Hold on this, GitHub only showed me a subset of changes lol

src/codecs/resize/resize.ts
```diff
@@ -1,21 +1,28 @@
-import { nativeResize, NativeResizeMethod } from '../../lib/util';
+import { nativeResize, NativeResizeMethod, drawableToImageData } from '../../lib/util';
 
-export function resize(data: ImageData, opts: ResizeOptions): ImageData {
+function getCoverOffsets(sw: number, sh: number, dw: number, dh: number) {
+  const currentAspect = sw / sh;
+  const endAspect = dw / dh;
+
+  if (endAspect > currentAspect) {
+    const newSh = dh / (dw / sw);
```
surma: I think with some shuffling, we can make the math a bit more “obvious”:
`dh / (dw/sw) = sw * dh / dw = sw/endAspect`

WDYT?

src/codecs/resize/resize.ts
```diff
@@ -1,21 +1,28 @@
...
+    const newSy = (sh - newSh) / 2;
+    return { sw, sh: newSh, sx: 0, sy: newSy };
+  }
+
+  const newSw = dw / (dh / sh);
```
surma: Same trick here:

`dw / (dh / sh) = sh * dw / dh = sh * endAspect`

surma:
One thing I just noticed (when testing the `intro` branch):

Vector-aware resize works great, but the source image doesn’t incorporate zoom. So if I “resize” the squoosh logo SVG at 4000x4000, the output image will appear sharper than the original. Do we wanna do something about that?

screen shot 2018-10-11 at 12 10 51 pm

src/codecs/resize/resize.ts
jakearchibald: I'm not sure what's going on there, but it's definitely an invalid left-hand assignment. What am I missing?

jakearchibald: Ohhh you mean change to `sw / endAspect`! I see now

surma: Lol sorry. I am suggesting to use the expression on the very right and showing how I got there. This is not supposed to be JavaScript, just math :D My bad.

surma: ... GitHub really needs to fix their stupid code-review UI to update when someone comments.

jakearchibald:
Ahh, so the idea would be that the SVG stays sharp when scaled, but the compressed version is pixels? I can take a swing at that.

surma:
Yeah, I guess we should prevent rasterizing at every zoom interaction, but maybe we can rasterize at 200% once zoom is above 100%, rasterize at 400% once zoom is above 200% or something along those lines?

jakearchibald:
I think we get it for free if I add the `` to the output rather than a `<canvas>`. Zooming would be slower though.

surma:
Happy to leave this as a stretch goal.

jakearchibald:
Filed https://github.com/GoogleChromeLabs/squoosh/issues/190 for the sharpness thing. I agree it's worth doing.

PR #189, opened 2018-10-11
Codecs cleanup

### LINKED ISSUE #160: Switch all codecs to -Os

### LINKED ISSUE #162: Update build:codecs
In the root `package.json` we have a codec-building script, but it only handles mozjpeg right now. We should either remove this, or update it for the other codecs (and the quantizer).

Any preference @surma?

surma:
Closes #160 
Closes #162 

- Properly containerize all wasm bits (you should all be able to build the codecs now if you have Docker installed)
- Switch all codecs to `-Os`
- Remove old build targets from top-level package.json


jakearchibald:
Just tried building imageqaunt on this branch and my output looks very different. It might be asm.js? Is there something I need to do to update the docker container? If so, should that be part of the install script?

surma:
Do you have a super old version of the container image? emcc switch to wasm-by-default at some point during squoosh so I stopped passing `-s WASM=1`.

Run `docker pull  trzeci/emscripten` to update and try again plz?

jakearchibald:
That did the trick. Is it worth adding `docker pull trzeci/emscripten` as a `postinstall`?

surma:
An update is usually around 500MB. Not sure I wanna force that onto everyone automatically.

jakearchibald:
What about a warning if `trzeci/emscripten` is out of date?

surma:
I’ll look into that

jakearchibald:
I've now built all the codecs and got the same results as the stuff checked in \o/

surma:
It’s actually surprisingly hard. I can’t find a way to check for a new version on the Docker Hub. Can we punt on the check?

surma:
@jakearchibald Done :)

PR #193, opened 2018-10-12
Lazy-loading the main part of the app

jakearchibald:
**Do not merge** but please review. Not based on master.


src/components/Output/index.tsx
```diff
@@ -44,62 +44,88 @@ export default class Output extends Component<Props, State> {
...
+    }
+
+    if (this.props.originalImage !== prevProps.originalImage && this.pinchZoomLeft) {
+      // New image? Reset the pinch-zoom.
+      this.pinchZoomLeft.setTransform({
+        allowChangeEvent: true,
+        x: 0,
+        y: 0,
+        scale: 1,
+      });
```
jakearchibald: I snuck this fix in. Otherwise, it caused images to disappear to the side if the dimensions were particularly different.

src/components/compress/index.tsx
```diff
@@ -0,0 +1,408 @@
...
+        // Ditch previous encodings
+        const downloadUrl = this.state.images[i].downloadUrl;
+        if (downloadUrl) URL.revokeObjectURL(downloadUrl!);
+
+        newState = cleanMerge(newState, `images.${i}`, {
+          preprocessed: undefined,
+          file: undefined,
+          downloadUrl: undefined,
+          data: undefined,
+        });
```
jakearchibald: This fixes the issue where you'd see two different halves of an image when loading a new image.

PR #195, opened 2018-10-12
Fix <range-input disabled>

### LINKED ISSUE #142: Range input: implement 'disabled'.
In the WebP options "Auto adjust filter strength" disables the subsequent range. The range input in https://github.com/GoogleChromeLabs/squoosh/pull/136 doesn't support this.

developit:
This fixes the `.disabled` property reflection for range-input. Previously, only the `disabled` attribute worked.

Fixes #142.

### DIFF
--- a/src/custom-els/RangeInput/index.ts
+++ b/src/custom-els/RangeInput/index.ts
@@ -3,8 +3,8 @@ import './styles.css';
 
 const RETARGETED_EVENTS = ['focus', 'blur'];
 const UPDATE_EVENTS = ['input', 'change'];
-const REFLECTED_PROPERTIES = ['name', 'min', 'max', 'step', 'value', 'disabled'];
-const REFLECTED_ATTRIBUTES = ['name', 'min', 'max', 'step', 'value', 'disabled'];
+const REFLECTED_PROPERTIES = ['name', 'min', 'max', 'step', 'value'];
+const REFLECTED_ATTRIBUTES = ['name', 'min', 'max', 'step', 'value'];
 
 class RangeInputElement extends HTMLElement {
   private _input = document.createElement('input');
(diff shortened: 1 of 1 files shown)


jakearchibald:
It looks like `disabled` never reaches the underlying input. Also, I can't figure out why it didn't work before. The property getter/setter reflected the value from the input.

developit:
The styling relies on `range-input[disabled]`.

jakearchibald:
This PR may fix the styling but it breaks the behaviour 😄 .

Feels like the original problem is with `if (attributeValue === null) continue;` in `_reflectAttributes()`, and it should remove the attribute instead? However, from memory, I think that broke something, so if you do that be sure to test.

developit:
Yeah I'm kinda stumped here - it seems like the (prior to this PR) reflection should have worked since it does:

```js
this._input.disabled = true
this.setAttribute('disabled', this._input.getAttribute('disabled'))
```

...which should be equivalent to `setAttribute('disabled', '')`...

surma:
> `this.setAttribute('disabled', this._input.getAttribute('disabled')`
> ...which should be equivalent to setAttribute('disabled', '')...

For the `disabled=true` case that is correct, but not for `disabled=false`.

+1 to Jake. I think the error is in line 97:

`if (attributeValue === null) continue;`

should do an `removeAttribute()`.

surma:
Side note: The `REFLECTED_X` array and the method `_reflectAttributes()` should really be called `_passthroughAttributes()` or something. This isn’t really the normal reflection, isn’t it?

surma:
We had a rogue for-in loop instead of a for-of loop. And the line Jake and I pointed out.

screen shot 2018-10-19 at 1 34 28 pm

developit:
Ahh wow yeah that'd be from the back and forth we did on the reflection approach. The hasAttribute thing is fine, though it doesn't seem like that was the cause of the .disabled bug.

surma:
The difference is the explicit "removeAttribute" call. See previous commit.

PR #196, opened 2018-10-12
Remove firebase travis deploy (for now)

### LINKED ISSUE #151: Remove Travis and Firebase
It appears that we like netlify. We should clean up the travis config and firebase config if we are not using.

developit:
Fixes #151 (because we're keeping Travis for tests)


jakearchibald:
What about `.firebaserc`?

developit:
Whoop, good call. rmeoved.

PR #197, opened 2018-10-12
Lazy-loading the main part of the app

jakearchibald:
Also improving the transition from intro to compressor.

### DIFF
--- a/src/components/App/index.tsx
+++ b/src/components/App/index.tsx
@@ -1,46 +1,15 @@
 import { h, Component } from 'preact';
 
-import { bind, linkRef, Fileish, blobToImg, drawableToImageData, blobToText } from '../../lib/util';
+import { bind, linkRef, Fileish } from '../../lib/initial-util';
 import * as style from './style.scss';
-import Output from '../Output';
-import Options from '../Options';
 import { FileDropEvent } from './custom-els/FileDrop';
 import './custom-els/FileDrop';
-import ResultCache from './result-cache';
-
-import * as quantizer from '../../codecs/imagequant/quantizer';
-import * as optiPNG from '../../codecs/optipng/encoder';
-import * as resizer from '../../codecs/resize/resize';
-import * as mozJPEG from '../../codecs/mozjpeg/encoder';
-import * as webP from '../../codecs/webp/encoder';
-import * as identity from '../../codecs/identity/encoder';
-import * as browserPNG from '../../codecs/browser-png/encoder';
-import * as browserJPEG from '../../codecs/browser-jpeg/encoder';
-import * as browserWebP from '../../codecs/browser-webp/encoder';
-import * as browserGIF from '../../codecs/browser-gif/encoder';
-import * as browserTIFF from '../../codecs/browser-tiff/encoder';
-import * as browserJP2 from '../../codecs/browser-jp2/encoder';
-import * as browserBMP from '../../codecs/browser-bmp/encoder';
-import * as browserPDF from '../../codecs/browser-pdf/encoder';
-import {
-    EncoderState,
-    EncoderType,
-    EncoderOptions,
-    encoderMap,
-} from '../../codecs/encoders';
 import SnackBarElement from '../../lib/SnackBar';
 import '../../lib/SnackBar';
-
-import {
-  PreprocessorState,
-  defaultPreprocessorState,
-} from '../../codecs/preprocessors';
-
-import { decodeImage } from '../../codecs/decoders';
-import { cleanMerge, cleanSet } from '../../lib/clean-modify';
 import Intro from '../intro';
 
-type Orientation = 'horizontal' | 'vertical';
+// This is imported for TypeScript only. It isn't used.
+import Compress from '../compress';
 
 export interface SourceImage {
   file: File | Fileish;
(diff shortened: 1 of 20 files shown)


src/components/Output/custom-els/PinchZoom/index.ts
```diff
@@ -272,10 +272,7 @@ export default class PinchZoom extends HTMLElement {
   private _stageElChange() {
     this._positioningEl = undefined;
 
-    if (this.children.length === 0) {
-      console.warn('There should be at least one child in <pinch-zoom>.');
-      return;
-    }
+    if (this.children.length === 0) return;
```
jakearchibald: This warning was just making the console noisy.

src/components/Output/index.tsx
```diff
@@ -3,16 +3,16 @@ import PinchZoom, { ScaleToOpts } from './custom-els/PinchZoom';
...
 import './custom-els/TwoUp';
 import * as style from './style.scss';
-import { bind, shallowEqual, drawDataToCanvas, linkRef } from '../../lib/util';
+import { bind, linkRef } from '../../lib/initial-util';
+import { shallowEqual, drawDataToCanvas } from '../../lib/util';
 import { ToggleIcon, AddIcon, RemoveIcon } from '../../lib/icons';
 import { twoUpHandle } from './custom-els/TwoUp/styles.css';
 
 interface Props {
+  originalImage?: ImageData;
```
jakearchibald: Bringing this in so we can detect when the original image changes (it's used to reset the pinch zoom).

src/components/Output/index.tsx
```diff
@@ -44,62 +44,88 @@ export default class Output extends Component<Props, State> {
...
+    return props.leftCompressed || props.originalImage;
+  }
+
+  private rightDrawable(props: Props = this.props): ImageData | undefined {
+    return props.rightCompressed || props.originalImage;
+  }
+
   @bind
-  toggleBackground() {
+  private toggleBackground() {
```
jakearchibald: I'm making all internal methods private, as it makes it easier to spot methods that aren't used.

src/components/compress/index.tsx
```diff
@@ -0,0 +1,408 @@
+import { h, Component } from 'preact';
```
jakearchibald: This is the old `App/index.tsx`

src/components/compress/index.tsx
```diff
@@ -0,0 +1,408 @@
...
+        // Ditch previous encodings
+        const downloadUrl = this.state.images[i].downloadUrl;
+        if (downloadUrl) URL.revokeObjectURL(downloadUrl!);
+
+        newState = cleanMerge(newState, `images.${i}`, {
+          preprocessed: undefined,
+          file: undefined,
+          downloadUrl: undefined,
+          data: undefined,
+        });
```
jakearchibald: This fixes the issue where you'd see two different halves of an image when loading a new image.

src/components/compress/index.tsx
developit: anything particularly meaningfully changed here? looks the same at first glance.

src/components/App/index.tsx
```diff
@@ -1,187 +1,44 @@
...
+    Compress: undefined,
   };
 
   snackbar?: SnackBarElement;
-  readonly encodeCache = new ResultCache();
 
   constructor() {
     super();
+
+    import('../compress').then((module) => {
```
developit: Nit: should we handle errors here? It could just use the snackbar to display something about the failure.

src/components/App/index.tsx
```diff
@@ -191,232 +48,37 @@ export default class App extends Component<Props, State> {
...
-    });
-
-    this.setState({ images });
+  private async onIntroPickFile(file: File | Fileish) {
+    this.setState({ file });
   }
 
   @bind
-  showError (error: string) {
+  private showError(error: string) {
```
developit: Thought: could/should we lazyily load the snackbar too? it's not critical UI. Not sure if we're hoping this initial bundle payload is small enough to inline..

src/components/compress/index.tsx
developit: Saw that. Wasn't sure what we wanted to do, seemed like there were arguments for either case (show old data vs rerender and wait for new).

src/components/App/index.tsx
```diff
@@ -191,232 +48,37 @@ export default class App extends Component<Props, State> {
...
-      file,
-      data,
-      preprocessed,
-      downloadUrl: URL.createObjectURL(file),
-      loading: images[index].loadingCounter !== loadingCounter,
-      loadedCounter: loadingCounter,
-    });
-
-    this.setState({ images });
+  private async onIntroPickFile(file: File | Fileish) {
```
developit: nit: not `async`

src/components/compress/index.tsx
jakearchibald: https://gist.github.com/jakearchibald/7a52541725f34f5f10fff1596febde0d here's a diff

src/components/compress/index.tsx
jakearchibald: It currently shows the old data until the new file is decoded, then it shows the new file while waiting for the encoding.

src/components/App/index.tsx
jakearchibald: Yeah, fair point.

src/components/App/index.tsx
jakearchibald: Good idea. I'll see how much it adds to the bundle. Currently I think we should inline the CSS but not the JS (assuming we server render enough HTML for a meaningful render).

src/components/App/index.tsx
jakearchibald: lol oops

src/components/App/index.tsx
jakearchibald: CSS, JS
With this PR: 5.5k, 8.1k.
Without snackbar: 5.1k, 7.7k.

So that's a 0.8k saving, which we may lose by adding code to load the snackbar async (e.g. what if an error happens before the snackbar is ready).

I don't think it's worth it for the complexity jump. We'd save more by dropping the font (which I'm getting pretty tempted by).

PR #198, opened 2018-10-13
Two workers & worker termination

jakearchibald:
Building on the ideas in #192, but ensuring we only have two workers on the go.

This involved a lot of refactoring, so I'm really sorry that this PR ballooned. Free free to throw it back at me (if you've got a better way of doing it).

I'll leave some comments to try and make things easier.

@surma it doesn't look like free() is working correctly for mozjpeg. Could you take a look? I updated `example.html` to show the issue. I was getting "out of bounds" errors, but now it just seems to screw up the second image.


jakearchibald:
I changed the shape of the encoders/preprocessors. `encoder-meta.ts` contains stuff like default options, types, basically things we're happy being on the main thread. `encoder.ts` becomes the wasm stuff.

src/codecs/imagequant/processor.ts
```diff
@@ -0,0 +1,43 @@
+import imagequant, { QuantizerModule } from '../../../codecs/imagequant/imagequant';
+import wasmUrl from '../../../codecs/imagequant/imagequant.wasm';
+import { QuantizeOptions } from './processor-meta';
+
+let emscriptenModule: Promise<QuantizerModule>;
```
jakearchibald: I changed how we do these types of files a bit. Instead of having a class, I just treat it as a module. The emscripten module is initialised on first use.

src/codecs/processor-worker.ts
```diff
@@ -0,0 +1,41 @@
+import { expose } from 'comlink';
```
jakearchibald: This is the worker. It exposes a method for everything we workerise, but it loads the bulk of the code lazily.

src/codecs/processor.ts
```diff
@@ -0,0 +1,194 @@
+import { proxy } from 'comlink';
```
jakearchibald: This is the main-thread part of processor-worker. It exposes all the image processing methods, and cancels any pending jobs.

surma:
Not to go all “It works on my machine” on you, but it works on my machine? I.e. your modified `example.html` runs without any errors for me.

jakearchibald:
Huh, does the 2nd image not look all fucked? I guess I'll show you next
week.

src/codecs/imagequant/processor.ts
```diff
@@ -0,0 +1,43 @@
...
+        if (url.endsWith('.wasm')) return wasmUrl;
+        return url;
+      },
+      onRuntimeInitialized() {
+        // An Emscripten is a then-able that, for some reason, `then()`s itself,
+        // causing an infite loop when you wrap it in a real promise. Deleten the `then`
+        // prop solves this for now.
+        // See: https://github.com/kripken/emscripten/blob/incoming/src/postamble.js#L129
+        // TODO(surma@): File a bug with Emscripten on this.
+        delete (m as any).then;
```
developit: I know this isn't part of the PR, but could we change this so it doesn't require the delete?  Looking at the source, the thennable never fails, so we could just do:

```js
m.then(resolve)
```

developit: Also I bet we could extract the initialization into a helper method:

```js
// initWasmModule(imagequant, wasmUrl)
function initWasmModule(fn, wasmUrl) {
  return new Promise((resolve) => {
    const m = fn({
      // Just to be safe, don’t automatically invoke any wasm functions
      noInitialRun: false,
      locateFile(url: string): string {
        // Redirect the request for the wasm binary to whatever webpack gave us.
        if (url.endsWith('.wasm')) return wasmUrl;
        return url;
      },
      onRuntimeInitialized() {
        m.then(resolve);
      }
    });
  });
}
```

src/codecs/imagequant/processor.ts
```diff
@@ -0,0 +1,43 @@
...
+import wasmUrl from '../../../codecs/imagequant/imagequant.wasm';
+import { QuantizeOptions } from './processor-meta';
+
+let emscriptenModule: Promise<QuantizerModule>;
+
+function initModule(): Promise<QuantizerModule> {
+  return new Promise((resolve) => {
+    const m = imagequant({
+      // Just to be safe, don’t automatically invoke any wasm functions
+      noInitialRun: false,
```
developit: Just noticed, this seems to do the opposite of what the comment above indicates.

`no initial run = false` --> `initial run = true`

src/codecs/processor-worker.ts
developit: I like this a lot better than the per-encoder approach.

src/codecs/processor.ts
```diff
@@ -0,0 +1,194 @@
...
+
+      descriptor.value = async function (this: Processor, ...args: any[]) {
+        this._latestJobId += 1;
+        const jobId = this._latestJobId;
+        this.abortCurrent();
+
+        if (needsWorker) self.clearTimeout(this._workerTimeout);
+
+        if (!this._worker && needsWorker) {
+          // Webpack's worker loader does magic here.
```
developit: nit: worker-plugin

src/codecs/imagequant/processor.ts
jakearchibald: > `m.then(resolve)`

Yeah, that seems less weird.

> Also I bet we could extract the initialization into a helper method

Agreed. Good point.

src/codecs/imagequant/processor.ts
jakearchibald: @surma ^

jakearchibald:
@developit 

> `m.then(resolve)`

Unfortunately that doesn't work. The module resolves with itself, and `resolve()` will attempt to unwrap the thennables, which it can't because Emscripten is doing extra weird stuff.

src/codecs/util.ts
```diff
@@ -0,0 +1,27 @@
...
+      noInitialRun: false,
+      locateFile(url: string): string {
+        // Redirect the request for the wasm binary to whatever webpack gave us.
+        if (url.endsWith('.wasm')) return wasmUrl;
+        return url;
+      },
+      onRuntimeInitialized() {
+        // An Emscripten is a then-able that resolves with itself, causing an infite loop when you
+        // wrap it in a real promise. Delete the `then` prop solves this for now.
+        // https://github.com/kripken/emscripten/issues/5820
```
surma: Thanks!

src/codecs/decoders.ts
```diff
@@ -1,47 +1,21 @@
-import * as wasmWebp from './webp/decoder';
-import * as browserWebp from './webp/decoder';
```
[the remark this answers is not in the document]
jakearchibald: Feels like a drop in complexity while we only have one decoder. This file is smaller & IMO easier to follow as a result.

jakearchibald:
@surma feedback should be addressed now

PR #212, opened 2018-10-15
Fix for loading SVG in Safari.

jakearchibald:


src/lib/util.ts
```diff
@@ -90,16 +90,22 @@ export async function blobToImg(blob: Blob): Promise<HTMLImageElement> {
...
+        // https://bugs.webkit.org/show_bug.cgi?id=188347
+        await loaded;
+      }
     } else {
-      // Main thread decoding :(
-      await new Promise((resolve, reject) => {
-        img.onload = () => resolve();
-        img.onerror = () => reject(Error('Image loading error'));
-      });
+      await loaded;
```
surma: Looking at the control flow here, I _think_ we can simplify and remove the entire `if` statement.

```js
try {
	await img.decode();
} 
catch(e) {}
finally {
	await loaded;
}

```

jakearchibald: I think it can be even simpler than that. I think I was having a bad code day.

jakearchibald:
@surma happy with this (imo) tidier version?

PR #226, opened 2018-10-20
Startup optimizations

### LINKED ISSUE #221: Load main script in head, deferred
It's currently async & at the bottom of the body.

developit:
- Prerender ("static SSR")
- Critical CSS inlining (ends up just being the intro UI, seems to work well)

I played with both to get the HTML payload to 2.27kB.

Also fixes #221. `<script defer>` in head actually performed worse than `<script async>` at the end of body with a preload in head, though I didn't test exhaustively (mostly successive LH runs).


src/components/intro/style.scss
```diff
@@ -47,7 +49,7 @@
 }
 
 .open-image-guide {
-  font: 300 11vw intro-text;
+  font: 300 11vw intro-text,helvetica,arial;
```
mathiasbynens: FWIW, `sans-serif` has the same effect as `helvetica,arial`.

surma:
I think we should inline a subset font for the first render, mainly because of two reasons:

- Lots of web devs get told “Use our CI’s webfont or bust”. We have the chance to them that you can do that without paying a ridiculous cost.
- We have the budget. Our TTFCP/TTI is amazing, we have available size budget to inline the font. Yes we could be smaller, but that’s what budgets are for.

src/components/intro/style.scss
developit: iirc just helvetica would as well, since windows substitutes Arial anyway. good catch.

developit:
I'm fine inlining the font or removing it. TBH I'd rather spend those bytes on inlining handlers for drop+click that enqueue to get our TTI down.

developit:
Alright here's the latest result with the fonts inlined and scripts back to `defer` in the head.
I think we're good?
screen shot 2018-10-29 at 10 48 38 am

src/components/intro/style.scss
```diff
@@ -10,6 +11,7 @@
   font-family: 'intro-text';
   font-style: normal;
   font-weight: 500;
+  font-display: fallback;
```
jakearchibald: Does this do anything now we're inlining?

src/components/intro/index.tsx
```diff
@@ -89,7 +89,7 @@ export default class Intro extends Component<Props, State> {
         <div>
           <div class={style.logoSizer}>
             <div class={style.logoContainer}>
-              <img src={logo} class={style.logo} alt="Squoosh" />
+              <img src={logo} class={style.logo} alt="Squoosh" decoding="async" />
```
jakearchibald: I don't think this does anything for SVG… but it isn't doing any harm.

src/index.ts
```diff
@@ -1,9 +1,21 @@
 declare module '@webcomponents/custom-elements';
 
-(async function () {
-  if (!('customElements' in self)) {
-    await import('@webcomponents/custom-elements');
-  }
-
+if (typeof PRERENDER === 'boolean') {
```
jakearchibald: Add a comment here to explain what's happening?

Is it correct that this block should happen if `PRERENDER` is `false`?

src/index.ts
```diff
@@ -1,9 +1,21 @@
...
+  // @ts-ignore
+  window.customElements = {
+    define() {},
+  };
```
jakearchibald: Is this because it's running in jsdom?

src/index.ts
jakearchibald: fwiw `'PRERENDER' in self` would be more consistent with the `else if`.

src/index.ts
developit: yup.

src/index.ts
developit: TBH this is a workaround. Critters is supposed to set `PRERENDER` to `true` during prerendering and `false` for the runtime build, but the latter isn't working.  I'll try to get that fixed upstream and correct it here. It should just be as simple as:

```
if (PRERENDER) {

}
```

Webpack's definePlugin doesn't support static property existence checks, so `in` doesn't work here.

src/index.ts
developit: I'll add a comment.

src/components/intro/style.scss
developit: The data URLs do take a few ms according to lighthouse. I guess `block` might better suit that?

jakearchibald: Yeah make it block. That covers browsers that might allow a swap (Edge, maybe?)

developit:
update: fixed the readfile stuff, but I'm still struggling to get both compilers producing the same contenthashes. Taking forever.

developit:
Root cause turned out to be this bug that was fixed last month. Updating and tweaking a couple things fixed hashes and thus the build:
https://github.com/webpack-contrib/mini-css-extract-plugin/issues/281

developit:
final result

screen shot 2018-10-31 at 12 40 07 am

src/index.ts
developit: I moved all this shimming into prerender-loader where it belonged. Now our index is super clean!

src/index.ts
developit: Re-affirmed that it is in fact not possible to modify parser plugins once the build has begun, making it impossible for prerender-loader to set `PRERENDER=true` / `PRERENDER=false`.  I've stripped out all the conditional logic instead, since it belonged in prerender-loader itself anyway. We're clean and there's no more conditional stuff!

src/components/intro/index.tsx
developit: @jakearchibald - I'd thought about having these images get skipped when prerendering. Right now because they're sent as static HTML, they kick of 4 requests that extend our "load" time out by quite a bit (which is more of a metrics issue, but still).  Thoughts?

src/components/intro/index.tsx
```diff
@@ -110,7 +110,7 @@ export default class Intro extends Component<Props, State> {
                   <div class={style.demo}>
                     <div class={style.demoImgContainer}>
                       <div class={style.demoImgAspect}>
-                        <img class={style.demoIcon} src={demo.iconUrl} alt=""/>
+                        <img class={style.demoIcon} src={demo.iconUrl} alt="" decoding="async" />
```
developit: @jakearchibald - I'd thought about having these images get skipped when prerendering. Right now because they're sent as static HTML, they kick of 4 requests that extend our "load" time out by quite a bit (which is more of a metrics issue, but still). Thoughts?

jakearchibald: The only metrics I care about are "time until useful pixels are on the screen" and "time until the user can perform the first interaction".

If one of Google's many acronyms reports the wrong time for those, it's their bug, not our's 😄 (we should report it though).

jakearchibald:
@developit that bug sounded like a nightmare to track down. Thanks for working through it.

developit:
@jakearchibald I put some time into improving Critters, and it now strips inlined critical styles out of the lazy-loaded noncritical stylesheets.  I also added a threshold option so it doesn't do something silly like lazy-load 300b of CSS and instead just inlines the rest.  With those in-place, I think everything is optimal here:

screen shot 2018-11-03 at 8 53 34 pm

Our `index.html` is 6.84kB with 100% of the CSS and fonts inlined to keep layout stable.

developit:
Alright, fairly happy with this now:
https://www.webpagetest.org/result/181104_90_329591be7c50cebf5e4564c5494638a2/

src/index.ts
```diff
@@ -1,9 +1,9 @@
 declare module '@webcomponents/custom-elements';
 
-(async function () {
-  if (!('customElements' in self)) {
-    await import('@webcomponents/custom-elements');
-  }
-
+if (!('customElements' in self)) {
```
jakearchibald: If this is fragile (as in, it needs to be written in this way else webpack or whatever doesn't do the right thing), add a comment stating that, else I'll probably break it again 😀

jakearchibald:
In fact, the more I state at the webpagetest results, we could save half a second by inlining the first-interaction JS

jakearchibald:
I've rebased this onto the service worker branch for now

jakearchibald:
lol aaaaand back to master

jakearchibald:
https://www.webpagetest.org/result/181108_DQ_daaae433b6dd49e4cd0dea126b3e1983/ here's the result of inlining. It seems to regress performance.

PR #227, opened 2018-10-28
Mobile ui

jakearchibald:
**Not based on master**.

This is the MVP mobile UI. Testing on iOS is appreciated, as I don't have a device here.

I'm considering preventing regular pinch-zoom behaviour, which I generally don't like, but given the app already has pinch zoom, it might improve UX by avoiding two levels of pinch zoom. Thoughts?

Again, really sorry about the size of this PR. I'll go through and add comments to justify changes.


src/components/Options/index.tsx
```diff
@@ -76,39 +70,18 @@ interface Props {
...
-      clearTimeout(this.loadingTimeoutId);
-      this.setState({ showLoadingState: false });
-    } else if (!prevProps.loading && this.props.loading) {
-      // Just started loading
-      this.loadingTimeoutId = self.setTimeout(
-        () => this.setState({ showLoadingState: true }),
-        loadingReactionDelay,
-      );
-    }
-  }
```
jakearchibald: A lot of this stuff was needed for the "results" view (as in "154kb 91% smaller") which I've moved into its own component.

src/components/compress/custom-els/MultiPanel/index.ts
```diff
@@ -1,19 +1,75 @@
...
 
-function getClosestHeading(el: Element) {
-  const closestEl = el.closest('multi-panel > *');
-  if (closestEl && closestEl.classList.contains('panel-heading')) {
-    return closestEl;
+interface CloseAllOptions {
+  exceptFirst?: boolean;
+}
+
+const openOneOnlyAttr = 'open-one-only';
```
jakearchibald: New feature in multi-panel. This attribute ensures a maximum of zero panels can be open at once.

src/components/compress/custom-els/MultiPanel/index.ts
```diff
@@ -1,19 +1,75 @@
...
+
+function getClosestHeading(el: Element): HTMLElement | undefined {
+  // Look for the child of multi-panel, but stop at interactive elements like links & buttons
+  const closestEl = el.closest('multi-panel > *, a, button');
```
jakearchibald: This prevents the panel toggling when you click things like the download button.

src/components/compress/custom-els/MultiPanel/index.ts
```diff
@@ -1,19 +1,75 @@
...
+async function close(heading: HTMLElement) {
+  const content = heading.nextElementSibling as HTMLElement;
+
+  // if there is no content, nothing to expand
+  if (!content) return;
+
+  const from = content.getBoundingClientRect().height;
+
+  heading.removeAttribute('content-expanded');
+  content.setAttribute('aria-expanded', 'false');
```
jakearchibald: I've added animation to the component, but also changed how the attributes work. `aria-expanded` stays on the content as before, but `content-expanded` is added to the heading. This makes it easier to style the heading based on the expanded state of the content.

src/components/compress/custom-els/MultiPanel/index.ts
```diff
@@ -1,19 +1,75 @@
...
+
+  // Wait a microtask so other calls to open/close can get the final sizes.
+  await null;
```
jakearchibald: This is horrible and I'm open to better ideas.

src/components/compress/style.scss
```diff
@@ -3,20 +3,72 @@
...
+
+.multi-panel {
+  position: relative;
+  display: flex;
+  flex-flow: column;
 
-  &.vertical {
-    // TODO: make the mobile view work
-    background: red;
+  // Reorder so headings appear after content:
```
jakearchibald: The multi-panel expects header-content-header-content, but design-wise we want the header at the bottom.

src/components/compress/custom-els/MultiPanel/index.ts
surma: Eh, works for now. The proper solution would be an orchestrator, which I’m sure no-one wants to write rn

PR #229, opened 2018-10-28
Copy to side ui

jakearchibald:
Trying a different button position. What do you think?

screen shot 2018-10-28 at 11 37 56

### DIFF
--- a/src/components/results/style.scss
+++ b/src/components/results/style.scss
@@ -16,7 +16,7 @@
 
 .results {
   display: grid;
-  grid-template-columns: 1fr auto auto;
+  grid-template-columns: [text] 1fr [copy-button] auto [download-button] auto;
   background: rgba(0, 0, 0, 0.9);
   font-size: 1rem;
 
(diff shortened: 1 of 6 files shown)


jakearchibald:
Preview URL https://deploy-preview-229--squoosh.netlify.com/

mustafa-x:
can we add an undo function? so if someone copies settings across, they can undo them

jakearchibald:
The technical side of this is easy, but I'm not sure what the UI would be.

jakearchibald:
Maybe we could have a snackbar that says "Settings copied" with an undo button

mustafa-x:
We should a pill that comes up when a user undoes or redoes something. That animates then disappears. (see bottom left|)

example

mustafa-x:
Also I think separating the buttons download and copy across would work better.

mustafa-x:
I was commenting on the first design :)

developit:
undo notification could be a toast with 500ms timeout.

jakearchibald:
@developit exactly that in https://github.com/GoogleChromeLabs/squoosh/pull/233 😄

jakearchibald:
(well, I think it's a few seconds rather than 500ms)

mustafa-x:
screen shot 2018-10-30 at 1 36 44 pm

Might be worth disabling the button if both sides are the same, or after it has been duplicated.

mustafa-x:
I was reviewing https://deploy-preview-233--squoosh.netlify.com/

jakearchibald:
Cool. Those changes are over in https://github.com/GoogleChromeLabs/squoosh/pull/233.

The "233" in the URL refers to the pull request number.

jakearchibald:
@mustafa-x you can get to the result of a PR in the "Show all checks" section:

screen shot 2018-10-30 at 13 55 56

src/components/results/style.scss
```diff
@@ -16,31 +16,55 @@
 
 .results {
   display: grid;
-  grid-template-columns: 1fr auto;
+  grid-template-columns: [text] 1fr [copy-button] auto [download-button] auto;
```
surma: YEAAAAH, grid featuerzzz 😎

PR #233, opened 2018-10-29
Snackbar defaults & copy undo

jakearchibald:
* Snackbar default timeout removed & replaced with "dismiss" button (otherwise, large WASM errors were lost, and it'd be nice if folks could copy those for bug reports).
* Copying settings across can now be undone.


jakearchibald:
TODO: fix the build. Oops.

jakearchibald:
Oh, I just told Netlify to rebuild and it worked, so I guess it wasn't my fault.

PR #236, opened 2018-10-30
Fixing lossless slider for webP. Previously you couldn't select "9" :D

jakearchibald:
Fixing a little bug with the slider.


jakearchibald:
@surma `losslessPresets` contains 10 entries which map to our "effort" slider in lossless mode. Because 'quality' 90 appears twice, it was alway matching on the first, but never the second.

PR #240, opened 2018-11-02
_headers & _redirects generation

developit:
Generate `_headers` and `_redirects` by passing assets through ejs templates.

Note: based against `service-worker` branch, wait to merge.


_headers.ejs
```diff
@@ -0,0 +1,30 @@
+/
+  Cache-Control: nocache
```
jakearchibald: `no-cache`

_headers.ejs
```diff
@@ -0,0 +1,30 @@
...
+
+/serviceworker.js
+  Cache-Control: nocache
```
jakearchibald: `no-cache`

_redirects.ejs
```diff
@@ -0,0 +1 @@
+/*    /index.html   200
```
jakearchibald: Seems better to do:

```
/    /index.html
/*  / 307
```

That way we have a canonical URL for the app.

_headers.ejs
```diff
@@ -0,0 +1,30 @@
...
+  Cache-Control: max-age=31536000
+/:a.png
+  Cache-Control: max-age=31536000
+/:a.svg
+  Cache-Control: max-age=31536000
+
+# List of build assets at time of generation:
+<%_ for (let file of files) { _%>
+#  <%= file %>
+<%_ } _%>
```
jakearchibald: I guess we can remove the generated for now, until we need something more complicated?

_headers.ejs
developit: ack.

_redirects.ejs
developit: this is the default behavior IIRC.

_headers.ejs
developit: I was waiting until we determined if the URL patterns were functioning correctly.

developit:
@jakearchibald up for re-review

jakearchibald:
To be fair, if that rule was working, we'd redirect on all our subresources. I think I'll just remove the redirects & merge this.

jakearchibald:
And that's the winner!

PR #245, opened 2018-11-06
Add analytics script (fixes #174)

### LINKED ISSUE #174: Add analytics

surma:


jakearchibald:
Are we saving much by putting `analytics.init` in its own chunk? I wonder if it vs the module definition in webpack are around the same.

We could even put it in the main bundle rather than the first interaction bundle, WDYT?

surma:
No strong opinion. I just did “keep main bundle impact as small as possible, defer everything”. But you are right, the actual code is very small and could go into the main bundle.

src/index.ts
```diff
@@ -6,4 +6,16 @@ declare module '@webcomponents/custom-elements';
   }
 
   require('./init-app.tsx');
+
+  // Initialize the command queue in case analytics.js hasn't loaded yet.
+  window.ga = window.ga || ((...args) => (ga.q = ga.q || []).push(args));
```
mathiasbynens: I don’t if/how this gets transpiled by TypeScript, but have you made sure that this actually works? (You can just check the DevTools Network tab to see if a beacon goes out.)

I’m asking because of https://github.com/v8/v8.dev/blob/d6805e0423c7301d019a5fba3c19bc1c8c304cda/src/_js/main.mjs#L45-L61, which seems like a similar scenario that definitely surprised me at the time.

surma: I got this from [@philipwalton’s blog post](https://philipwalton.com/articles/the-google-analytics-setup-i-use-on-every-site-i-build/).

I didn’t realize this wasn’t the official snippet. But you are right, the param spread might actually break in some browsers. I’ll fix that. Thanks!

mathiasbynens: Note that the issue I ran into was not even related to browser support. Glad to see that `ga.js` apparently doesn’t care about whether something is an `arguments` object or an actual array (unlike `gtag`)!

src/components/results/index.tsx
```diff
@@ -58,6 +58,16 @@ export default class Results extends Component<Props, State> {
     this.props.onCopyToOtherClick();
   }
 
+  @bind
+  onDownload() {
+    ga('send', 'event', 'compression', 'download', {
```
philipwalton: Suggestion: you might consider wrapping this in a `requestIdleCallback()`

philipwalton: You don't have an event label here. It might be interesting to set the file type as the event label. Then you could report on which file types got the most savings.

PR #246, opened 2018-11-06
Android tablet bugs

jakearchibald:
Fixing a few little issues I saw while testing a Pixel C.

### DIFF
--- a/src/components/Options/style.scss
+++ b/src/components/Options/style.scss
@@ -54,4 +54,5 @@ $horizontalPadding: 15px;
 .options-scroller {
   overflow-x: hidden;
   overflow-y: auto;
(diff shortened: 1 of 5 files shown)


src/components/Options/style.scss
```diff
@@ -54,4 +54,5 @@ $horizontalPadding: 15px;
 .options-scroller {
   overflow-x: hidden;
   overflow-y: auto;
+  -webkit-overflow-scrolling: touch;
```
surma: Android does something with this? 😮

jakearchibald: Ah, no sorry, I snuck that in as an iOS fix.

PR #249, opened 2018-11-08
Enhanced offline

jakearchibald:
Adding things like an "update available" notification.


src/changelog.json
```diff
@@ -0,0 +1,3 @@
+{
+  "0.0.1": ["Pre-release"]
+}
```
jakearchibald: The first entry in the changelog is the current version. I want to find a way to display the changes between versions, but UI takes me a long time and I don't have enough of it before CDS. Will look into it after.

src/components/App/index.tsx
```diff
@@ -12,6 +12,15 @@ import '../custom-els/LoadingSpinner';
...
 import Compress from '../compress';
 
+const compressPromise = import(
+  /* webpackChunkName: "main-app" */
+  '../compress',
+);
+const offlinerPromise = import(
+  /* webpackChunkName: "offliner" */
+  '../../lib/offliner',
+);
```
jakearchibald: I moved the service worker stuff into a lazy-loaded module to keep our first interaction size down.

src/lib/SnackBar/index.ts
```diff
@@ -8,12 +8,9 @@ export interface SnackOptions {
 function createSnack(message: string, options: SnackOptions): [Element, Promise<string>] {
   const {
     timeout = 0,
-    actions = [],
+    actions = ['dismiss'],
```
jakearchibald: Snuck this in. Realised it doesn't make sense to have a snack that doesn't have a dismiss button.

webpack.config.js
```diff
@@ -141,12 +142,6 @@ module.exports = function (_, env) {
           exclude: nodeModules,
           loader: 'ts-loader'
         },
-        {
-          test: /\.jsx?$/,
-          loader: 'babel-loader',
-          // Don't respect any Babel RC files found on the filesystem:
-          options: Object.assign(readJson('.babelrc'), { babelrc: false })
-        },
```
jakearchibald: Pretty sure we don't need this.

webpack.config.js
```diff
@@ -242,30 +237,22 @@ module.exports = function (_, env) {
...
-      // Here we replace the constructor and message with a static throw, leaving the message to be DCE'd.
-      // This is useful since it shows the message in SourceMapped code when debugging.
-      isProd && new ReplacePlugin({
-        include: /babel-helper$/,
-        patterns: [{
-          regex: /throw\s+(?:new\s+)?((?:Type|Reference)?Error)\s*\(/g,
-          value: (s, type) => `throw 'babel error'; (`
-        }]
-      }),
-
```
jakearchibald: Pretty sure we don't need this.

src/lib/offliner.ts
```diff
@@ -0,0 +1,91 @@
...
+  await updateReady(reg);
+
+  // Ask the user if they want to update.
+  const result = await showSnack('Update available', {
+    actions: ['reload', 'dismiss'],
+  });
+
+  // Tell the waiting worker to activate, this will change the controller and cause a reload (see
+  // 'controllerchange')
+  if (result === 'reload') skipWaiting();
```
surma: So just wondering: If I have 1 old and 1 new tab with Squoosh, and the new one shows me the reload prompt and I click reload, will the old tab reload as well or could we be breaking the old tab?

webpack.config.js
```diff
@@ -19,6 +18,8 @@ function readJson (filename) {
   return JSON.parse(fs.readFileSync(filename));
 }
 
+const VERSION = Object.keys(readJson('./src/changelog.json'))[0];
```
surma: Should we rather read `package.json` for `VERSION`?

src/lib/offliner.ts
jakearchibald: Hmm, it will reload all tabs.

webpack.config.js
jakearchibald: Ohh good question. Do we want to force ourselves to write a changelog like this? I'll change it to use package json for this PR.

PR #253, opened 2018-11-09
Update file drop

jakearchibald:


jakearchibald:
12 bytes saved. Oh yes. 12 whole bytes.

PR #254, opened 2018-11-09
Ta-da. Back button.

jakearchibald:


src/components/Output/custom-els/TwoUp/index.ts
```diff
@@ -70,12 +70,13 @@ export default class TwoUp extends HTMLElement {
   connectedCallback() {
     this._childrenChange();
 
+    this._handle.innerHTML = `<div class="${styles.scrubber}">${
+      `<svg viewBox="0 0 27 20" fill="currentColor">${
+        '<path d="M17 19.2l9.5-9.6L16.9 0zM9.6 0L0 9.6l9.6 9.6z"/>'
+      }</svg>`
+    }</div>`;
```
jakearchibald: I moved this, because otherwise preact was getting rid of it. Not sure if there's a better way to handle that.

PR #269, opened 2018-11-13
ci: test Node.js 8, 10 and 11

DanielRuf:
We should test all current LTS branches to ensure that Squoosh work there and can be built using them.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,5 @@ node_js:
 - node
 - 10
 - 8
-- 6
 cache: npm
 script: npm run build || npm run build # scss ts definitions need to be generated before an actual build


.travis.yml
```diff
@@ -1,5 +1,8 @@
 language: node_js
 node_js:
 - node
+- 10
+- 8
+- 6
```
surma: I’d prefer we don’t support Node 6. This is not server-side code, so I don’t think we have to be that strict (and Node 6 is, like, really limited 😅)

DanielRuf: Should we make this clear then in the `engines` field of package.json?

surma: Sure! That’s probably a good idea :)

mathiasbynens: For a project like this — one that is not going to be a dependency in other projects — there should ideally only be a single officially-supported Node.js version, documented in the `.nvmrc` file. It doesn’t make sense to spend time supporting (and testing) multiple versions, IMHO.

DanielRuf: @mathiasbynens not everyone uses nvm. For example we use nodenv with `.node-version` so it would not work in all cases.

mathiasbynens: @DanielRuf Sure — every project can choose how they codify the explicitly supported Node.js version very project. I think it’s fine for a project like Squoosh to say “if you want to build this locally, follow these steps”.

`.nvmrc` seems to be the de facto standard for most Node.js projects, and Travis CI implicitly picks it up. I don’t think that’s true for `.node-version`.

surma: Happy to accept a .nvmrc in a separate PR.

@DanielRuf You mind removing Node 6, then we can go ahead with this one :)

DanielRuf: > @DanielRuf You mind removing Node 6, then we can go ahead with this one :)

Done.

PR #273, opened 2018-11-13
Allow text fields next to range inputs to be empty 

jakearchibald:


jakearchibald:
@kosamari reviewed this.

PR #275, opened 2018-11-13
Switch UglifyJS for Terser, Upgrade devDependencies, use `webpack-dev-server`

KraigWalker:
What moves faster these days, the code, or the tooling? 

* UglifyJS uses a package called `uglify-es` which is now no-longer supported. The React team have adopted TerserJS, which is a faster copy with a nearly identical API. Webpack will use it from v5.
* Upgrade `devDependencies`, including `typescript` and `babel` to the latest versions.
* Use `webpack-dev-server` instead of deprecated `webpack serve`
* Fix a TSLint warning `Warning: The 'no-boolean-literal-compare' rule requires type information.` by explicitly using tsconfig.json to specify what files will be checked (`tslint -p tsconfig.json`).
* Set to `strictNullChecks` to false. Avoids a compiler error where TypeScript complains that `document.activeElement` could be null, even though we immediately us it in an if statement to catch falsy value.
* Set `noImplicitAny` to false. Prevents compiler from complaining about key members of `encoderOptionsComponentMap` in the Options component.
* Allow `ga` to exist in the `window` object for analytics
* Fix .`_1iYR2> -> '._1iYR2>' is not a valid selector` warning. Saves a whopping **16 bytes** of critical css! Every little helps, right?


global.d.ts
```diff
@@ -6,7 +6,8 @@ declare interface NodeModule {
 }
 
 declare interface Window {
-  STATE: any
+  STATE: any;
+  ga: any;
```
surma: I have a more strict definition in [`src/missing-types.d.ts`](https://github.com/GoogleChromeLabs/squoosh/blob/master/src/missing-types.d.ts#L38)

package.json
```diff
@@ -4,75 +4,78 @@
   "version": "1.0.2",
   "license": "apache-2.0",
   "scripts": {
-    "start": "webpack serve --host 0.0.0.0 --hot",
+    "start": "webpack-dev-server --host 0.0.0.0 --hot",
```
surma: What is the difference to `webpack serve`? Why is it worth the switch (and the extra dependency)?

tsconfig.json
```diff
@@ -7,6 +7,8 @@
     "moduleResolution": "node",
     "experimentalDecorators": true,
     "noUnusedLocals": true,
+    "strictNullChecks": false,
+    "noImplicitAny": false,
```
surma: I disagree with both of these. I’d rather add some extra code to make sure the values are correct than deal with false negatives in production. If necessary, I’d rather resort to `// @ts-ignore` for a problematic line of code and add an explanatory comment.

jakearchibald:
fwiw I used TerserJS in https://github.com/GoogleChromeLabs/pinch-zoom and it worked well.

package.json
KraigWalker: `serve` has actually been moved out of webpack in the latest versions, and the old webpack-serve package is deprecated. It's now succeeded by `webpack-dev-server` [here](https://github.com/webpack/webpack-cli/blob/1a8099c6e762249d05d3ec41478f23ad08deee4c/packages/serve/index.ts)

package.json
```diff
@@ -4,75 +4,78 @@
...
+    "babel-loader": "^8.0.4",
     "babel-plugin-jsx-pragmatic": "^1.0.2",
     "babel-plugin-syntax-dynamic-import": "^6.18.0",
     "babel-plugin-transform-class-properties": "^6.24.1",
     "babel-plugin-transform-decorators-legacy": "^1.3.5",
     "babel-plugin-transform-object-rest-spread": "^6.26.0",
     "babel-plugin-transform-react-constant-elements": "^6.23.0",
     "babel-plugin-transform-react-jsx": "^6.24.1",
-    "babel-plugin-transform-react-remove-prop-types": "^0.4.19",
+    "babel-plugin-transform-react-remove-prop-types": "^0.4.20",
```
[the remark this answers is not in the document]
surma: I was wondering why Babel was in there! Yeah, let’s get rid of it. @KraigWalker, would you mind doing that?

package.json
surma: Sg! Then let’s keep this :)

package.json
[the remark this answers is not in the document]
KraigWalker: np :)

jakearchibald:
@KraigWalker thanks for looking into this btw! We built up a lot of tech debt here.

package.json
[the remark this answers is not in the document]
KraigWalker: Bye bye babel

global.d.ts
KraigWalker: I've moved that to global.d.ts now to use if that's alright.

tsconfig.json
KraigWalker: Good point. I've reverted tsconfig back, and weirdly, not getting any Implicit any warnings anymore when I build. Guess that was a bit of compiler panic...

jakearchibald:
I'm going to remove the ts-ignores in multipanel and do something less ignorey, then merge this.

jakearchibald:
Closed via https://github.com/GoogleChromeLabs/squoosh/pull/289.

@KraigWalker huge thanks for this! Really appreciated.

PR #289, opened 2018-11-14
Building on #275

jakearchibald:


jakearchibald:
Netlify is happy so I'm happy.

This doesn't need a version bump because there's no behaviour changes.

Issue #274, opened 2018-11-13
Accidental click & drag on home screen loads logo image.


GifCo:
If you accidentally click and drag the logo on the home screen it will load the logo into the image comparison module.

Not sure if this is desired behavior? If not should have an `e.preventDefault()` on the logo img tag.

surma:
That’s definitely not intended, but I can‘t click-n-drag the logo. Which browser and OS are you on?

GifCo:
@surma Im on Win10 and I can recreate the behavior on Edge 42.17134.1.0
I could have swore it was doing the same in Chrome but I literally just updated to newest version (70.0.3538) 20min ago and I can no longer see the behavior.

JamieFarrelly:
Same here on Windows 10 and Edge. Even adding "pointer-events: none" to the image seems to fix it.

jakearchibald:
This sounds like a fun issue to fix. Thanks for reporting!

JamieFarrelly:
Sorry for the multiple pull requests but I've opened a pull request to fix this #290. The sample images on the homepage had the same issue.

kosamari:
closing the issue since PR was merged.

PR #290, opened 2018-11-14
Preventing images from being dragged in Edge

JamieFarrelly:
This is to fix the [Accidental click & drag on home screen loads logo image](https://github.com/GoogleChromeLabs/squoosh/issues/274) issue. Microsoft Edge doesn't support the second block of CSS in [this file](https://github.com/GoogleChromeLabs/squoosh/blob/master/src/style/reset.scss) to prevent dragging. 

While testing this I also noticed it's not just the logo image that can be dragged in Edge on the homepage, the sample images under the "Or try one of these:" section also had the same issue, although it's not as big of an issue. At first I thought this was intentional but then I noticed on other browsers those can't be dragged either so to make it consistent I prevented dragging on those too.

### DIFF
--- a/src/components/intro/style.scss
+++ b/src/components/intro/style.scss
@@ -48,6 +48,7 @@
 
 .logo {
   composes: abs-fill from '../../lib/util.scss';
+  pointer-events: none;
 }
 
 .open-image-guide {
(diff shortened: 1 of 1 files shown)


JamieFarrelly:
@surma Yep, still works! pointer-events works in all versions of Edge too 👍

jakearchibald:
Thank you @JamieFarrelly!

Issue #288, opened 2018-11-14
Browser back button should go from compression screen back to home screen


jakearchibald:
Uh oh it's `pushState` time.

fstanis:
Does it also make sense to serialize the compression options in the URL somehow (e.g. `https://squoosh.app/compress?left_codec=original&right_codec=MozJPEG&right_quality=40&right_etc=...`? This could solve the aforementioned issue of losing them when going back.

Not sure what to do with the loaded image though. Some ideas:

- Save last used to cache, load from cache
- Load second view without image and ask the user to drag and drop it [again]
- Add third view, asking the user to drag and drop image to proceed

PR #309, opened 2018-11-15
Add basic history handling (#288)

maciejmatu:
I added super simple window location handling with browser history api. It's hidden behind a facade in `utils/history` so that a poly-fill could be easily added if needed.

It covers following cases:
- [x] When entering the page (initial `<App />` mounting), if the location is other that `'/'`, and the `file` is not preset (which for now is always the case), we replace the location back to `'/'`. 
- [x] When user selects a file, we push `'/editor'` to history.
- [x] When user goes back manually, we render the initial view and replace state to `'/'`
- [x] When user goes back using browser built-in mechanism, we to the same thing as above. 

I feel like configurator state persistence should be done separately from this simple fix, as it will be a bit more complex 😄


kosamari:
Thanks for the PR - just wanted to let you know, the team is taking time off after Chrome Dev Summit so we'll probably get to this in December. Cheers!

jakearchibald:
Thanks for the PR!

This is adding to the initial interaction bundle, so we should look to see if we can avoid that.

surma:
You can put it into `Compress`. History doesn’t make sense to load in the initial bundle as there’s no history to do. Once we load `Compress`, we can be sure that there’s at least one history state.

maciejmatu:
That's a good idea 👍 This way we also have less code. The only downside is that now if somebody refreshes the page while on `/editor` route, they'll see the `<Intro />` component but the history won't update (since it is managed inside `<Compress />` that hasn't mounted yet).

If I wanted to re-set the initial state I could do it in `<Intro />` component, but it's in the initial bundle. Other way would be to do it in `Compress` module outside of react component, but it would happen async and feels a bit hacky 😄

src/lib/history.ts
```diff
@@ -0,0 +1,11 @@
+export default {
+  on: window.addEventListener.bind(window),
```
surma: Nit: Not a fan on the `on()` syntax. That’s a node idiom and something that we have avoided in this code base. I’d prefer either an explicit `addEventListener` or maybe even just a `addPopStateListener`.

src/lib/history.ts
```diff
@@ -0,0 +1,11 @@
...
+  off: window.removeEventListener.bind(window),
+  pathname: window.location.pathname,
```
surma: This is not used, can we remove it?

src/components/compress/index.tsx
```diff
@@ -232,6 +236,16 @@ export default class Compress extends Component<Props, State> {
     }
   }
 
+  componentDidMount() {
+    history.push(ROUTE_EDITOR);
+    history.on('popstate', this.props.onBack);
+  }
+
+  componentWillUnmount() {
+    history.replace(ROUTE_INDEX);
```
surma: I’m a bit confused by this one. `componentWillUnmount()` will happen when the user pressed back, so the route should already be correct, no?

maciejmatu: Correct, but the code in `componentWillUnmount()` covers a case, where the user presses a non-browser back button, the one render on top-left corner and triggers `onBack` function. If we didn't have the replace here then the history wouldn't update after clicking on it.

src/lib/history.ts
maciejmatu: sure thing 👍

src/lib/history.ts
maciejmatu: `addPopStateListener` sounds good, I'll use that.

src/components/compress/index.tsx
surma: Ah, you raise a good point. I think we shouldn’t be using `replaceState()` here, tho. Because it’s a back button, we should probably call `history.back()` when that button gets clicked.

So I’d vote for rejigging the button, remove the `replace()` here (and, by conclusion, remove `replace()` from the history module.

maciejmatu: Good idea. You can take a look now, I pushed an update where I trigger `history.back` after pressing the button  instead of using it in `componentWillUnmount()`.

jakearchibald:
Also, pressing "forward" in the browser doesn't work. Do we want it to?

maciejmatu:
I think the forward button would be a good idea if we were to somehow cache the uploaded file, or at least not remove it from state and keep a reference in memory. This way we could restore it after going forward. At least if the user didn't refresh the page. 

Alternatively I could try expanding this ticket and do file-persistence in IndexedDB, and filter config serialisation in the url. This way we could cover actually restoring user state even on refresh.

As for the fallback I can add a Netlify redirect to `/` for just `/editor` route, or for all routes. I personally prefer using regular history over hashes, but I leave the choice to you. I can implement both solutions 😄

jakearchibald:
@maciejmatu really appreciate your work on this. Sorry for the back and forth. Let's decide on the interaction we want before we waste any more of your finger skin 😀

How about, as it is now:

* Opening an image pops to `/editor`.
* Pressing back goes to home screen.

But also:

* Navigating to `/editor` redirects to `/` (we'll need to set up the same redirect in the service worker).
* Pressing forward to `/editor` opens the last image, but doesn't need to remember any of the settings, unless there's a way to get that 'for free'.

To do this, I guess the top level needs to control things? We could lazy-import it like we do with the offline stuff.

@surma what do you think?

@maciejmatu again, apologies for the back & forth. Totally understand if you tell us to go do it ourselves 😄.

jakearchibald:
Oh, any data we persist should be in memory, or something tied to the tab like the history API. If we start putting the user's image data into IDB it complicates our privacy story.

maciejmatu:
@jakearchibald no worries, I quite enjoy the learning experience 😄. I'll try implementing the missing features you specified in the evening

surma:
Conceptually this looks good to me. @maciejmatu Thanks so much for staying on this! Very much appreciated :)

maciejmatu:
Okay so I did the basic functionality without moving the code to an async loaded module yet. Going forward should work fine. I also added netlify redirect for all routes to `/`. I wanted to use a `301` status code but it only worked if I used `/index.html` instead of `/`, as if the custom redirect overwrites the netlify default one, so for now I left it with `200`. 

I still need to add the one in serviceWorker. Would something like this be sufficient?

```js
if (event.request.headers.get('accept').includes('text/html')) {
  event.respondWith(caches.open(versionedCache).then((cache) => {
    return cache.match(new Request('/'));
  });
}
```

maciejmatu:
I also played around with moving the router module to an async loaded module. I'm not sure if I went in the right direction but here is how it *might* look: https://github.com/maciejmatu/squoosh/compare/fix-history...fix-history-async 
The difference from the synchronous solution is that after uploading file, if the routing module didn't finish loading yet, we'll see a loader, and when it finishes, the awaited `pushState` will occur, and we'll see the editor then.

src/components/App/index.tsx
```diff
@@ -65,15 +70,14 @@ export default class App extends Component<Props, State> {
   }
 
   @bind
-  private onFileDrop(event: FileDropEvent) {
-    const { file } = event;
+  private onFileDrop({ file }: FileDropEvent) {
     if (!file) return;
-    this.setState({ file });
+    this.setState({ file }, this.openEditor);
```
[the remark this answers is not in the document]
maciejmatu: This is just a callback function to ensure that `this.openEditor` is called after the app has rendered with a new state.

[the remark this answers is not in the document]
surma: @jakearchibald @developit WDYT? I find it a bit odd to scatter the side-effects. I’d prefer having this in `render()`.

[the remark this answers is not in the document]
maciejmatu: How would we trigger it from render on both `onFileDrop`, and `onIntroPickFile`? 
Actually, when I look at it, `onFileDrop` and `onIntroPickFile` are nearly identical, so I could remove the `openEditor` function at all, move it's logic to `onIntroPickFile`, and inside `onFileDrop`, just call `onIntroPickFile` if the file was selected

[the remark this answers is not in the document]
maciejmatu:
```jsx
@bind
private onFileDrop({ file }: FileDropEvent) {
  file && this.onIntroPickFile(file);
}

@bind
private onIntroPickFile(file: File | Fileish) {
  this.setState({ file, isEditorOpen: true });
  history.push(ROUTE_EDITOR);
}
```

src/components/App/index.tsx
```diff
@@ -83,18 +87,34 @@ export default class App extends Component<Props, State> {
...
+    history.removePopStateListener(this.onPopState);
   }
 
-  render({}: Props, { file, Compress }: State) {
+  render({}: Props, { file, isEditorOpen, Compress }: State) {
     return (
       <div id="app" class={style.app}>
         <file-drop accept="image/*" onfiledrop={this.onFileDrop} class={style.drop}>
-          {(!file)
+          {(!file || !isEditorOpen)
```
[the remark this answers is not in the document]
maciejmatu: I can remove but it, but then typescript throws an error that the `file` could be undefined while passing it to `Compress`. I could cast the type and do `<Compress file={file as File as Fileish} ... />` unless there is a better/cleaner way to do it.

src/components/App/index.tsx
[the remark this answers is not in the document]
maciejmatu: If we went with it I could also rename `onIntroPickFile` to e.g. `onFileSelected`, to not be tied to `<Intro />`

_redirects.ejs
```diff
@@ -0,0 +1 @@
+/* / 200
```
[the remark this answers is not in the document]
maciejmatu: 307 redirect didn't rewrite the url. I made it work with: 
```
/index.html / 301
/* /index.html 301
```
For some reason this works, but `/* / 301` does not.

_redirects.ejs
```diff
@@ -0,0 +1,2 @@
+/index.html / 301
+/* /index.html 301
```
jakearchibald: If this line worked, wouldn't all our image, script, & CSS requests fail?

src/components/App/index.tsx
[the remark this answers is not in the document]
jakearchibald: > This is just a callback function to ensure that this.openEditor is called after the app has rendered with a new state.

What would happen if we didn't use the callback, and just called `this.openEditor`?

src/components/App/index.tsx
```diff
@@ -83,18 +87,32 @@ export default class App extends Component<Props, State> {
...
+    history.push(ROUTE_EDITOR);
+    this.setState({ isEditorOpen: true });
+  }
+
+  componentDidMount() {
+    history.addPopStateListener(this.onPopState);
+  }
+
+  componentWillUnmount() {
+    history.removePopStateListener(this.onPopState);
```
jakearchibald: When does the main app unmount?

_redirects.ejs
maciejmatu: If netlify's server finds a static file it will return it anyway. If we wanted to override all files we would need to pass an exclamation mark e.g. `300!`. You can test it on deploy preview

src/components/App/index.tsx
maciejmatu: Right now, only on full page refresh so this might not be needed 😄

_redirects.ejs
jakearchibald: TIL, thanks!

src/components/App/index.tsx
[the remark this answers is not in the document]
surma: You can do `file={file!}`

surma:
@maciejmatu Once you get around to address the last bits of feedback, please also rebase against master :)

src/components/App/index.tsx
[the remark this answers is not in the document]
maciejmatu: Ah right, thanks

src/components/App/index.tsx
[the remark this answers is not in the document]
maciejmatu: I'm not 100% sure how Preact handles rendering, but it looks like after calling `this.openEditor` not as a callback, everything works fine, so it probably batches both state updates into a single render.

src/components/App/index.tsx
```diff
@@ -74,18 +80,28 @@ export default class App extends Component<Props, State> {
...
+  }
+
+  @bind
+  private openEditor() {
+    history.push(ROUTE_EDITOR);
+    this.setState({ isEditorOpen: true });
+  }
+
+  componentDidMount() {
+    history.addPopStateListener(this.onPopState);
```
jakearchibald: This could just be in the constructor, right?

src/components/App/index.tsx
```diff
@@ -74,18 +80,28 @@ export default class App extends Component<Props, State> {
...
   @bind
-  private onBack() {
-    this.setState({ file: undefined });
+  private onPopState() {
+    this.setState({ isEditorOpen: history.pathname === ROUTE_EDITOR });
+  }
+
+  @bind
+  private openEditor() {
+    history.push(ROUTE_EDITOR);
```
jakearchibald: Maybe there's a better way, but I guess this could be done in one of the state-change callbacks, as in if `state.file` goes from undefined to defined, then push state. That would avoid pushing state when the file is just changed.

maciejmatu: I added a check, if `isEditorOpen` is true then we don't do anything, since the file will change by itself, and we don't need to update the history

src/components/App/index.tsx
maciejmatu: done 👍

jakearchibald:
Thanks for persevering with this!

jakearchibald:
The latest changes appear to have broken opening files. I'd like to get this landed so I'm going to take a swing at the final changes.

maciejmatu:
@jakearchibald sorry for not testing it thoroughly. I quickly checked it out and it seems like I messed up the early return condition 🤦‍♂️ Not sure if it fixed everything though

jakearchibald:
Added a commit in https://github.com/GoogleChromeLabs/squoosh/pull/370

jakearchibald:
@maciejmatu apparently I could have just pushed to this PR. TIL.

jakearchibald:
@maciejmatu thank you for all your work on this!

jakearchibald:
https://github.com/GoogleChromeLabs/squoosh/pull/371 is the PR for pushing this (and another bug fix) to live. Once that merges, it'll be live in a couple of minutes.

PR #322, opened 2018-11-19
Rotate

jakearchibald:
Ready!


src/codecs/resize/options.tsx
```diff
@@ -135,7 +135,7 @@ export default class ResizerOptions extends Component<Props, State> {
                 onChange={this.onChange}
               >
                 <option value="stretch">Stretch</option>
-                <option value="cover">Cover</option>
+                <option value="contain">Contain</option>
```
jakearchibald: I have no idea why this was called "cover" before. It's clearly "contain", and other bits of code referred to it as "contain".

src/codecs/rotate-flip/options.tsx
```diff
@@ -0,0 +1,70 @@
...
+          <Range
+            name="rotate"
+            min="0"
+            max="270"
+            step="90"
+            value={'' + options.rotate}
+            onInput={this.onChange}
+          >
+            Rotate:
+          </Range>
```
jakearchibald: I'm not sure if this is the best input to use. Alternatively we could have a number input with buttons either side that increment/decrement by 90.

src/codecs/rotate-flip/processor.ts
```diff
@@ -0,0 +1,100 @@
...
+      d2Start = 0;
+    }
+  }
+
+  for (let d2 = d2Start; d2 >= 0 && d2 < d2Limit; d2 += d2Advance) {
+    for (let d1 = d1Start; d1 >= 0 && d1 < d1Limit; d1 += d1Advance) {
+      yield ((d1 * d1Multiplier) + (d2 * d2Multiplier)) * bpp;
+    }
+  }
+}
```
jakearchibald: I was happy when I wrote this, but now it looks like an unmaintainable mess. I'm going to give offscreen canvas a try, and for other browsers… main thread canvas? It'd avoid forking the logic.

src/components/Options/index.tsx
```diff
@@ -91,29 +93,56 @@ export default class Options extends Component<Props, State> {
...
-    );
+    let newState = this.props.preprocessorState;
+
+    // If orientation has changed, we should flip the resize values.
+    if (preprocessor === 'rotateFlip' && this.props.preprocessorState.rotateFlip.rotate % 180) {
+      newState = cleanMerge(newState, 'resize', {
+        width: this.props.preprocessorState.resize.height,
+        height: this.props.preprocessorState.resize.width,
+      });
+    }
```
jakearchibald: Oh shit this is wrong. brb

jakearchibald:
Hmm, stand by, there are bugs that I didn't notice in dev.

jakearchibald:
Oh great. The bugs don't appear in dev.

jakearchibald:
@developit one for you https://github.com/GoogleChromeLabs/critters/issues/17

jakearchibald:
I've hacked around the issue for now, but we shouldn't merge with the hack.

jakearchibald:
Yeah, `OffscreenCanvas` seems worth it. Rotating the red panda image goes from 900ms to 300ms.

…no wait, lol, the offscreen canvas causes Chrome to switch OSX to the discrete GPU. That means *serious jank* when the worker loads the offscreen canvas code, and another jank when the worker is terminated and it switches back to the integrated GPU. Ok, I'll create a test case and file.

jakearchibald:
https://bugs.chromium.org/p/chromium/issues/detail?id=906619

jakearchibald:
Given that `OffscreenCanvas` is off the table, I should write/find a C++ solution and WASM it.

jakearchibald:
Maybe it's time to bring in https://github.com/ImageMagick/ImageMagick. Once we have that it'll be easy to add loads of different transforms.

surma:
We could implement 90° image rotation in JavaScript using `ImageData` and create a new RGBA buffer. Probably much smaller and sufficiently fast.

ImageMagick seems quite heavy for this but I’m not sure what else you had in mind. More scaling algorithms?

jakearchibald:
> We could implement 90° image rotation in JavaScript using ImageData and create a new RGBA buffer. Probably much smaller and sufficiently fast.

That's what this PR is doing currently.

> ImageMagick seems quite heavy for this but I’m not sure what else you had in mind. More scaling algorithms?

Yeah, the resize algorithms I'd like to add are all in ImageMagick. But maybe I'm overcomplicating things, and we could just land my JS solution for now.

jakearchibald:
Hah, I've just brought my implementation down from 900ms to 300ms, same as the offscreen canvas. It could probably be faster with multithreading too.

I was using a generator in the previous implementation. Removed it.

surma:
screen shot 2018-11-20 at 1 04 10 pm

Also: Should the rotate/flip transform be special and get applied to original image as well?

jakearchibald:
@surma `DOMMatrix` isn't in Safari, and the CSS matrix fallback isn't available in a worker.

I started looking at bringing in gl-matrix for this. WDYT?

surma:
Of course it isn’t. Eh, `gl-matrix` is probably more code than what you’ve written at this point, so let’s stay on your code and we can look at ImageMagick eventually.

jakearchibald:
I'll give the polyfill a go. It would make the code a whole lot more readable.

jakearchibald:
This PR breaks the pinch zoom's boundary checking, since the element is centered using transforms. I should go back to the previous method.

Also, the Chrome graphics glitch is back, probably because I introduced a wrapping element but didn't move the fix into that element.

Haprog:
> Also: Should the rotate/flip transform be special and get applied to original image as well?

I think yes. Otherwise the split view is just plain weird and not that useful anymore.

One idea: Would it be simpler to apply the real image data rotation only when saving/exporting, and rotate the whole split view area (including both images) via a CSS transform only (just to get the preview showing in correct orientation, and then the original doesn't need to be rotated in a different way)?

Or would it make sense to have an option to preserve the Exif Orientation data in the output image (even if stripping all other Exif data) instead of doing a rotation on the image data?

jakearchibald:
I gave `DOMMatrix` a spin:

```ts
export function rotateFlip(data: ImageData, opts: RotateFlipOptions): ImageData {
  const flipDimensions = opts.rotate % 180 !== 0;
  const width = flipDimensions ? data.height : data.width;
  const height = flipDimensions ? data.width : data.height;
  const out = new ImageData(width, height);
  const { width: inputWidth, height: inputHeight } = data;

  const matrix = new DOMMatrix();
  matrix.translateSelf(width / 2, height / 2);
  matrix.scaleSelf(
    opts.flipHorizontal ? -1 : 1,
    opts.flipVertical ? -1 : 1,
    1,
  );
  matrix.rotateSelf(opts.rotate);
  matrix.translateSelf(-inputWidth / 2, -inputHeight / 2);

  const point = new DOMPoint();

  for (let y = 0; y < inputHeight; y += 1) {
    for (let x = 0; x < inputWidth; x += 1) {
      point.x = x;
      point.y = y;
      const destPoint = matrix.transformPoint(point);
      const fromStart = (y * inputWidth + x) * bpp;
      const destStart = (Math.round(destPoint.y) * width + Math.round(destPoint.x)) * bpp;
      for (let i = 0; i < bpp; i += 1) {
        out.data[destStart + i] = data.data[fromStart + i];
      }
    }
  }
  return out;
}
```

Much simpler code, but rotating the red panda image now takes **37.5 seconds**, compared to 280ms. I'm guessing the creation of point objects and the necessary rounding makes the difference. Using gl-matrix will remove the object creation, but we'd still have the rounding to deal with.

Going to stick with the currently pushed solution.

jakearchibald:
I now agree that having a flip/rotate on one side only isn't particularly useful.

It feels like we need a concept like pre-preprocessors, that operate on the original image data. I guess the UI could appear as options of "original image".

Thoughts?

surma:
Agree on the data flow. As for UI: How about we acknowledge how special rotation is and put a rotate button next to the “Toggle Background” button (potentially removing the text from the background button to free up some screen estate).

surma:
(That sounded very passive aggressive. Apologies. Legit a suggestion, nothing more.)

jakearchibald:
My only worry about putting more UI there is it doesn't scale, but maybe this is the only input processor we'll ever have…?

surma:
It is for now. If we find something new we can revisit UI placement, I'd say.

jakearchibald:
Agreed. Ok, I'll give it a spin. I confidently predict I can land this feature before the heat death of the universe.

src/components/App/index.tsx
```diff
@@ -18,12 +18,6 @@ const offlinerPromise = import(
   '../../lib/offliner',
 );
 
-export interface SourceImage {
-  file: File | Fileish;
-  data: ImageData;
-  vectorImage?: HTMLImageElement;
-}
-
```
jakearchibald: I have no idea what this was doing here. Probably left over from when we split the compressor out into its own module.

src/components/Options/index.tsx
```diff
@@ -91,24 +91,25 @@ export default class Options extends Component<Props, State> {
...
     const el = event.currentTarget as HTMLInputElement;
     const preprocessor = el.name.split('.')[0] as keyof PreprocessorState;
 
-    this.props.onPreprocessorOptionsChange(
-      cleanSet(this.props.preprocessorState, `${preprocessor}.enabled`, el.checked),
-    );
+    let newState = this.props.preprocessorState;
+
+    newState = cleanSet(newState, `${preprocessor}.enabled`, el.checked);
+    this.props.onPreprocessorOptionsChange(newState);
```
jakearchibald: I'll revert this change.

src/components/Output/index.tsx
```diff
@@ -226,30 +264,32 @@ export default class Output extends Component<Props, State> {
...
+                ref={linkRef(this, 'canvasLeft')}
+                width={leftDraw && leftDraw.width}
+                height={leftDraw && leftDraw.height}
+                style={{
+                  width: originalImage && originalImage.width,
+                  height: originalImage && originalImage.height,
+                  objectFit: leftImgContain ? 'contain' : '',
+                }}
+              />
+            </div>
```
jakearchibald: I'm going to try and remove the need for this boxing.

src/components/Output/style.scss
```diff
@@ -31,6 +31,20 @@
...
 }
 
+.pinch-target {
+  display: flex;
+  flex-shrink: 0;
+  justify-content: center;
+  align-items: center;
+  // This is a hack to work around https://github.com/GoogleChromeLabs/critters/issues/17.
+  // DO NOT LET THIS MERGE. I'm just putting it here so the UI can be reviewed in the PR.
+  contain: layout !important;
```
jakearchibald: Might not need this hack anymore. Will check.

src/components/compress/result-cache.ts
```diff
@@ -15,7 +14,7 @@ interface CacheResult {
 interface CacheEntry extends CacheResult {
   preprocessorState: PreprocessorState;
   encoderState: EncoderState;
-  source: SourceImage;
+  sourceData: ImageData;
```
jakearchibald: The source image now has data that doesn't matter to the result cache (eg, rotation settings).

surma:
UI choice looks really good for me!

src/lib/icons.tsx
```diff
@@ -54,6 +66,23 @@ export const BackIcon = (props: JSX.HTMLAttributes) => (
...
+  direction: 'horizontal' | 'vertical';
+}
+
+export const FlipIcon = (props: FlipIconProps) => {
+  const { direction, ...otherProps } = props;
+
+  return (
+    <Icon {...otherProps}>
+      <path
+        transform={`rotate(${ direction === 'horizontal' ? 90 : 0 })`}
```
jakearchibald: This transform is almost certainly wrong, but I'll catch it when I add the buttons properly.

jakearchibald:
@surma @kosamari @developit how important do we think 'flip' is? It's easy to add, but I'm worried about usefulness vs the UI clutter.

surma:
Fairly low, imo. IIUC, the main demand is coming from importing pictures from your phone, where native pixel order is almost always landscape, even if the picture was taken portrait.

jakearchibald:
I've removed flip. We can always bring it back later.

jakearchibald:
I think this is ready. @surma mind taking another look?

src/codecs/processor-worker/index.ts
```diff
@@ -14,16 +10,29 @@ async function mozjpegEncode(
...
+  data: ImageData, opts: import('../imagequant/processor-meta').QuantizeOptions,
+): Promise<ImageData> {
   const { process } = await import(
     /* webpackChunkName: "process-imagequant" */
     '../imagequant/processor',
   );
   return process(data, opts);
 }
 
+async function rotateFlip(
```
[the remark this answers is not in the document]
jakearchibald: Hah well spotted. I thought I caught all of these. Should have done a search.

src/codecs/rotate/processor-meta.ts
```diff
@@ -0,0 +1,3 @@
+export type OptionType = 0 | 90 | 180 | 270;
```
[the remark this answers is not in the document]
jakearchibald: As in, an interface with a single property? `{ rotate }`?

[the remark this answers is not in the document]
jakearchibald: So calls would be:

```js
rotate(data, { rotate: 90 });
```

src/codecs/rotate/processor.ts
```diff
@@ -0,0 +1,75 @@
...
+
+export function rotate(data: ImageData, rotate: 0 | 90 | 180 | 270): ImageData {
+  // Early exit if there's no transform.
+  if (rotate === 0) return data;
+
+  const flipDimensions = rotate % 180 !== 0;
+  const width = flipDimensions ? data.height : data.width;
+  const height = flipDimensions ? data.width : data.height;
+  const out = new ImageData(width, height);
+  const { width: inputWidth, height: inputHeight } = data;
```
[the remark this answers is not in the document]
jakearchibald: Good shout

Issue #244, opened 2018-11-05
Display file name somewhere


jakearchibald:
Some feedback:

> it would be really nice to have the file name somewhere. I was working through a handful of files and after the 4th or 5th, I couldn't remember which one I was on - and essentially had to guess. 

Seems like a good idea. A quick win would be to change the page title, but we could do something in the main UI too. Maybe think about this along with https://github.com/GoogleChromeLabs/squoosh/issues/243.

pewstiepoll:
Hello!
Please feel free to check out my first step to implement the solution #326 
Awaiting for your feedback!

kosamari:
Note: follow up with @argyleink about better way to display image title outside of tab title.

jakearchibald:
The tab title might be enough

argyleink:
tab title is great, hard to tell if it's enough or not unless we did some user testing..

I like @kosamari's type display idea [here](https://github.com/GoogleChromeLabs/squoosh/issues/101#issuecomment-872391602). This issue (and #101) could adopt her idea by adding the file name and type, like this:

![Screen Shot 2021-07-09 at 10 14 13 AM](https://user-images.githubusercontent.com/1134620/125114171-7528e100-e09e-11eb-95c9-cfac19678bee.png)

we kinda get 2 birds with 1 PR: reinforce the file name again, inline with the UI (not just page title) and also display the original type. thoughts?

whoisvadym:
@argyleink I really like the idea of reinforcing the file name + extension in the UI.
Would be glad to work on the implementation if other maintainers find it useful.

harsh26shah03:
#1354 #1352

harsh26shah03:
#1354

PR #326, opened 2018-11-19
Display uploaded file name in the document title (#244)

pewstiepoll:
Hello!
This is the first time I contribute to open-source.
First I decided to divide an issue into smaller pieces:
1. Display filename in the title
2. Display filename in UI

I've already implemented the first part and would really like to hear feedback from you!


jakearchibald:
@pewstiepoll honoured that this project is your first OS contribution.

I'm pretty happy with using the document title for this. We can figure out where else to display it later. What do you think @developit @kosamari @surma?

src/components/App/index.tsx
```diff
@@ -60,8 +62,13 @@ export default class App extends Component<Props, State> {
       this.componentDidUpdate = (props, state) => {
         if (oldCDU) oldCDU.call(this, props, state);
         window.STATE = this.state;
+
+        // change title once component updated
+        this.changeTitle();
```
jakearchibald: This is inside `if (process.env.NODE_ENV === 'development') {` so it won't work in production.

src/components/App/index.tsx
```diff
@@ -87,6 +94,19 @@ export default class App extends Component<Props, State> {
...
   }
 
+  @bind
+  private changeTitle() {
+    // Add the filename to the document title
+    const { file } = this.state;
+
+    // If there's a file - update title
+    if (file && file.name) {
+      document.title = `${this.defaultTitle} • ${file.name}`;
```
jakearchibald: I think we can put the file name first. Eg, https://www.bbc.co.uk/news/uk-politics-46269757 - the headline is in the title before "BBC news".

pewstiepoll:
@jakearchibald thanks for quick reply!
About moving code to `compress/index.tsx` : in this case it will make Compress component perform the side effect (e.g. document changes) and from that perspective it would make it harder to debug in future. But from your presentation on Google Dev Summit I can assume that squoosh is about "first interaction" therefore moving things like this from init bundle makes sense. Just wanted to know your opinion about such trade-off.

About other comments: I agree in both cases and will fix it in my next changes.

jakearchibald:
> in this case it will make Compress component perform the side effect

I see your point, but given it's *almost* the top level component, I'm fine with it. It also keeps the bulk of the "handle new image" code in one place.

pewstiepoll:
@jakearchibald hi!
Just updated my changes corresponding to your review.
Also trying to figure out where to put filename in UI. Any suggestions?

surma:
Why don't we scope this PR to document title and do a UI tweak separately? Wdyt?

pewstiepoll:
@surma I think it makes sense. Just wonder if it's ok to have such a tiny changes in PR.

surma:
No change is too small for its own PR. Makes it much easier to review and to reason about :)

src/components/compress/index.tsx
```diff
@@ -162,6 +162,8 @@ const resultTitles = ['Top', 'Bottom'];
 const buttonPositions =
   ['download-left', 'download-right'] as ('download-left' | 'download-right')[];
 
+const originDocumentTitle = document.title;
```
jakearchibald: This this a typo? As in, should it be `originalDocumentTitle`?

src/components/compress/index.tsx
```diff
@@ -226,12 +228,23 @@ export default class Compress extends Component<Props, State> {
     });
   }
 
+  private updateDocumentTitle(filename : string = '') {
```
jakearchibald: No space before the `:`, so `filename: string = ''`. Sorry, I thought our linter would catch this.

jakearchibald: Also, when is filename empty?

src/components/compress/index.tsx
```diff
@@ -226,12 +228,23 @@ export default class Compress extends Component<Props, State> {
...
+    const titleWithFilename = `${filename} - ${originDocumentTitle}`;
+    const newTitle : string = filename ? titleWithFilename : originDocumentTitle;
```
jakearchibald: It seems odd to create `titleWithFilename` and then discard it. So:

```ts
document.title = filename ? `${filename} - ${originDocumentTitle}` : originDocumentTitle;
```

src/components/compress/index.tsx
jakearchibald: Ohh I see, it's on unmount. Gotcha.

src/components/compress/index.tsx
```diff
@@ -328,7 +341,9 @@ export default class Compress extends Component<Props, State> {
         });
       }
 
-      this.setState(newState);
+      this.setState(newState, () => {
+        this.updateDocumentTitle(file.name);
```
jakearchibald: I don't think we need to use the callback here, right? We can just call `updateDocumentTitle`.

pewstiepoll: I thought we might want to display filename only after the file is loaded into the state, but if there's no reason - I'll change it.

pewstiepoll:
@jakearchibald done.
Can't wait to see my changes in prod ^_^

src/components/compress/index.tsx
jakearchibald: You'll always see the result of `setState` before the next paint (unless I'm missing something), so without the callback is fine IMO.

jakearchibald:
@pewstiepoll I'm really happy with this, thanks for making it happen & putting up with the review process.

pewstiepoll:
@jakearchibald I’m so proud to make my first steps into OS under your guidance! Thank you! :)

jakearchibald:
Your changes are now live 🎉

pewstiepoll:
@jakearchbald, hooray!
Should I delete the branch now?

jakearchibald:
You're welcome to (it's your branch). We don't need it anymore.

PR #328, opened 2018-11-20
add Build Status

jf17:


jakearchibald:
@surma what do you think? I haven't found these things particularly useful in the past, especially now we never merge if Travis fails.

I have seen these images fail to load though, and they make projects look broken when they aren't.

surma:
No strong opinions on my end. I agree with your point that it might end up making Squoosh look broken when it’s not.

It did get me thinking if we wanna give a shout-out to the services/libraries that we rely on, though.

PR #337, opened 2018-11-25
[BUILD] Manually running typed css module loader

abhijit-paul:
Running typed css module loader ahead of the build ensures all the typed styles are present before running the build; thereby ensuring the success of first build itself.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,8 @@
   "license": "apache-2.0",
   "scripts": {
     "start": "webpack-dev-server --host 0.0.0.0 --hot",
-    "build": "webpack -p",
+    "typecss": "find src -name '*css' -exec tcm -c -p {} \\;",
+    "build": "npm run typecss; webpack -p",
     "lint": "tslint -c tslint.json  -p tsconfig.json -t verbose 'src/**/*.{ts,tsx,js,jsx}'",
     "lintfix": "tslint -c tslint.json -p tsconfig.json -t verbose --fix 'src/**/*.{ts,tsx,js,jsx}'"
   },
@@ -57,6 +58,7 @@
     "tslint-config-airbnb": "^5.11.0",
     "tslint-config-semistandard": "^7.0.0",
     "tslint-react": "^3.6.0",
+    "typed-css-modules-loader": "0.0.17",
     "typescript": "^3.1.6",
     "typings-for-css-modules-loader": "^1.7.0",
     "url-loader": "^1.1.2",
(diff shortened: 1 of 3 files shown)


src/components/Output/style.scss
```diff
@@ -137,7 +137,7 @@
   flex-shrink: 0;
   // This fixes a severe painting bug in Chrome.
   // We should try to remove this once the issue is fixed.
-  // https://bugs.chromium.org/p/chromium/issues/detail?id=870222#c10
+  // https://bugs.chromium.org/p/chromium/issues/detail?id=870222#c10;
```
jakearchibald: What is this change for?

package.json
```diff
@@ -57,6 +58,7 @@
     "tslint-config-airbnb": "^5.11.0",
     "tslint-config-semistandard": "^7.0.0",
     "tslint-react": "^3.6.0",
+    "typed-css-modules-loader": "0.0.17",
```
jakearchibald: Isn't this also a webpack loader? Doesn't that mean we have two now?

package.json
```diff
@@ -5,7 +5,8 @@
   "license": "apache-2.0",
   "scripts": {
     "start": "webpack-dev-server --host 0.0.0.0 --hot",
-    "build": "webpack -p",
+    "typecss": "find src -name '*css' -exec tcm -c -p {} \\;",
```
jakearchibald: Is `find` necessary here? https://github.com/Quramy/typed-css-modules suggests it isn't needed.

src/components/Output/style.scss
abhijit-paul: This semicolon ensures the the typed css module loader doesn't error out on the css to typescript conversion. Otherwise, it complains for lack of delimiter at the end.

package.json
abhijit-paul: This brings the command line tool for typed css module loader.

package.json
abhijit-paul: _tcm -c -p src/**/*css_ doesn't work
With Regex
Single file

src/components/Output/style.scss
jakearchibald: Whoa, that's a pretty weird bug. We'd have to at least document this, as otherwise it looks like a mistake.

package.json
jakearchibald: Isn't https://github.com/Quramy/typed-css-modules the main module for the cli tool?

jakearchibald:
Closing in favour of https://github.com/GoogleChromeLabs/squoosh/pull/350. 

@abhijit-paul thanks for taking a swing at this!

PR #340, opened 2018-11-27
Removing if-env (it used the dreaded event-stream). Fixes 339.

jakearchibald:
Also updating node-sass. We'll see if the build issues have resolved.

We weren't using if-env btw (as far as I can tell)


jakearchibald:
It builds. Good.

package.json
```diff
@@ -34,12 +34,11 @@
     "html-webpack-plugin": "^3.2.0",
     "husky": "^1.1.4",
     "idb-keyval": "^3.1.0",
-    "if-env": "^1.0.4",
     "linkstate": "^1.1.1",
     "loader-utils": "^1.1.0",
     "mini-css-extract-plugin": "^0.4.4",
     "minimatch": "^3.0.4",
-    "node-sass": "^4.9.0",
+    "node-sass": "^4.11.0",
```
surma: Ah good, this seems fixed now as well :)

jakearchibald: Yeah. It was the source of `npm audit` complaints, so thought I may as well give updating a swing.

DanielRuf: Awesome 👍

PR #349, opened 2018-11-28
Specify lib for TypeScript

TimvdLippe:
This makes the typings work if you are in a worker, which the
`processor-worker` is.

(This apparently also caught an issue with the HTMLElement.children iterator)

### DIFF
--- a/src/components/compress/custom-els/MultiPanel/index.ts
+++ b/src/components/compress/custom-els/MultiPanel/index.ts
@@ -170,7 +170,7 @@ export default class MultiPanel extends HTMLElement {
 
   private _closeAll(options: CloseAllOptions = {}): void {
     const { exceptFirst = false } = options;
-    let els = [...this.children].filter(el => el.matches('[content-expanded]')) as HTMLElement[];
+    let els = [...Array.from(this.children)].filter(el => el.matches('[content-expanded]')) as HTMLElement[];
 
     if (exceptFirst) {
       els = els.slice(1);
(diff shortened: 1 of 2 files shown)


jakearchibald:
@TimvdLippe can you go into a little more detail about the issue you're seeing here? I'm not seeing any errors in `processor-worker.ts`, and neither is CI.

TypeScript is treating `processor-worker.ts` like as if it's in a document, yes, but if we're going to fix that we should isolate it the same way we've done with the service worker, right?

TimvdLippe:
@jakearchibald It was not an issue in the current codebase, but I was working on some changes that used properties in a worker (e.g. `self.onmessage`) that would not compile. I figured that separating out the changes to the lib in a PR made more sense than to bundle them in the same PR.

I could move the files to a separate directory if you want to, but I felt like the changes here made more sense, as specifying a lib at all seemed like a good thing on its own. Your call :smile:

jakearchibald:
If the problem we're trying to fix is "TypeScript doesn't realise `processor-worker.ts` is a worker", I don't think this PR is really doing that.

Unless I'm missing something, this PR makes all files in the project some kind of DOM/worker hybrid (but with some things missing like DOM iterators).

surma:
From my experience with Comlink, you need to do `lib: ["dom", "worker"]` to be able to force a file into worker mode via `declare var self: DedicatedWorkerGlobalScope;`. So this does make sense to me. But I _think_ the `Array.from()` cast can go.

jakearchibald:
Really?? Ughhh why is TypeScript so shit with workers

jakearchibald:
https://github.com/GoogleChromeLabs/squoosh/pull/351 seems to do the trick.

jakearchibald:
Fixed in #351

PR #350, opened 2018-11-28
Create CSS typings before build. Fixes #251.

### LINKED ISSUE #251: Fix css.d.ts double-build issues
It'll half our deploy time 😄.

The answer is to use https://github.com/Quramy/typed-css-modules rather than (or in addition to) the webpack plugin.

jakearchibald:
Now we only need to build once per deploy. Yay!


jakearchibald:
Boo, this is causing the build to hang. I guess the watcher isn't closing properly. Investigating…

jakearchibald:
Kinda annoying that I can't recreate the problem locally.

jakearchibald:
Fixed. Although netlify is still stuck on the previous deploy, and there doesn't appear to be a way to abort it (cc @phae, @philhawksworth).

jakearchibald:
Here's the stuck deploy https://app.netlify.com/sites/squoosh/deploys/5bfe9d10f6d5ea4d85dec2e1. I guess it'll time out at some point?

jakearchibald:
Ok, I pushed a dummy branch, and that doesn't queue in the same way the current deploy is.

https://app.netlify.com/sites/squoosh/deploys/5bfea28bdaeb0a2cb838e417 - build log
https://fix-double-buildss--squoosh.netlify.com/ - preview

webpack.config.js
```diff
@@ -15,14 +15,15 @@ const WorkerPlugin = require('worker-plugin');
...
+const addCssTypes = require('./config/add-css-types');
 
 function readJson (filename) {
   return JSON.parse(fs.readFileSync(filename));
 }
 
 const VERSION = readJson('./package.json').version;
 
-module.exports = function (_, env) {
+module.exports = async function (_, env) {
```
developit: TIL webpack allows this

developit:
Does this create a race condition for CSS updates where types might not be generated before webpack's recompilation kicks in and compiles the TS against them?

webpack.config.js
jakearchibald: I didn't realise it was a thing either

jakearchibald:
> Does this create a race condition for CSS updates where types might not be generated before webpack's recompilation kicks in and compiles the TS against them?

Maybe. But then it'd see the `.d.ts` change and recompile. I think.

PR #352, opened 2018-11-29
Ship 1.1.0

jakearchibald:
Now we have a separate branch for live, it feels like we should do a PR for updating.

All the code in this PR has been reviewed already, but it gives us an opportunity to have a quick check before flipping the switch.


jakearchibald:
Gross. GitHub doesn't provide an easy way to just move the live branch to the same position as master. Anyway, I've done it manually.

PR #353, opened 2018-11-29
Client-side API

surma:
Here’s the current, rather hacky implementation of the client-side API so we can start discussing the implementation approach.

The basic approach is to define an API and expose it to `window.parent` via Comlink, this way squoosh can be embedded using `<iframe>` and the API methods can just be invoked.

Here’s [the code](https://gist.github.com/surma/d0fdb1ff0c31e279307fcb037ce79ed4) for an example batch processing app.


src/components/App/index.tsx
```diff
@@ -87,14 +94,52 @@ export default class App extends Component<Props, State> {
     this.setState({ file: undefined });
   }
 
+  private exposeAPI() {
```
surma: I was thinking of putting this into a lazy-loaded component, but it also needs access to the internal state of the `Compress` instance. My current idea would be to get a ref and pass it to the API component. Not pretty, but it would work 🤷‍♂️

src/components/compress/index.tsx
```diff
@@ -352,7 +353,9 @@ export default class Compress extends Component<Props, State> {
       loading: true,
     });
 
-    this.setState({ images });
+    this.setState({ images }, () => this.base!.dispatchEvent(
+      new CustomEvent('squooshingdone', { bubbles: true }),
```
surma: Also somewhat of a hacky shortcut. Not sure what y’all think of this.

jakearchibald:
This is a 12% increase to our first-interaction bundle. Is there anything that can be done to avoid this?

What does usage look like? How does the user know when they can make calls to the API?

surma:
@jakearchibald PTAL! Lazy-loading the API now :)

I also updated the [example batch app](https://gist.github.com/surma/d0fdb1ff0c31e279307fcb037ce79ed4), which actually looks really nice in terms of code now. I’d love to bundle `squoosh.api.js` with our project and serve it as its own bundle from squoosh.app, but I couldn’t get webpack to do that lol.

src/components/App/index.tsx
```diff
@@ -9,6 +9,9 @@ import '../../lib/SnackBar';
 import Intro from '../intro';
 import '../custom-els/LoadingSpinner';
 
+// This is imported for TypeScript only. It isn't used.
+import Compress from '../compress';
```
jakearchibald: Use the `import('../compress')` syntax instead then?

src/components/App/index.tsx
```diff
@@ -53,6 +57,8 @@ export default class App extends Component<Props, State> {
         window.STATE = this.state;
       };
     }
+
+    import('./client-api').then(m => m.exposeAPI(this));
```
jakearchibald: You probably need to give this a `webpackChunkName` as above.

src/components/App/client-api.ts
```diff
@@ -0,0 +1,37 @@
...
+
+import { expose } from 'comlink';
+
+export function exposeAPI(app: App) {
+  self.parent.postMessage('READY', '*');
+  self.addEventListener('message', (ev: MessageEvent) => {
+    if (ev.data !== 'READY?') {
+      return;
+    }
+    ev.stopPropagation();
```
jakearchibald: What's the benefit of this?

src/components/compress/index.tsx
jakearchibald: Won't this fire way before squooshing is done? This `setState` is just for the loading state.

Feels like we need something else like the compression job promises.

src/components/compress/index.tsx
```diff
@@ -301,7 +302,7 @@ export default class Compress extends Component<Props, State> {
       // The image only needs updated if the encoder/preprocessor settings have changed, or the
       // source has changed.
       if (sourceDataChanged || encoderChanged || preprocessorChanged) {
-        this.updateImage(i, {
+        this.compressionJobs[i] = this.updateImage(i, {
```
jakearchibald: I'm not sure these reflect "abort" correctly. I'm not sure what the right thing to do there is. Can comlink handle DOMException?

jakearchibald (changes requested):
Coming along nicely. I agree that we should use this for the share target stuff.

src/components/App/index.tsx
jakearchibald: Oh, I think I wrote this, but it's been replaced. Maybe it crept back in in a rebase?

jakearchibald:
We need to decide if the API is solely for computation, or if it's a UI driver. From this PR it seems like the latter, and I think that's the right direction. That works for things like a share target API, it works for creating a glitch that opens a particular image with particular settings (once we expand the API to include settings). But that means the API needs to cope with user interactions that might happen during its driving, so calls should `AbortError` if the user interrupts.

I don't have a strong opinion if this is done with events, or promises stored on the instance.

Some things to think about:

If the user changes some compression settings on the left, then the API changes settings on the right, should the API returned promise wait on both sides to finish compressing, or just the right?

If this is a UI driver, should we separate the setting of things from the completion of things? Eg, if I `setFile`, should that resolve once the image has been set from the UI point of view, as in before it's decoded? We could have other methods for observing things like decoded, and compression complete.

What's our backwards compatibility story here? Should the API involve some sort of version check? We could change our major version when the API breaks compatibility.

surma:
> We need to decide if the API is solely for computation, or if it's a UI driver. From this PR it seems like the latter, 

The more I think about it, the more I agree it’s what we should aim for.

> But that means the API needs to cope with user interactions that might happen during its driving, so calls should AbortError if the user interrupts.

I’ll see if I can add a got way to add that.

> If the user changes some compression settings on the left, then the API changes settings on the right, should the API returned promise wait on both sides to finish compressing, or just the right?

I don’t think it should. `setFile()` just waits for the state change to happen. `getBlob(side)` waits for the selected side to finish encoding. This is where I’d need to add error handling for when that job gets aborted. I think that model works.

> If this is a UI driver, should we separate the setting of things from the completion of things? Eg, if I setFile, should that resolve once the image has been set from the UI point of view, as in before it's decoded? We could have other methods for observing things like decoded, and compression complete.

It’s mostly how it works right now, but I think you are right that more explicit events (?) for intermediate steps would be nice.

> What's our backwards compatibility story here? Should the API involve some sort of version check? We could change our major version when the API breaks compatibility.

It’s probably a good call to add a the major version to the handshake and just abort on mismatch.

surma:
@jakearchibald PTAL

src/components/compress/index.tsx
```diff
@@ -470,14 +469,39 @@ export default class Compress extends Component<Props, State> {
...
           console.error(err);
         });
       },
       delay,
     );
   }
 
+  @bind
+  private dispatchCustomEvent(name: string, detail?: {}) {
+    (this.base || document).dispatchEvent(new CustomEvent(name, { detail, bubbles: true }));
```
jakearchibald: Is there a benefit to firing it on `this.base`?

src/components/compress/index.tsx
```diff
@@ -539,8 +563,13 @@ export default class Compress extends Component<Props, State> {
           });
         }
       } catch (err) {
-        if (err.name === 'AbortError') return;
-        this.props.showSnack(`Processing error (type=${settings.encoderState.type}): ${err}`);
+        if (err.name === 'AbortError') {
+          this.signalProcessingAbort(index as 0 | 1);
```
jakearchibald: `0|1` for consistency?

jakearchibald (changes requested):

src/components/App/client-api.ts
```diff
@@ -0,0 +1,74 @@
...
+const API_VERSION = 1;
+
+export function exposeAPI(app: App) {
+  self.parent.postMessage({ type: 'READY', version: API_VERSION }, '*');
+  self.addEventListener('message', (ev: MessageEvent) => {
+    if (ev.data !== 'READY?') {
+      return;
+    }
+    ev.stopPropagation();
+    self.parent.postMessage({ type: 'READY', version: API_VERSION }, '*');
```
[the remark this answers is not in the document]
surma: Maybe. I thought there could be a scenario where an squoosh iframe is reused and the next user might send a new `READY?` message.

src/components/compress/index.tsx
surma: I found that `this.base` is not set when we are just switching to from the `Intro` view to the `Compress` view. I couldn’t come up with a better workaround.

src/components/App/client-api.ts
```diff
@@ -0,0 +1,74 @@
...
+  });
+  expose(new API(app), self.parent);
+}
+
+class API {
+  constructor(private app: App) { }
+
+  async setFile(blob: Blob, name: string) {
```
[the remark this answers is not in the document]
jakearchibald: Why not `async setFile(file: File | Fileish)` then?

src/components/compress/index.tsx
jakearchibald: Is there a disadvantage to just firing it on the document all the time?

src/components/App/client-api.ts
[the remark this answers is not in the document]
surma: `Fileish` is our own type that we can’t really expose to a client and `File` alone suffers from the Edge bug, no?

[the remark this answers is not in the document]
jakearchibald: Ah, so it is because of the file issues in Edge.

I'm torn on this. Once we drop EdgeHTML support, we can drop `Fileish`, but with this API we're kinda locking ourselves into a weird pattern because of EdgeHTML.

src/components/compress/index.tsx
surma: Not really no

src/components/App/client-api.ts
```diff
@@ -0,0 +1,74 @@
+import App from './index';
+
+import { expose } from 'comlink';
+
+const API_VERSION = 1;
```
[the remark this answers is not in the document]
surma: How would I do that in webpack land?

surma:
@jakearchibald PTAL

src/components/App/client-api.ts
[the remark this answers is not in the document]
jakearchibald: We're already doing it for `VERSION`, so I think you'd just have to add `MAJOR_VERSION`.

src/components/App/index.tsx
jakearchibald: And make sure it's cached offline https://github.com/GoogleChromeLabs/squoosh/blob/master/src/sw/util.ts#L58.

src/components/App/client-api.ts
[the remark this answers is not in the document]
jakearchibald: Ahh yes, of course

src/components/App/index.tsx
jakearchibald: `webpackChunkName` doesn't appear to be working. I think it needs to come before the string. See the other examples in this file.

src/components/compress/index.tsx
```diff
@@ -399,8 +419,7 @@ export default class Compress extends Component<Props, State> {
     // Either processor is good enough here.
     const processor = this.leftProcessor;
 
-    this.setState({ loadingCounter, loading: true });
-
+    this.setState({ loadingCounter, loading: true }, this.signalProcessingStart);
```
jakearchibald: Given that processing starts every time settings are changed, I'm not sure this is accurate. It seems like you're actually wanted an event for "file loaded".

jakearchibald (changes requested):
Caught a few more things.

Also, there should be docs for this, which we can link to from the main README.

src/components/compress/index.tsx
```diff
@@ -38,6 +38,26 @@ import { ExpandIcon, CopyAcrossIconProps } from '../../lib/icons';
...
+  START = 'squoosh:START',
+  DONE = 'squoosh:done',
+  ABORT = 'squoosh:abort',
+  ERROR = 'squoosh:error',
+}
+export interface SideEventInit extends EventInit {
+  side?: 0|1;
+  error?: Error;
+}
+export class SideEvent extends Event {
```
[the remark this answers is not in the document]
surma: What am I looking at? That linked file also creates a class that extends `Event`?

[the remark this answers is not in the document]
jakearchibald: `fixExtendedEvent`

src/components/compress/index.tsx
```diff
@@ -38,6 +38,26 @@ import { ExpandIcon, CopyAcrossIconProps } from '../../lib/icons';
...
+  public side?: 0|1;
```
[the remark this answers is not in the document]
surma: No, the start event doesn’t. I guess I should turn those into separate events?

src/components/compress/index.tsx
```diff
@@ -38,6 +38,26 @@ import { ExpandIcon, CopyAcrossIconProps } from '../../lib/icons';
...
+  public error?: Error;
```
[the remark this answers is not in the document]
surma: Left it in there. There are 3 types of events a “side” can emit. `DONE`, `ABORT` and `ERROR`, which I think makes sense.

surma:
@jakearchibald PTAL

jakearchibald (commented):
Only nits.

Add some docs and a demo, and we're good to go I think.

surma:
This is ready for final review and hopefully a merge :D

I pushed the small batch processing demo to [a glitch](https://squoosh-api-demo.glitch.me), using the real SDK file exposed from this PR endpoint.

@jakearchibald PTAL

package.json
```diff
@@ -4,9 +4,10 @@
   "version": "1.3.3",
   "license": "apache-2.0",
   "scripts": {
+    "build:sdk": "microbundle --compress -f es -o build -i src/sdk.ts",
```
developit: I think `--compress` is the default and `-i` is optional.

```suggestion
    "build:sdk": "microbundle -f es -o build src/sdk.ts",
```

surma:
@jakearchibald ping :)

_headers.ejs
```diff
@@ -9,6 +9,10 @@
 /serviceworker.js
   Cache-Control: no-cache
 
+# Module imports need CORS, so we enable it for the SDK file.
+/squoosh.mjs
+  Access-Control-Allow-Origin: *
+
```
jakearchibald: Add the header to all responses. May as well.

jakearchibald: This resource has a max-age of 31536000. Feels like it should be no-cache right?

package.json
jakearchibald: This seems to be adding a load of definition files into the build.

_headers.ejs
jakearchibald: In general, what's the benefit of self-hosting this rather than providing it as a separate package? I'm worried about making changes to this resource that breaks people's sites.

src/sdk.ts
```diff
@@ -0,0 +1,41 @@
...
+        return;
+      }
+      if (msg.version !== MAJOR_VERSION) {
+        throw Error(
+          `Version mismatch. SDK version ${MAJOR_VERSION}, Squoosh version ${
+            msg.version
+          }`,
+        );
+      }
+      ev.stopPropagation();
```
jakearchibald: Maybe we already discussed this, but what's the benefit here? What other kind of listeners do we expect?

src/components/compress/index.tsx
```diff
@@ -38,6 +38,51 @@ import { ExpandIcon, CopyAcrossIconProps } from '../../lib/icons';
...
+    super(SquooshStartEventType.START, init);
+    fixExtendedEvent(this, SquooshStartEvent);
+  }
+}
+
+export enum SquooshSideEventType {
+  DONE = 'squoosh:done',
+  ABORT = 'squoosh:abort',
+  ERROR = 'squoosh:error',
+}
```
jakearchibald: Do we need the extra enum features here, or will `const enum` do?

```ts
enum SquooshSideEventType {
  DONE = 'squoosh:done',
  ABORT = 'squoosh:abort',
  ERROR = 'squoosh:error',
}

console.log(SquooshSideEventType.DONE);
```

becomes

```js
var SquooshSideEventType;
(function (SquooshSideEventType) {
    SquooshSideEventType["DONE"] = "squoosh:done";
    SquooshSideEventType["ABORT"] = "squoosh:abort";
    SquooshSideEventType["ERROR"] = "squoosh:error";
})(SquooshSideEventType || (SquooshSideEventType = {}));

console.log(SquooshSideEventType.DONE);
```

Whereas

```ts
const enum SquooshSideEventType {
  DONE = 'squoosh:done',
  ABORT = 'squoosh:abort',
  ERROR = 'squoosh:error',
}

console.log(SquooshSideEventType.DONE);
```

Becomes

```js
console.log("squoosh:done" /* DONE */);
```

jakearchibald (changes requested):
Does this create a privacy problem?

Eg, `evil.com` contains a link "Check out squoosh.app!" which, when clicked, opens a new window. To the user it looks like a regular tab with "squoosh.app" in the address bar, but if they drop any private images onto the site, `evil.com` can read it.

It seems like we need to limit usage of this API to cases where the source of control is obvious, eg iframes, as the URL bar will display `evil.com` rather than `squoosh.app`.

src/sdk.ts
surma: The idea was that there could be a scenario where an squoosh iframe is reused and the next user might send a new READY? message. It’s mostly me playing it safe.

package.json
surma: True. Do you think that’s a problem? There’s technically no harm in emitting them and I have no easy way to teach `microbundle` to _not_ do that.

_headers.ejs
surma: I thought that this way at least SDK and Squoosh stay in sync in terms of versioning. We can also not host it and publish it as a package, but not sure it’s that different in terms of breakage.

surma:
Hopefully this limits it to iframes only.

mbrevda:
Is this blocked on something?

vroussel35:
Hi guys,
Can you let us know when you plan to deliver this request?
(no pressure but having a little bit of visibility on this client-side API functionality would be great to plan on our side)
Looking forward to hear from you.
Thanks

usualoma:
Great PR!
This PR only works on Google Chrome or Firefox, but it is still useful enough.

This is an example of integration with a CMS created based on this PR.
https://github.com/usualoma/mt-plugin-Squoosh/

I hope this PR will be merged into master.

mbrevda:
Any context on why this was closed?

dgrammatiko:
> Any context on why this was closed?

Pretty sure this was due to moving the branch `master` to `live` and in order to remove the first one this was necessary. I've faced that on my own repos. I'll guess @surma will reopen it again at some point

mbrevda:
Pretty sure you can just select a new branch for the PR without needing to close it

developit:
Changing the master branch on GitHub auto-closes all PR's targeted at that branch.

surma:
Closing this for now in favor of the CLI for batch processing (and also Share Target fulfills most of the purpose).

mbrevda:
Can Share Target be used to programmatically compress images client side (without user interaction)?

PR #363, opened 2018-12-07
Rotation optimise. Fixes #362

### LINKED ISSUE #362: 0 rotation optimisation happens too late.
The rotation method just passes the input back for a 0 rotation, but in order to do that, it has to start the worker & load the module within the worker.

That's kinda silly, and not great for performance. It also isn't great for memory, as the worker is being brought to life to do, well, nothing.

The 0 rotation check should happen before the worker.

jakearchibald:
Spotted that we were loading our rotation code when we didn't need to.

Also, our worker timeout was set to 1 second. It's supposed to be 10 and I have no idea when/how I managed to fuck that up.

### DIFF
--- a/src/components/compress/index.tsx
+++ b/src/components/compress/index.tsx
@@ -85,12 +85,18 @@ interface UpdateImageOptions {
   skipPreprocessing?: boolean;
 }
 
-function processInput(
+async function processInput(
   data: ImageData,
   inputProcessData: InputProcessorState,
   processor: Processor,
 ) {
-  return processor.rotate(data, inputProcessData.rotate);
+  let processedData = data;
+
+  if (inputProcessData.rotate.rotate !== 0) {
+    processedData = await processor.rotate(processedData, inputProcessData.rotate);
+  }
+
+  return processedData;
 }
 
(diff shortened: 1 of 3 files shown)


src/components/compress/index.tsx
```diff
@@ -85,12 +85,18 @@ interface UpdateImageOptions {
...
   processor: Processor,
 ) {
-  return processor.rotate(data, inputProcessData.rotate);
+  let processedData = data;
+
+  if (inputProcessData.rotate.rotate !== 0) {
+    processedData = await processor.rotate(processedData, inputProcessData.rotate);
+  }
+
+  return processedData;
```
jakearchibald: I could simplify this to:

```js
if (inputProcessData.rotate.rotate === 0) return data;
return processor.rotate(processedData, inputProcessData.rotate);
```

The current pattern is more like what we have for preprocessors now. I guess the pattern in the PR is a bit premature?

developit: Probably compresses to the same output ;)

jakearchibald:
Sold!

PR #368, opened 2018-12-11
Fix download error after copy settings button is pressed

yun77op:
If current slide's downloadUrl is copied to other slide's settings, then current slide's downloadUrl will be revoked.

The error can be reproduced. First click copy settings button, then click current slide's download button.


jakearchibald:
Thanks for reporting this. I can recreate the issue.

I've created a PR in https://github.com/GoogleChromeLabs/squoosh/pull/369 which is basically the same as yours but with a comment that'll hopefully prevent regressions here. Thank you!

PR #370, opened 2018-12-12
Add basic history handling (#288)

jakearchibald:
This is https://github.com/GoogleChromeLabs/squoosh/pull/309 with an extra tidy-up commit.


jakearchibald:
I moved this commit into the other PR

Issue #342, opened 2018-11-27
TypeError in offliner.ts


TimvdLippe:
**Describe the bug**
On load, I receive the following TypeError:

```
Uncaught (in promise) TypeError: Cannot read property 'controller' of undefined
    at offliner (offliner.ts:52)
    at App.offlinerPromise.then (index.tsx:54)
```

**To Reproduce**
1. Start devserver
2. Load the webpage

**Expected behavior**
No error

**Version:**
 - OS w/ version: Linux
 - Browser w/ version Chrome 70
 - Node version: 10.9.0
 - npm version: 6.4.1

**Is your issue related to the quality of image compression?**
No

**Additional context, screenshots, screencasts**

TimvdLippe:
Seems to be fixed.

Issue #345, opened 2018-11-27
Add better zoom interaction 


kosamari:
Splitting from issue #270  

-------

While fluid zooming may seem cool (and being fast – even useful), there is no replacement for seeing an image at multiples of 100%, hence: a quick way to reset zoom to 100% (doubleclick on an image? on a zoom control?) and plus and minus buttons to always go back to clearly defined steps (25% both ways off 100%?) would be cool to have.

Issue #292, opened 2018-11-15
Prevent image from being too far from splitter


fstanis:
I find it a bit annoying that zooming out can put the image in some awkward corner.

Perhaps it's a good idea to prevent the image from going too far, e.g. ensure it's (1) always visible (so you can't scroll in a way which puts it off-screen) and (2) image edges snap to splitter if taken too much to the left or right.

So if the zoom or pan results in something like this:

![image](https://user-images.githubusercontent.com/250887/48522215-a9d03600-e82c-11e8-94d9-398e721eca35.png)

...it would just snap to this:

![image](https://user-images.githubusercontent.com/250887/48522197-92914880-e82c-11e8-9a3b-ff00e6c8bd8c.png)

mathiasbynens:
FWIW, this happened with several CDS attendees that tried to use Squoosh for the first time while visiting our forum space booth.

fstanis:
> CDS attendees that tried to use Squoosh for the first time

+1 same happened with our Pixelbook booth as well. People would get confused when they zoomed and panned the image and it disappeared somewhere.

> Sometimes you might not want the handle to cover parts of the image.

True - maybe snap to image edge + some margin to account for handle?

Or maybe an easier / simpler to implement solution is to just:

- Ensure minimum zoom level is such that picture always takes at least 10-20% of the viewport.
- Ensure the picture is never fully outside viewport.
- Maybe add something like "press space to re-center" or an option to lock the image to center (disable panning). IMO panning is not that useful since the splitter is movable.

surma:
Yeah, I talked to @mathiasbynens offline.

I think snapping to the handle is the wrong solution. Something like a minimum zoom-level so that an image occupies at least 10px by 10px or something.

I think we can do better by making it harder to lose the picture, but I don’t think we can reliably prevent it 100%. 

I was wondering if we should add a double-tap gesture on the zoom percentage to reset zoom and position.

kosamari:
At meeting today, we discussed that there should be 
- one button to reset position (move to center)

Leaving this issue open & tagging as good first issue. We welcome contribution if someone want to take it!

Kirill486:
Hi, everyone)
I see there's a work related to this issue in process. But this issue looks left behind.
So, I wondered is it ok for me to come up with a solution. It still needs to be improved in a few different ways))
https://github.com/GoogleChromeLabs/squoosh/pull/574

PR #372, opened 2018-12-12
Add position reset button and update zoom interaction. (#292) (#345)

alisaitbilgi:
There are two things that I am not sure about. First one is reset button's icon. I left it as "reset" but i think, it is better to replace it with a meaningful icon :) and secondly, I just thought that for each zoom interaction triggered by zoom-in / zoom-out buttons, we should center the image in order not to loose it. When image is close to corners, if zoom-in / zoom-out buttons are pressed, it is getting disappeared. Other than those, I hope I could be useful :)

### DIFF
--- a/src/components/Output/index.tsx
+++ b/src/components/Output/index.tsx
@@ -34,6 +34,12 @@ interface State {
   scale: number;
   editingScale: boolean;
   altBackground: boolean;
+  isIntersecting: boolean;
+}
+
+interface IntersectionObserverEntry {
+  readonly intersectionRatio: number;
+  readonly isIntersecting: boolean;
 }
 
 const scaleToOpts: ScaleToOpts = {
@@ -48,26 +54,19 @@ export default class Output extends Component<Props, State> {
     scale: 1,
     editingScale: false,
     altBackground: false,
+    isIntersecting: true,
   };
   canvasLeft?: HTMLCanvasElement;
   canvasRight?: HTMLCanvasElement;
   pinchZoomLeft?: PinchZoom;
   pinchZoomRight?: PinchZoom;
   scaleInput?: HTMLInputElement;
+  threshold: number = 0;
   retargetedEvents = new WeakSet<Event>();
 
   componentDidMount() {
     const leftDraw = this.leftDrawable();
     const rightDraw = this.rightDrawable();
-
-    // Reset the pinch zoom, which may have an position set from the previous view, after pressing
-    // the back button.
-    this.pinchZoomLeft!.setTransform({
-      allowChangeEvent: true,
-      x: 0,
-      y: 0,
-      scale: 1,
-    });
 
     if (this.canvasLeft && leftDraw) {
       drawDataToCanvas(this.canvasLeft, leftDraw);
@@ -75,6 +74,11 @@ export default class Output extends Component<Props, State> {
     if (this.canvasRight && rightDraw) {
       drawDataToCanvas(this.canvasRight, rightDraw);
     }
+
+    // Reset the pinch zoom, which may have an position set from the previous view, after pressing
+    // the back button.
+    this.resetPosition();
+    this.observeIntersection();
   }
 
   componentDidUpdate(prevProps: Props, prevState: State) {
@@ -127,12 +131,78 @@ export default class Output extends Component<Props, State> {
     return !shallowEqual(this.props, nextProps) || !shallowEqual(this.state, nextState);
   }
 
+  @bind
+  private async initializeImage() {
+    await this.setRotation(true)();
+    this.setZoom(1)();
+    this.resetPosition();
+    this.setState({ altBackground: false });
+  }
+
+  @bind
+  private handleIntersect(entries: IntersectionObserverEntry[]) {
+    entries.forEach((entry: IntersectionObserverEntry) => {
+      // Value of isIntersecting is depended on threshold value on chrome.
+      // However, for safari, firefox and polyfill we just need to check also intersectionRatio.
+      // Realized different behavior: https://github.com/w3c/IntersectionObserver/issues/345
+      const isIntersecting = entry.isIntersecting && entry.intersectionRatio > this.threshold;
+      this.setState({ isIntersecting });
+    });
+  }
+
+  @bind
+  private async observeIntersection() {
+    if (!('intersectionObserver' in window)) {
+      await import('intersection-observer');
+    }
+
+    if (!this.pinchZoomLeft || !this.canvasLeft) return;
+
+    const options = {
+      root: this.pinchZoomLeft,
+      rootMargin: '0px',
+      threshold: this.threshold,
+    };
+    const observer = new IntersectionObserver(this.handleIntersect, options);
+
+    observer.observe(this.canvasLeft);
+  }
+
   private leftDrawable(props: Props = this.props): ImageData | undefined {
     return props.leftCompressed || (props.source && props.source.processed);
   }
 
   private rightDrawable(props: Props = this.props): ImageData | undefined {
     return props.rightCompressed || (props.source && props.source.processed);
+  }
+
+  // initial coordinates depends on the current scale and dimensions of the image.
+  @bind
+  private resetPosition(scaleRatio: number = this.state.scale) {
+    if (this.canvasLeft) {
+      const { width, height } = this.canvasLeft;
+
+      this.pinchZoomLeft!.setTransform({
+        allowChangeEvent: true,
+        x: (width / 2) * (1 - scaleRatio),
+        y: (height / 2) * (1 - scaleRatio),
+        scale: scaleRatio,
+      });
+    }
+  }
+
+  private setZoom(scaleRatio: number = 1) {
+    return () => {
+      if (!this.pinchZoomLeft) throw Error('Missing pinch-zoom element');
+
+      this.pinchZoomLeft.scaleTo(scaleRatio, scaleToOpts);
+
+      // Now, reset position will be triggered when the image
+      // has been lost from the viewport with 0.2 threshold.
+      if (!this.state.isIntersecting) {
+        this.resetPosition(scaleRatio);
+      }
+    };
   }
 
   @bind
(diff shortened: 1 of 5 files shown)


alisaitbilgi:
> Thanks for your pull request. It looks like this may be your first contribution to a Google open source project (if not, look below for help). Before we can look at your pull request, you'll need to sign a Contributor License Agreement (CLA).
> 
> 📝 **Please visit https://cla.developers.google.com/ to sign.**
> 
> Once you've signed (or fixed any issues), please reply here (e.g. `I signed it!`) and we'll verify it.
> 
> #### What to do if you already signed the CLA
> ##### Individual signers
> * It's possible we don't have your GitHub username or you're using a different email address on your commit. Check [your existing CLA data](https://cla.developers.google.com/clas) and verify that your [email is set on your git commits](https://help.github.com/articles/setting-your-email-in-git/).
> 
> ##### Corporate signers
> * Your company has a Point of Contact who decides which employees are authorized to participate. Ask your POC to be added to the group of authorized contributors. If you don't know who your Point of Contact is, direct the Google project maintainer to [go/cla#troubleshoot](http://go/cla#troubleshoot) ([Public version](https://opensource.google.com/docs/cla/#troubleshoot)).
> * The email used to register you as an authorized contributor must be the email used for the Git commit. Check [your existing CLA data](https://cla.developers.google.com/clas) and verify that your [email is set on your git commits](https://help.github.com/articles/setting-your-email-in-git/).
> * The email used to register you as an authorized contributor must also be [attached to your GitHub account](https://github.com/settings/emails).

I signed it!

jakearchibald:
@kosamari @mustafa-x @developit @surma what do you think about this?

**Clicking zoom buttons centers image.** I think we debated this at some point, but I think it might make sense, especially if the image at the destination zoom can be contained within the view. Maybe not when it covers the view.

The zoom seems to have a minimum of 25% which seems too high. Also, it feels like the steps should be more exponential.

**Reset button.** I expected it to also reset zoom to 100%. I'm not sure if this is useful.

I haven't really looked at the code yet, so I'm just looking at the interactions.

alisaitbilgi:
@jakearchibald thanks a lot for the initial reviews,
*Reset button*: It just resets the position as mentioned at #292 . On the other hand, when you double-click, image is reset to 100% zoom and center position as mentioned at #345 . But, I think two issues can be somehow merged and since zoom-in/out buttons now resets the position (if you think so), it may become redundant to have as a special button.

jakearchibald:
Ping @kosamari @mustafa-x @developit @surma. Curious to know your thoughts on this feature.

Preview: https://deploy-preview-372--squoosh.netlify.com/

surma:
Thanks @alisaitbilgi for working on this! 

Here’s my feedback:
- I agree with Jake that the reset functionality should also reset zoom level to 100%
- I feel weird about the +/- buttons resetting the position. My gut feeling is that a common workflow is to center the ROI and then zoom in and out, trying different codecs. That would get really annoying with this change.
- Locking to 25% steps seems quite big at the smaller zoom levels and very small for the higher zoom levels. The difference from 25% to 50% is _massive_, while the difference from 625% to 650% is almost negligible. I think exponential steps are preferable.
- I am a bit hesitant on the button as screen estate is scarce, especially on mobile. As you mentioned, we originally considered to reset zoom _and_ position when you double-click the zoom value. Do you wanna do that instead?

mustafa-x:
At a glance, I thought the reset button would reset everything to its initial loaded state. Also the text feels a tad too small. Zooming should be consistent, the snapping to the centre looks a bit broken. 

The rotate buttons has a weird glitch where it warps the image before rotation. 


idea-small

jakearchibald:
The rotate glitch is unrelated to this PR. I think that happened once we started debouncing encoding.

alisaitbilgi:
@jakearchibald , @surma and @mustafa-x  thanks a lot for the reviews. Here are my ideas:

- **Reset button**:  I think I can remove that button and instead, extend the behaviour of **double click** as making it reset the image to its initial position. 
- **Zoom level steps**:  It can stay as it is. (Already exponential)
- **Resetting position**: I think to overcome the loosing the image issue when zoom in/out, instead of resetting its position on every click, I can observe the intersection within the viewport and then, reset the position when it is not contained in the view.

What do you think?

surma:
>  I can observe the intersection within the viewport and then, reset the position when it is not contained in the view.

That sounds reasonable! I’d have to actually play with it to see if it makes sense, but it sounds good on paper!

alisaitbilgi:
> > I can observe the intersection within the viewport and then, reset the position when it is not contained in the view.
> 
> That sounds reasonable! I’d have to actually play with it to see if it makes sense, but it sounds good on paper!

Thanks a lot @surma , I'm planning to use [Intersection Observer API](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API) but, haven't decided the threshold value yet, I think it is better to see then decide.

alisaitbilgi:
By the way, sorry about forgotten intersection observer polyfill. I don't know the right way is whether to  install [polyfill](https://github.com/w3c/IntersectionObserver/tree/master/polyfill) or just feature detect the API?

src/components/Output/index.tsx
```diff
@@ -127,6 +123,35 @@ export default class Output extends Component<Props, State> {
...
+  @bind
+  private observeIntersection() {
+    if (!this.pinchZoomLeft || !this.canvasLeft) return;
+
+    const options = {
+      root: this.pinchZoomLeft,
+      rootMargin: '0px',
+      threshold: 0.5,
+    };
+    const observer = new IntersectionObserver(this.handleIntersect, options);
```
surma: You’d want a feature detection here and lazy-load a polyfill if `IntersectionObserver` is unavailable.

surma (changes requested):

surma:
Good work!

Couple of things:

- Yes, definitely lazy-load a IO polyfill :)
- I don’t think we want to reset on double-click _anywhere_. It’s a common gesture to zoom on double-tap on native apps. While we don’t necessarily need to implement _that_ gesture, we definitely shouldn’t change the meaning of the gesture. I’d say we put the double click gesture onto the percentage.
- It’s still fairly easy to lose the image if you put the image in one corner and your mouse in the opposite corner
squoosh

alisaitbilgi:
Thank you @surma :) 

- When I load the [polyfill](https://github.com/w3c/IntersectionObserver/tree/master/polyfill) at first glance, I've realized an issue about threshold's being handled. I am first planning to be sure about the issue and try to fix, then apply here as I think it's an obvious one. 
- After some conversations I thought that with **double-click** we can have a **reset-everything** behaviour, but I'm not really sure about what to do :) 
- First I just controlled the image to hold it in viewport for every user interactions then I thought that, mouse is already a user controlled thing, so it might be better to leave them as they want. I think I can implement two options ( Reset position for every interactions or just for buttons ) then we can see and decide. 

What do you think?

alisaitbilgi:
> * When I load the [polyfill](https://github.com/w3c/IntersectionObserver/tree/master/polyfill) at first glance, I've realized an issue about threshold's being handled. I am first planning to be sure about the issue and try to fix, then apply here as I think it's an obvious one.

Related issue: https://github.com/w3c/IntersectionObserver/issues/345

alisaitbilgi:
Here is the implementation @surma , I've just added a **( weird :) )** checkbox on top-right corner, so that we can see the difference between controlling 'wheel event' or not.

surma:
- Good work on the lazyloading. That looks good to me.

- The zooming with the checkbox enabled is definitely worse because I can’t zoom into the top right corner of the image. I think what I showed in the GIF is a corner case, so we can ignore that one. The behavior when the checkbox is disabled seems good :)

- I think double-click _on the percentage field_ to reset everything makes more sense than double-click anywhere.

alisaitbilgi:
Hey @surma , I think I made things more complicated while fixing my 'commit author' mistake. :)  I was using my brother's computer and his name has appeared as commit author for my 2 commits. ([1dceeae](https://github.com/GoogleChromeLabs/squoosh/pull/372/commits/1dceeae94e2c23dbed0c8ec68ae18e8910eaadbb) and [868e8bb](https://github.com/GoogleChromeLabs/squoosh/pull/372/commits/868e8bb82410b2819a1edd0218060fbe0852b9bd) I've tried to update them but since one of them ([1dceeae](https://github.com/GoogleChromeLabs/squoosh/pull/372/commits/1dceeae94e2c23dbed0c8ec68ae18e8910eaadbb)) is a merge commit, the situation got complicated. What do you think?

surma:
Uh, this looks a bit messy, doesn’t it 😂 

I’d say we just squash them and go from there. You comfortable doing that? Otherwise, with your permission, I can do it for you.

alisaitbilgi:
Thanks a lot for the help :) , you mean to squash my latest (messy) commits?

surma:
Yeah, basically turn this PR into a single commit based on current master, and then we can go from there :)

alisaitbilgi:
Sorry for late response, @surma. I am not sure about squashing merge commits and re-writing their history. I hope, didn't make things much more complicated.

surma:
@alisaitbilgi I fixed it for you :)

Please make sure you reset your local branch appropriately.

In case you don’t know how that works, here is a step-by-step:

```
$ git fetch --all
$ git checkout update-zoom-interaction
$ git reset --hard origin/update-zoom-interaction
```

(You might have to replace `origin` with whatever you called your fork)

alisaitbilgi:
Thanks a lot and done :) What about the PR then?

surma (approved):

surma:
Right! So this looks good to me!

Imma hand the final LGTM to @jakearchibald for a sanity check. Thanks so much @alisaitbilgi

jakearchibald:
Can someone write a summary of what the intended changes are from a user perspective?

Right now, if I open the image of the red panda, position its eye in the centre, and start pressing +, at some point the image jumps so I'm no longer zooming into the eye. It isn't clear if this is an error or intended behaviour.

Also, if I double click the % value, I get "Processing error". I'm assuming that isn't intended behaviour 😄.

alisaitbilgi:
> Can someone write a summary of what the intended changes are from a user perspective?

We decided to make 2 things:
1) Add reset everything behavior when % value is double clicked.
2) Add reset position behavior when image is lost from the viewport (threshold: 0.5) if it's being lost by clicking zoom in/out buttons.

> Also, if I double click the % value, I get "Processing error". I'm assuming that isn't intended behaviour 😄.

Yeah, I was looking at that error. I was using '360' for rotation value and now updating as '0'. (this will fix) On the other hand, found one more error for development mode and investigating to fix it. HMR cannot dynamically import 'intersection-observer-polyfill' and throws an error.

alisaitbilgi:
Here are the fixes @surma.

1) There was an HMR issue. ( other than rotation issue which is also fixed ) Since dynamically importing 'intersection-observer' fails (because of HMR), next lines are re-run while prev-lines becomes unchanged and not compiled/run( I have no idea why ) an error was thrown. I've fixed this dev-only issue by only swapping 2 checks, but it may be done something like that:

```
if (!('intersectionObserver' in window)) {
      if (process.env.NODE_ENV === 'development') {
        require('intersection-observer');
      } else {
        await import('intersection-observer');
      }
}
```

   I thought, writing a dev-only code is really a bad idea. So, only swap lines as a workaround. 

2) There was a closed issue #342. I think it is still alive and tried to fix it.

src/lib/offliner.ts
```diff
@@ -49,6 +49,8 @@ export async function offliner(showSnack: SnackBarElement['showSnackbar']) {
     navigator.serviceWorker.register('../sw');
   }
 
+  if (!('serviceWorker' in navigator)) return;
```
surma: This should happen before the `register` call.

surma (changes requested):
- Double-clicking the percentage to reset works for me without a processing error. @jakearchibald can you confirm?

- The +/- buttons are still confusing. I don’t the the position should reset when I zoom into a part of a picture. It should only reset once the image is completely invisible in my opinion.


kapture 2019-03-05 at 12 41 32

PR #393, opened 2018-12-17
Fixing sharp & preprocess settings. Fixes #392.

### LINKED ISSUE #392: WebP encoder: can't use_sharp_yuv
`use_sharp_yuv` option is not affects an image.

**To Reproduce**
I use checkbox in app.
Also download codec from this repo.
Any combinations of settings can't force encoder to use sharp subsampling.

**Expected behavior**
Colors of output image should be close to source.

**Versions**
OS: win 10
Browsers: Chrome Canary 72.0.3623.0, FF Developer 65.0b4

**Source**

unilever

**Output of Squoosh**

unilever squoosh-sharp-yuv

**For comparison** — result of command

```
cwebp -sharp_yuv unilever.png -o unilever.cwebp-sharp-yuv.webp
```

unilever cwebp-sharp-yuv

jakearchibald:
Going to ping the experts about `use_argb`. I'm not entirely sure what the option does. I mean, we're always giving the encoder ARGB, and it seems to be treating it as ARGB anyway, but this seems to be set for `use_sharp_yuv` to have an effect.


jakearchibald:
Actually, I figured it out while writing an email to the experts 😀.

`use_argb` is preferred for bits of processing, but things are faster if it's off. I've changed the c++ so it's only used if it's needed.

PR #394, opened 2018-12-17
Debouncing input. Fixes #277

### LINKED ISSUE #277: Smoothing selection bar creates jank
Using the smoothing option selection bar and drag it around is really janky on Chrome (on Mac). When I load the large image and grab the slider and move it around, the following performance jank happens:
image

jakearchibald:


src/components/compress/index.tsx
```diff
@@ -240,6 +240,7 @@ export default class Compress extends Component<Props, State> {
   private readonly encodeCache = new ResultCache();
   private readonly leftProcessor = new Processor();
   private readonly rightProcessor = new Processor();
+  private readonly updateImageTimeoutIds: [number, number] = [0, 0];
```
TimvdLippe: My hunch is that these 2 zeroes represent the imaginary timeout IDs of both processors, to not have optional numbers in TypeScript here? I think it would be clearer to make this an object instead or else add a clarifying comment.

src/components/compress/index.tsx
```diff
@@ -457,6 +457,21 @@ export default class Compress extends Component<Props, State> {
...
+    // Otherwise, the thrashing causes jank, and sometimes crashes iOS Safari.
+    clearTimeout(this.updateImageTimeoutIds[index]);
+
+    this.updateImageTimeoutIds[index] = self.setTimeout(
+      () => {
+        this.updateImage(index, options).catch((err) => {
+          console.error(err);
+        });
+      },
+      100,
```
surma: Tiny nit: Put this in a top-level const?

PR #395, opened 2018-12-17
Preventing zoom in iOS Safari.

jakearchibald:
It's a shame that this ends up in our main bundle, but can't think of a better way around it.

It's so small that putting it in its own module doesn't seem worth it.

### DIFF
--- a/src/components/App/index.tsx
+++ b/src/components/App/index.tsx
@@ -62,6 +62,14 @@ export default class App extends Component<Props, State> {
       };
     }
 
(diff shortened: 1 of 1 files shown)


jakearchibald:
> I’m guessing there is not equivalent <meta> or something that iOS Safari understands?

It doesn't seem like there is.

> I wonder if we can get away with adding that handler to the Compress UI on the basis that it’s much less likely for people to zoom by accident on the Intro screen.

Yeah, I tried this. If the user zooms on the main screen, they're then locked into that zoom once they get to the compression screen. It seems kinda worse since they can't even zoom out.

PR #437, opened 2019-01-24
Assemblyscript experiment

surma:


surma:
Wait hold on I don’t think I’m not actually loading the module.

surma:
Stupid rebase.

@jakearchibald, PTAL :)

codecs/rotate/package.json
```diff
@@ -0,0 +1,9 @@
+{
+  "name": "rotate",
+  "scripts": {
+    "build": "mv rotate.{as,ts} && asc rotate.ts -b rotate.wasm --validate --optimize && mv rotate.{ts,as}"
```
MaxGraey: Nice!. Btw you could get more performant output use `-O3` instead `--optimize`

codecs/rotate/rotate.as
```diff
@@ -0,0 +1,63 @@
+export function rotate(inputWidth: i32, inputHeight: i32, rotate: i32): void {
+  const bpp = 4;
+  let offset = inputWidth * inputHeight * bpp;
```
jakearchibald: `const`?

src/codecs/rotate/processor.ts
```diff
@@ -1,73 +1,30 @@
...
+  opts: RotateOptions,
+): Promise<ImageData> {
+  const flipDimensions = opts.rotate % 180 !== 0;
+  // Number of wasm memory pages (á 64KiB) needed to store the image twice.
+  const bytesPerImage = data.width * data.height * 4;
+  const numPagesNeeded = Math.ceil(bytesPerImage * 2 / (64 * 1024));
+  // const memory = new Uint8ClampedArray(data.data.length * 2);
+  // const memory = new WebAssembly.Memory({
+  //   initial: Math.ceil(data.width * data.height * 4 * 2 / (64 * 1024)),
+  // });
```
jakearchibald: Delete commented stuff?

src/codecs/rotate/processor.ts
```diff
@@ -1,73 +1,30 @@
...
-  for (let d2 = d2Start; d2 >= 0 && d2 < d2Limit; d2 += d2Advance) {
-    for (let d1 = d1Start; d1 >= 0 && d1 < d1Limit; d1 += d1Advance) {
-      const start = ((d1 * d1Multiplier) + (d2 * d2Multiplier));
-      outB[i] = inB[start];
-      i += 1;
-    }
-  }
-
-  return out;
+  instance.exports.memory.grow(numPagesNeeded);
```
jakearchibald: I guess there isn't a way to set the memory during initialisation?

src/codecs/rotate/processor.ts
```diff
@@ -1,73 +1,30 @@
...
+  const view = new Uint8ClampedArray(instance.exports.memory.buffer);
+  view.set(data.data);
+  (self as any).instance = instance;
```
jakearchibald: Is this debugging code?

codecs/rotate/rotate.as
```diff
@@ -0,0 +1,63 @@
...
+    d1Multiplier = inputWidth;
+    d2Start = inputWidth - 1;
+    d2Limit = inputWidth;
+    d2Advance = -1;
+    d2Multiplier = 1;
+  }
+
+  for (let d2 = d2Start; d2 >= 0 && d2 < d2Limit; d2 += d2Advance) {
+    for (let d1 = d1Start; d1 >= 0 && d1 < d1Limit; d1 += d1Advance) {
+      let start = ((d1 * d1Multiplier) + (d2 * d2Multiplier));
```
MaxGraey: you could cache `(d2 * d2Multiplier)` and move out of internal loop:
```ts
  for (let d2 = d2Start; ... ) {
    let d2offset = d2 * d2Multiplier;
    for (let d1 = d1Start;  ...) {
      let start = d1 * d1Multiplier + d2offset;
```

jakearchibald: Interesting. For some reason I thought the compiler might have optimised for this.

MaxGraey: LICM pass currently disable on binaryen

codecs/rotate/package.json
surma: Interesting! Why doesn’t the scaffold of `asinit .` use that?

codecs/rotate/rotate.as
surma: No. So AssemblyScript is not quite like TypeScript in that regard. AssemblyScript uses `const` to denote _compile time_ constants.

src/codecs/rotate/processor.ts
surma: Yes, thank you. My bad.

src/codecs/rotate/processor.ts
surma: Haven’t found a way. @MaxGraey?

src/codecs/rotate/processor.ts
surma: ...

codecs/rotate/rotate.as
surma: I was working with the same assumptions as @jakearchibald. I’ll move it out.

codecs/rotate/rotate.as
jakearchibald: Ohhhh interesting!

codecs/rotate/package.json
MaxGraey: `optimize` flag equivalent to `-O2s` - which mean moderate speed and size. This do most of compilers because `-O3` produce maximal performant output but also slow compilation and not always best binary size.

surma: I know `-Os`, `-Oz`, `-O2` and `-O3`, but this is the first time I have heard of `-O2s`. 

Considering we are at ~300 bytes, I pushed for `-O3` ;)

surma:
@jakearchibald PTAL

src/codecs/rotate/processor.ts
MaxGraey: you can do it during instantiation inside `env` object:
```js
const limitInBytes = 2 * 1024 * 1024; // 2 Mb
const limitPages = ((limitInBytes + 0xffff) & ~0xffff) >>> 16; // convert to pages
env: {
   ...
   memory: new WebAssembly.Memory({ initial: 1, maximum: limitPages })
} 
```

Also need add `--importMemory` for `asc`

codecs/rotate/package.json
```diff
@@ -0,0 +1,9 @@
+{
+  "name": "rotate",
+  "scripts": {
+    "build": "mv rotate.{as,ts} && asc rotate.ts -b rotate.wasm --validate -O3 && mv rotate.{ts,as}"
```
jakearchibald: Does asc depend on the extension?

MaxGraey: Currently it uses only `.ts`.

jakearchibald:
I'll do some double checking tomorrow but this looks good

src/codecs/rotate/processor.ts
surma: Ah cool. Thanks!

surma: I was missing the flag.

codecs/rotate/rotate.as
MaxGraey: this behaviour probably will change in future for consistent with javascript. But currently yes, in this context better use `let` or `var`.

codecs/rotate/rotate.as
```diff
@@ -0,0 +1,64 @@
...
+    d1Multiplier = inputWidth;
+    d2Start = inputWidth - 1;
+    d2Limit = inputWidth;
+    d2Advance = -1;
+    d2Multiplier = 1;
+  }
+
+  for (let d2 = d2Start; d2 >= 0 && d2 < d2Limit; d2 += d2Advance) {
+    let d2offset = d2 * d2Multiplier;
+    for (let d1 = d1Start; d1 >= 0 && d1 < d1Limit; d1 += d1Advance) {
```
MaxGraey: It seems `d1 >= 0` and `d2 >= 0` is unnecessary. Because `d2Start` and `d1Start` never changed during loop and never less than zero. wdyt?

codecs/rotate/rotate.as
```diff
@@ -0,0 +1,64 @@
...
+    d1Limit = inputHeight;
+    d1Advance = 1;
+    d1Multiplier = inputWidth;
+    d2Start = inputWidth - 1;
+    d2Limit = inputWidth;
+    d2Advance = -1;
+    d2Multiplier = 1;
+  }
+
+  for (let d2 = d2Start; d2 >= 0 && d2 < d2Limit; d2 += d2Advance) {
```
MaxGraey: You could potentially speedup this more by make 4 different branches for different combinations of `d1Advance` and `d2Advance`. Like:

```ts
// d1Advance: 1, d2Advance: 1
if (d1Advance == 1 && d1Advance == 1) {
   for (let d2 = d2Start; d2 < d2Limit; d2++) {
      ...
      for (let d1 = d1Start; d1 < d1Limit; d1++) {
      ...
      }
   }
}
// d1Advance: -1, d2Advance: 1
else if (d1Advance == -1 && d2Advance == 1) {
  for (let d2 = d2Start; d2 < d2Limit; d2++) {
      ...
      for (let d1 = d1Start; d1 >= 0; d1--) {
      ...
      }
   }
}
// d1Advance: 1, d2Advance: -1
else if (d1Advance == 1 && d2Advance == -1) {
   ...
}
// // d1Advance: -1, d2Advance: -1
else {
   ...
}
```

MaxGraey:
Also you could improve readability and avoid low level `load`/`store` for raw memory pointers with this zero-cost abstraction class:

```ts
class Pointer<T> {
  constructor(offset: usize = 0) {
    return changetype<Pointer<T>>(offset);
  }

  @inline @operator("[]") 
  get(index: i32): T { // overload operator for getter `ptr[index]`
    return load<T>(changetype<usize>(this) + (<usize>index << alignof<T>()));
  }

  @inline @operator("[]=") 
  set(index: i32, value: T): void { // overload operator for setter `ptr[index] =`
    store<T>(changetype<usize>(this) + (<usize>index << alignof<T>()), value);
  }
}
```
and now define some one pointer which reference to global memory with some offsets:
```ts
let offset = inputWidth * inputHeight * bpp;

let input  = new Pointer<u32>(0);
let output = new Pointer<u32>(offset);
...
for (let d2 = d2Start; /*...*/) {
   for (let d1 = d1Start; /*...*/) {
      let start = d1 * d1Multiplier + d2offset;
      output[i] = input[start];  // now access to pointer entity as usual without load/store
      i++;
   }
}


```

codecs/rotate/package.json
MaxGraey: you can move as code inside `assembly` directory and add it to exclude in root `tsconfig.json`

jakearchibald:
@MaxGraey thanks for all your input on this, but I think we're going to go for Rust instead as the result is faster (although the wasm ends up larger).

@surma is writing up his experiences with this. We're really happy with where AssemblyScript is heading.

codecs/rotate/rotate.as
surma: I mean, we are already pretty fast, so I don’t see the need to sacrifice readability/elegance for speed.

Are there plans to integrate these kind of optimizations into `asc`?

surma:
@MaxGraey That Pointer class seems incredibly useful. Could it be bundled with an asc “standard library”?

surma:
@MaxGraey I was actually also trying to create a `ArrayBuffer` instance with a given address and length, but couldn’t get it to work. Maybe that’s another possible solution?

codecs/rotate/rotate.as
MaxGraey: Actually this kind of optimizations don't do by compiler. It's quite complicated even for LLVM

MaxGraey:
> @MaxGraey I was actually also trying to create a ArrayBuffer instance with a given address and length, but couldn’t get it to work. Maybe that’s another possible solution?

If you want create instance for `ArrayBuffer` or some typed array you should include allocator at first. AS support 3 types of allocators: `arena`, `tlsf` and `buddy`. If haven't plan deallocations and rust reset memory to initial state best solution use: `import "allocator/arena"`. After that you can instantiate classes on heap like `var arr = new UInt8Array(1);  var dataView = new DataView(arr.buffer)`. But you use external raw memory and this most simple and performant solution for you case.

surma:
I don’t plan on adding an allocator since we are not doing any allocations. The zero-cost `Pointer` abstraction seems incredibly useful for these kind of use-cases. Would love to see it shipped with `asc`.

MaxGraey:
@surma This abstractions already present but in [experimental mode](https://github.com/AssemblyScript/assemblyscript/blob/master/tests/compiler/std/pointer.ts) currently

MaxGraey:
@MaxGraey this not always true. See [this](https://github.com/AssemblyScript/assemblyscript/tree/master/examples/n-body#benchmark) and [this](https://github.com/mapbox/mapbox-gl-js/issues/4835#issuecomment-415191951) benchmarks. Of course in some situations Rust will be faster but not so much and usually in several times larger in binary size.

surma:
@MaxGraey Sorry, we didn’t mean “Rust is faster than ASC in general”, but that the Rust version of this particular program ends up being faster than the as version (~500ms vs ~300ms on a 4k by 4k image).

MaxGraey:
@surma Hmm. Could you share Rust version? Pretty interesting how LVM optimize this loops. This your AS version in WAS: https://webassembly.studio/?f=xq1pz1afjtf with measurements output in console. WAS also support Rust's template: https://webassembly.studio

surma:
@MaxGraey Absolutely! Here’s the [JS, ASC, C and Rust version](https://gist.github.com/surma/0eb306fa9acc8bdf2f58150b2f1e82b4) that I have been comparing.

Most recent results (when running with `?single`):

| Language | ms |
| --- |  ---|
| JS  | 336 |
|C | 355|
|ASC | 426|
|Rust | 293|

MaxGraey:
@surma 
I see "asc rotate.ts -b assemblyscript.wasm --validate --optimize" instead "asc ... -O3". `O3` usually `2x` faster than `optimize`

surma:
Closing this as I will re-open a new PR for Rust. But if anyone find anything new, please feel free to comment!

MaxGraey:
@surma Thanks for bench! My results:

### Chrome 72.0.3626.71 beta:
**JavaScript**
```
avg: 0.8487424999722861
``` 
**AssemblyScript**
```
avg: 1.1395414999424247
```
**Rust**
```
avg: 0.7038834999810206
```

### Firefox 65.0:
**JavaScript**
```
avg: 1.1848
``` 
**AssemblyScript**
```
avg: 0.668
```
**Rust**
```
avg: 0.5601
```

surma:
You are not running the page with `?single`. 

The benchmarks with 10000 iterations is not realistic as in our scenario the code will run once or twice, not 10000 times, so not much chance of any warm ups.

MaxGraey:
With `?single`:

### Chrome 72 beta:
js avg: 176.27499999071006
AS avg: 241.69500000425614
Rust avg: 203.32000000053085

### Firefox 65.0:
js avg: 267
AS avg: 191
Rust avg: 199

### Safari 12.0.2:
js avg: 730
AS avg: 234
Rust avg: 310

Hmm, pretty strangle. I think warmup need anyway. Because all engines using tier compilation. For example Chrome using `liftoff` in this case.

MaxGraey:
Also I use slightly modified version of AS in that time:
https://gist.github.com/MaxGraey/b9fd909d1da44aa4516c70fab1337712

MaxGraey:
I used `rustc 1.33.0-nightly (ceb251214 2019-01-16)` btw

surma:
>  I think warmup need anyway.

Yes, a warmup phase will increase performance, but is just not realistic for our use-case. The code will run once, maybe twice, that’s it.

MaxGraey:
Ok, in this case I can't see huge difference between AS and Rust. How you got 2x speed up on Rust compare to AS? May be I miss something?

surma:
I am not sure. It’s probably because we are testing on different machines. I am only running the page on my MBP on Mac OS.

screen shot 2019-01-26 at 2 12 10 pm

MaxGraey:
@surma I also run on MBP on Mac OS. But I tested on Chrome 72 instead Chrome 71. May be this matter.

Btw I refactored `rotate.ts` as I suggested earlier. You can see it here:
https://gist.github.com/MaxGraey/ab8654e65b00d8427ea1121add94fbdd

with this version I got following results for `?single`:

### Chrome 72 beta:

js avg: 224.66999999596737
AS avg: 207.34000000811648
Rust avg: 203.6400000069989

### Firefox 65.0:

js avg: 272
AS avg: 195
Rust avg: 204

### Safari 12.0.2:

js avg: 716
AS avg: 193
Rust avg: 300

PR #438, opened 2019-01-26
Rotate implementation in Rust

surma:
Alright, here we go! Rotate implementation in Rust!

(FYI @ashleygwilliams)


src/codecs/rotate/processor.ts
```diff
@@ -1,73 +1,26 @@
...
-  const out = new ImageData(outputWidth, outputHeight);
-  let i = 0;
+export async function rotate(
+  data: ImageData,
+  opts: RotateOptions,
+): Promise<ImageData> {
+  const flipDimensions = opts.rotate % 180 !== 0;
+  // Number of wasm memory pages (á 64KiB) needed to store the image twice.
+  const bytesPerImage = data.width * data.height * 4;
+  const numPagesNeeded = Math.ceil(bytesPerImage * 2 / (64 * 1024));
```
jakearchibald: We start at an offset of 4. Does this take that into account?

src/codecs/rotate/processor.ts
```diff
@@ -1,73 +1,26 @@
...
+  const { instance } = (await (WebAssembly as any).instantiateStreaming(
+    fetch(wasmUrl),
+  )) as { instance: RotateModuleInstance };
```
jakearchibald: Our other wasm modules can be used many times per instantiation. Do we want to do the same here?

surma:
There was a new Ubuntu and a new Rust image. I assume that’s where the difference came from. Making a new, up-to-date build.

surma:
🎉

PR #439, opened 2019-01-26
Update libwebp to 1.0.2

EwoutH:


surma:
It seems that libwebp v1.0.2 doesn’t compile — at least not in the same way it used to before:

```
$ npm run build

> webp_enc@ build /Users/surma/src/github.com/GoogleChromeLabs/squoosh/codecs/webp_enc
> docker run --rm -v $(pwd):/src trzeci/emscripten ./build.sh

=============================================
Compiling wasm bindings
=============================================
node_modules/libwebp/src/enc/histogram_enc.c:1189:19: error: cannot initialize a variable of type 'uint16_t *const' (aka 'unsigned short *const') with an rvalue of type 'void *'
  uint16_t* const map_tmp =
                  ^
1 error generated.
```

EwoutH:
@surma Great, thanks! Doing some more Travis runs here: https://travis-ci.com/EwoutH/squoosh/builds

Looks like the CI has no problem with it (except for Windows, but that was already broken).

surma:
That's because we check in the wasm files into git. Since you didn't rebuild, everything is working fine, but it's still the old library ;)

surma:
@EwoutH I fixed the build process by using libwebp’s cmake file. Does this look good  to you?

@jakearchibald PTAL

jakearchibald:
It seems like the changes in the build have increased the size of the wasm significantly:

|      | Master | This PR | This PR but with 1.0.0 |
|------|--------|---------|------------------------|
| wasm | 184684 | 261136  | 257141                 |
| js   | 48488  | 48632   | 48632                  |

I'll have a quick play and see if I can see anything obvious.

surma:
We might be able to reduce the file size again by looking the `configure` flags. I can take a look later as well

jakearchibald:
Yeah I'm out of my depth here. I tried setting a few flags but no change in size, so I guess I was doing it wrong.

My hunch is the optimise flags aren't being carried through to cmake, but I have no evidence for this.

surma:
Fixed it!

- `webp_enc.wasm`: 172K
- `webp_dec.wasm`: 98K

jakearchibald:
Brilliant! I fed this through the build reporter in #450:

screenshot 2019-02-05 at 15 29 11

jakearchibald:
(that does kinda highlight how we'd benefit from a separate manifest so the changes don't cascade)

PR #440, opened 2019-01-26
Add Travis Build Status badge to Readme

EwoutH:


surma:
Thanks for the PR! We decided against badging at the start of the project because it can make a project look broken or abandoned when it’s not.

Issue #446, opened 2019-02-04
Layout broken in stable Chrome


jakearchibald:
On Chrome mobile:

1. Open any image.
1. Expand mozjpeg options.
1. Expand advanced options.

The settings overflow.

![squoosh app_ pixel 2 xl](https://user-images.githubusercontent.com/93594/52205865-68ba5180-2870-11e9-9cb2-f3b944849a5c.png)

This is due to a recent behaviour change in Chrome. It happens in Firefox too, so my guess is we were previously relying on a browser bug & Chrome fixed itself.

I'm not sure how long it's been broken in Firefox. It might be that I didn't look at the mobile view in Firefox very closely.

Investigating.

PR #447, opened 2019-02-04
Fixing layout issue in #446

jakearchibald:
I'm not 100% sure why this fixes the problem. I guess I don't fully understand the grid & flexbox algorithms. However, I created a [little reduction](https://static-misc.glitch.me/flexbox-issue-reduction/), and this one-line change fixes it.


jakearchibald:
@surma said he was fine with this. He can call me a liar if he wants.

PR #450, opened 2019-02-05
Reporting size diff between builds. Fixes 

jakearchibald:
I've added a little script to log changes in sizes from a reference build:

screenshot 2019-02-05 at 14 23 01

Eg see https://travis-ci.org/GoogleChromeLabs/squoosh/builds/489062288 (expand `npm run sizereport`), which is just a demo branch where I made a couple of small changes.

The previous build info is currently stored in the travis log, but I could move this to a JSON file in the main build. The benefit of using the travis log is we could easily gather historical data. Dunno if that's useful.

It's highlighted something kinda alarming. The hashes of some files are different between Travis, Netlify, and my local machine. For instance, `main-app.[hash].css`:

* Travis: `main-app.ebd69.css`.
* Netlify: `main-app.0effa.css`.
* Local: `main-app.9fcff.css`.

@developit have you seen this kind of thing before? They're all using the same node version. Does it use something OS-level for hashing that might differ from system to system?

The script is currently comparing to the last successful build on size-report. I'll need to change that to live/master before merging.


developit:
This usually happens because there are absolute filepaths in the generated code, which alter the computed `[chunkhash]` ([discussion](https://github.com/webpack/webpack/issues/2215#issuecomment-393854279)).  This is likely due to a plugin, hopefully not one of mine ;)

developit:
@jakearchibald feature request: show the size delta percentage instead of % of previous build ("+4%").
Probably useless but [here's the code for printing that](https://github.com/GoogleChromeLabs/size-plugin/blob/master/src/index.mjs#L125-L140) in size-plugin.

jakearchibald:
@developit no probs. I think I'm the only person in the world that prefers % to be relative to 100.

jakearchibald:
@developit oh, and good catch on the absolute paths thing. That sounds like exactly what's happening.

surma:
I actually prefer 102%/95%. I always struggle with what -9% actually means. But no strong opinion from my side ;)

jakearchibald:
@surma yeah, but we also disagree with everyone on "next Wednesday".

@kosamari any opinion here?

jakearchibald:
@developit any idea how Travis decides which comments to fold away and which to reveal?

config/size-report.js
```diff
@@ -0,0 +1,217 @@
...
+const fetch = require('node-fetch');
+const prettyBytes = require('pretty-bytes');
+const escapeRE = require('escape-string-regexp');
+const chalk = new require('chalk').constructor({ level: 4 });
+
+const readdir = util.promisify(fs.readdir);
+const stat = util.promisify(fs.stat);
+
+function getTravis(path) {
+  return fetch('https://api.travis-ci.org' + path, {
```
surma: wanna use 

```js
const {URL} = require('url');
/...
function getTravis(path) {
  return fetch(new URL(path, 'https://api.travis-ci.org').href, ...)
}
```

Would catch things like forgetting a leading slash and do the percent encoding stuff.

config/size-report.js
```diff
@@ -0,0 +1,217 @@
...
+}
+
+function getTravisText(path) {
+  return getTravis(path).then(r => r.text());
+}
+
+/**
+ * Recursively-read a directory and turn it into an array of { name, size, gzipSize }
+ */
+async function dirToInfoArray(startPath, {
```
surma: Maybe rely on an existing impl here? Like [`readdirp`](https://www.npmjs.com/package/readdirp)?

config/size-report.js
```diff
@@ -0,0 +1,217 @@
...
+  const matchRe = new RegExp(`^${escapeRE(nameParts[1])}[a-f0-9]+${escapeRE(nameParts[2])}$`);
+  const matchingEntry = buildInfo.find(entry => matchRe.test(entry.name));
+  return matchingEntry;
+}
+
+const buildSizePrefix = '=== BUILD SIZES: ';
+const buildSizePrefixRe = new RegExp(`^${escapeRE(buildSizePrefix)}(.+)$`, 'm');
+
+async function getPreviousBuildInfo() {
+  const buildData = await getTravisJson('/repo/GoogleChromeLabs%2Fsquoosh/builds?branch.name=size-report&state=passed&limit=1');
```
surma: I’m guessing switching this to `master` is that “small change”? 😄

config/size-report.js
```diff
@@ -0,0 +1,217 @@
...
+
+  // One letter references, so it's easier to get the spacing right.
+  const y = chalk.yellow;
+  const g = chalk.green;
+  const r = chalk.red;
+
+  for (const change of buildChanges) {
+    // New file.
+    if (!change.beforeSize) {
+      console.log(`  ${g('ADDED')}   ${change.afterName} - ${prettyBytes(change.afterSize)}`);
```
surma: Have you tried `console.group()` to see how Travis handles that?

config/size-report.js
jakearchibald: Ah yes. My implementation is super lazy. I'll fix it up.

config/size-report.js
jakearchibald: Yeah I'll give it a shot.

config/size-report.js
jakearchibald: Yeeep 😄

config/size-report.js
jakearchibald: Travis appears to use markers for folding https://github.com/travis-ci/travis-ci/issues/2285#issuecomment-42724719, but it isn't clear how to unfold.

surma: Maybe shout into the Twitter void?

PR #452, opened 2019-02-07
Add link to intro page to install as a PWA

benfredwells:
Fixes issue 358


benfredwells:
Forgot to mention: this is my first react thing (apart from doing a tutorial many moons ago). Hopefully it isn't too silly :)

surma (changes requested):
Not silly code at all, well done!

I am wondering if it’s worth rolling our own install UI or if we should just yield to the UA. The second we register a listener for the event, we become eligible for the standard install banner. WDYT?

src/components/intro/index.tsx
```diff
@@ -66,6 +70,22 @@ export default class Intro extends Component<Props, State> {
     this.fileInput!.click();
   }
 
+  @bind
+  private onBeforeInstallPrompt(e: Event) {
+    this.setState({ canInstall: true });
+    this.bipEvent = e as BeforeInstallPromptEvent;
```
surma: If you wanna ship your own install UI, you need to `preventDefault()` here.

surma:
screenshot_20190207-145933

src/components/intro/index.tsx
benfredwells: Oops, I had that in onInstallClick. Fixed.

benfredwells:
> I am wondering if it’s worth rolling our own install UI or if we should just yield to the UA. The second we register a listener for the event, we become eligible for the standard install banner. WDYT?

I made this change with desktop in mind. There is no standard install banner there so I think the link is good.

For mobile, it's a bit confusing. The site gets the banner as long as there is no preventDefault called on the event (even if it is listened to). What squoosh should do on mobile is a good question though. Some options are:
- just use the link and hide the small banner everywhere
- have the link and also have the small banner. probably best to only have it on the intro page.
- keep relying on the banner.

Some time ago we (Chrome) moved from having the big banner to the current (smaller) banner, plus allowing sites to have a button if they  wanted. The idea was that we'd eventually remove all banners and just leave this up to sites, as banners are kinda annoying (e.g. they hide part of the page, once hidden are unpredictable) and we got feedback that sites wanted more control over their UI.

benfredwells:
BTW I'm happy to limit this to desktop if the question of what to do on mobile needs time / discussion.

benfredwells:
Some additional info about the mobile banner - once the user hides it, it's gone for 90 days. I hadn't realised it was so long ... an advantage of the site UI is that it can be shown as much as it wants.

benfredwells:
Ping - is there more that I need to do here? interested to hear thoughts about what y'all think the approach on mobile should be.

jakearchibald:
Couple of questions about the API in general:

* If we `preventDefault` on `beforeinstallprompt`, but don't call `prompt` during the life of the page, when will we get the opportunity to call it again?
* If we call `prompt`, but the user says no, when will we next get the `beforeinstallprompt` event?
* When in the page lifecycle is `beforeinstallprompt` called, or is it unpredictable?

Depending on the answers above, we need to think about when & how we'd show the install banner. I think the current solution is difficult to discover, and produces an unstable render (the install link will pop-in and move other elements around).

Maybe we should display a snack for this? But when would we show it? Do we want to provide the user with information about what 'install' means? If the user says 'no', are we doing enough to avoid being spammy?

Another option is to introduce a button in the top-left of the UI. There, it could fade in without displacing other UI.

Once we know roughly the direction we want to do, we'll get some design help.

benfredwells:
> Couple of questions about the API in general:
> 
> * If we `preventDefault` on `beforeinstallprompt`, but don't call `prompt` during the life of the page, when will we get the opportunity to call it again?

During the next page load you'll get another bip event.

> * If we call `prompt`, but the user says no, when will we next get the `beforeinstallprompt` event?

You get another one straight away.

> * When in the page lifecycle is `beforeinstallprompt` called, or is it unpredictable?

It is sometime after load. If the origin doesn't have any engagement the page won't get it until the user accumulates enough. The engagement required is minimal and if a user is active (i.e. clicking or scrolling) takes about ten seconds. If the site has enough engagement the event should come very close to load, after the installability checks have been completed.

Another point worth noting: a gesture is required to call prompt.

> 
> Depending on the answers above, we need to think about when & how we'd show the install banner. I think the current solution is difficult to discover, and produces an unstable render (the install link will pop-in and move other elements around).

Sounds good. I agree it isn't very noticeable as is, something more visible would be awesome (but will need design help). You're right about the pop in ... it's actually a bit worse actually. As implemented now the install link will also pop out as soon as the user clicks it, and then if the user cancels the prompt it will pop back in.

> 
> Maybe we should display a snack for this? But when would we show it? Do we want to provide the user with information about what 'install' means? If the user says 'no', are we doing enough to avoid being spammy?

Good questions ... there is a tradeoff between spamminess and noticeability. The bip event gives sites freedom to decide how they want to trade these things off :)

> 
> Another option is to introduce a button in the top-left of the UI. There, it could fade in without displacing other UI.

This sounds like a good option. You could also combine it with a snack (or something more noticeable) if you really wanted, with the snack highlighting the button. But that's probably overkill.

> 
> Once we know roughly the direction we want to do, we'll get some design help.

PR #458, opened 2019-02-11
Fix buffer offset/size calculations in rotate/processor.ts

jviide:
This pull request modifies how rotate/processor.ts copies the image data to & from the WASM work buffer.

The Rust rotate implementation merged from #438 adds a 4 byte padding to the beginning of the WASM. Currently rotate/processor.ts copies the image data to offset `0` instead of `4`, and reads the output buffer from offset `bytesPerImage` instead of `bytesPerImage + 4`. As a result the rotated images may have some artifacts. In the following example the artifacts are visible along some sides of the rotated versions (e.g. top left corner of the 270° one):

This pull request also makes a minor modification to how many pages are needed for the WASM buffer. Currently `4` gets added to `bytesPerImage * 2 / (64 * 1024)`, while the `4` should probably be added before the division.

### DIFF
--- a/src/codecs/rotate/processor.ts
+++ b/src/codecs/rotate/processor.ts
@@ -11,7 +11,7 @@ export async function rotate(
 
   // Number of wasm memory pages (á 64KiB) needed to store the image twice.
   const bytesPerImage = data.width * data.height * 4;
-  const numPagesNeeded = Math.ceil(bytesPerImage * 2 / (64 * 1024) + 4);
+  const numPagesNeeded = Math.ceil((bytesPerImage * 2 + 4) / (64 * 1024));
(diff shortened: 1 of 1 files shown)


jviide:
I have previously signed the CLA for another project.

surma:
Dayum, I fixed this, but must have not committed or something idiotic. Thanks so much for catching this.

Just gonna wait for CI and Netlify :)

jviide:
Kudos for the WASM build pipeline, it was lovely. Got it up and running without any problems 👍

surma:
> Kudos for the WASM build pipeline, it was lovely. Got it up and running without any problems 👍

That actually means a lot to me! Thank you 😇

PR #462, opened 2019-02-14
Make Rust rotate code smaller

surma:
@fitzgen told me about the [“Shrinking .wasm size” in the Rust wasm book](https://rustwasm.github.io/book/reference/code-size.html) which mentions that you slice access via index (`image[i]`) pulls in formatter code for the potential panic when you `i` is out-of-bounds.

You can instead you accessor methods and handle out-of-bounds yourself to avoid pulling in that formatter code.

The Rust module seems to have the same performance, but is only **364B** big.


codecs/rotate/rotate.rs
```diff
@@ -1,9 +1,19 @@
...
 
 use core::panic::PanicInfo;
 use core::slice::from_raw_parts_mut;
 
+#[inline]
+pub fn unwrap_abort<T>(o: Option<T>) -> T {
+    use core::intrinsics::abort;
+    match o {
+        Some(t) => t,
+        None => unsafe{abort()},
```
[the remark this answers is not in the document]
surma: Good shout!

surma:
Okay, so the magic keyword for Rust is “Link-Time Optimizations”. Those are disabled by default because they make compiling slower but they make the Rust binary size 100% competitive here (because at link time it is obvious that we are not making any allocations and that the allocator can get removed).

So we are now officially back on the happy path for Rust with a competitive file size.

codecs/rotate/rotate.rs
```diff
@@ -69,13 +83,8 @@ fn rotate(input_width: isize, input_height: isize, rotate: isize) {
   for d2 in 0..d2_limit {
     for d1 in 0..d1_limit {
       let in_idx = (d1_start + d1 * d1_advance) * d1_multiplier + (d2_start + d2 * d2_advance) * d2_multiplier;
-      out_b[i as usize] = in_b[in_idx as usize];
+      *unwrap_abort(out_b.get_mut(i as usize)) = *unwrap_abort(in_b.get(in_idx as usize));
```
Dr-Emann: Won't this not compile in debug mode? I think you meant to name the `unchecked_unwrap` function `unwrap_abort` (and also make it non-unsafe).

This also looks pretty ugly. May I recommend an extension trait:
```rust
trait UnwrapHard {
  type Output;
  fn unwrap_hard(self) -> Self::Output;
}

impl<T> UnwrapHard for Option<T> {
    type Output = T;
    #[cfg(not(debug_assertions))]
    #[inline]
    fn unwrap_hard(self) -> T {
        use std::process;
        match self {
            Some(t) => t,
            None => process::abort(),
        }
    }
    
    #[cfg(debug_assertions)]
    fn unwrap_hard(self) -> T {
        self.unwrap()
    }
}
```
To see it in action, see [the playground](https://play.rust-lang.org/?version=stable&mode=release&edition=2018&gist=198d299a0ebeb95c488a1971aed57233)

surma: You are correct. The unsafe/misnaming was an oversight copy/pasting from @fitzgen lol. I’m working on a new PR anyways and we shipped the release build, so we are good for now 😅

surma: Also good shout on the trait. I’m gonna use that in the next PR. Thank you very much!

PR #463, opened 2019-02-15
Use tiling in rotate.rs

jviide:
After the recent [Replacing a hot path in your app's JavaScript with WebAssembly](https://developers.google.com/web/updates/2019/02/hotpath-with-wasm) article the HackerNews [commenters raised a good point](https://news.ycombinator.com/item?id=19167078) about using loop tiling to speed up the rotation code. This pull request implements such tiling for the Rust code.

The main logic is also modified a bit: When the x and y coordinates iterated through they are used as-is to calculate the *input* image's linear index. The `d1_*` and `d2_*` variables are renamed to `x_*` and `y_*` and they now signify how the original image's x and y coordinates should be mapped to linear indexes for the *output* image.

I chose 16x16 tile size, as it seemed to give the best performance on my machine. Your mileage may vary. On Chrome 72 and Firefox 65 (MacOS) 90° and 270° rotations seems to take about half the time compared to the previous implementation (when tested with the "Large photo" example). 180° rotations appear to be unaffected.


surma:
Nice work! I’m gonna hold off on this one because I started working on a tiling implementation as well with a proper benchmarking rig, so I’ll finish that up, compare and see if we can merge our thoughts 😄

jviide:
Closing this in favor of #464. It's better 🙂

PR #464, opened 2019-02-15
Tiling for Rotate

surma:
Here’s my implementation of tiling rotation.

In this PR I did a couple of things:

- I split up implementations of the different rotations. It’s less DRY but made it easier to reason about.
- 90 degree and 270 degree rotation got tiling.
- I added a d8 benchmark script that tests all 4 rotation values with the baseline compiler and the optimizing compiler, takes 100 samples and gives you the stats after.
- I switched to a 8 byte offset to make sure we are memory aligned on 64-bit machines. I’m not actually sure it makes a difference, but paying 4 byte to make sure seems reasonable 🤷‍♂️.

The results are kinda impressive:
screen shot 2019-02-15 at 8 08 35 pm


codecs/rotate/rotate.rs
```diff
@@ -1,90 +1,82 @@
...
+    match rotate {
+        0 => {
+            for i in 0..num_pixels {
+                *out_b.get_mut(i).unwrap_hard() = *in_b.get(i).unwrap_hard();
+            }
+        }
+        90 => {
+            let new_width = height;
+            let _new_height = width;
+            for y_start in (0..height).step_by(TILE_SIZE) {
```
surma: @bmeurer @mathiasbynens The performance benefits are _real_. Is this something that v8 could do automatically?

codecs/rotate/package.json
```diff
@@ -2,6 +2,10 @@
   "name": "rotate",
   "scripts": {
     "build:image": "docker build -t squoosh-rotate .",
-    "build": "docker run --rm -v $(pwd):/src squoosh-rotate ./build.sh"
+    "build": "docker run --rm -v $(pwd):/src squoosh-rotate ./build.sh",
+    "benchmark": "echo File size after gzip && npm run benchmark:filesize && echo Baseline && npm run benchmark:baseline && echo Optimizing && npm run benchmark:optimizing",
+    "benchmark:baseline": "v8 --liftoff --no-wasm-tier-up --no-opt ./benchmark.js",
+    "benchmark:optimizing": "v8 ./benchmark.js",
```
surma: @bmeurer @mathiasbynens Would you say these are good flags to test the baseline wasm compiler vs the optimizing one?

codecs/rotate/rotate.rs
bmeurer: You mean something that Rust could do automatically?

(I'm maybe missing the relevant context here BTW... so please bear with me)

codecs/rotate/package.json
bmeurer: I don't know TBH. @hammacher and/or @gahaas might now.

codecs/rotate/rotate.rs
surma: Yeah sorry to just drop you in like that. We learned today that loops can/should be tiled when iterating over large chunks of memory to make better use of the processors L1 cache. 

The difference in the benchmark shows that this is indeed a massive performance gain. Isn't that something that could be optimized at the level of JS ans Wasm, so that developers don't have to implement the tiling by hand?

codecs/rotate/benchmark.js
```diff
@@ -0,0 +1,45 @@
+// THIS IS NOT A NODE SCRIPT
+// This is a d8 script. Please install jsvu[1] and install v8.
+// Then run `npm run --silent benchmark`.
```
developit: tip: `-s`

codecs/rotate/rotate.rs
bmeurer: What you are looking for is essentially a compiler pass that does LNO (loop nest optimization). This is most effective when there's a certain knowledge of the data structure (and often makes sense as a manual optimization applied by the programmer, who knows the data structure).

I'd assume that the Rust compiler or LLVM would be in a better position here. The WASM compiler would need to do quite a bit of analysis to reconstruct enough information to even do some kind of educated guess whether it's a good idea or not. Also when you put such heavy optimizations into the WASM JIT, you pay a runtime cost.

Maybe talk to the LLVM/Rust folks about it?

surma: Thanks for explaining! I wasn’t necessarily implying that V8 _should_ do this, I figured way smarter than me would have already done so if it was a good idea. But it seems it’s just not that easy to make that decision by looking at the byte code. Thank you again :)

codecs/rotate/benchmark.js
surma: Are we golfing `npm` invocations now? ^^ (Thanks!)

codecs/rotate/benchmark.js
```diff
@@ -0,0 +1,45 @@
...
+    }
+    const average = iterations.reduce((sum, c) => sum + c) / iterations.length;
+    const stddev = Math.sqrt(
+      iterations
+        .map(i => Math.pow(i - average, 2))
+        .reduce((sum, c) => sum + c) / iterations.length
+    );
+    print(`n = ${iterations.length}`);
+    print(`Average: ${average}`);
+    print(`StdDev: ${stddev}`);
```
mathiasbynens: Tip: you could use `console.log` instead of `print`. That way, this entire script becomes runnable in a browser if you “polyfill” `readbuffer` (and maybe prefix `readbuffer` with `await` to make that polyfill not be super ugly).

codecs/rotate/rotate.rs
mathiasbynens: @bmeurer The full context is here: https://news.ycombinator.com/item?id=19167078 So the question is not just about enabling this optimization starting from Wasm bytecode, but also for JS source code. I suppose the following applies to JS as well though:

> it’s just not that easy to make that decision by looking at the byte code.

...or the JavaScript source code.

bmeurer: Yes, it's pretty much orthogonal to the concrete programming language. And the same goes for the JavaScript JIT. Doing this properly requires some knowledge about the concrete data structure used in the application.

codecs/rotate/package.json
hammacher: In order to ensure that you never run baseline wasm code, you should add "--no-liftoff --no-wasm-tier-up" to the optimizing config.

jviide:
I tested it with Chrome 72 (MacOS) and there was something funky going on. I clocked the rotations with `console.time`, and 270° rotations appeared to be 2-3x slower than 90° rotations. Submitted #474 with more details and some suggested modifications.

surma:
90 and 270 degree rotation are now equally fast for me as well.

(For the record, I swear they used to be equal when I opened this PR. But I can reproduce that they are not anymore?!)

I’ll try to create a reduction case for the wasm team.

jviide:
I have signed the CLA (for a different project) and I am ok with my commits being contributed to this project.

surma:
For me tile size 16 is faster. I am gonna stay conservative and go with the lower one. It kinda makes more sense (in my head) that lower values would perform better on average across a bunch of devices 🤷‍♂️ 

<table>
  <thead>
    <tr>
      <th>Rotation</th>
      <th>Tile size = 64</th>
      <th>Tile size = 16</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td>90°</td>
      <td>µ=83.89</td>
      <td>µ=58.49</td>
    </tr>
    <tr>
      <td>270°</td>
      <td>µ=83.66</td>
      <td>µ=59.36</td>
    </tr>
  </tbody>
</table>

PR #468, opened 2019-02-19
Only compare against the last 'push' to master

jakearchibald:
Previously, it was comparing against the last pull request to master. I dunno the travis API is weird.


jakearchibald:
Merge commits count as a push to master.

PR #469, opened 2019-02-19
Add share target

jakearchibald:
I'm happy with this now, although I'm happy to block on #353.


src/lib/sw-bridge.ts
```diff
@@ -40,6 +40,23 @@ async function updateReady(reg: ServiceWorkerRegistration): Promise<void> {
...
+      if (event.data.action !== 'load-image') return;
+      resolve(event.data.file);
+      navigator.serviceWorker.removeEventListener('message', onmessage);
+    };
+
+    navigator.serviceWorker.addEventListener('message', onmessage);
+
+    // This message is picked up by the service worker - it's how it knows we're ready to receive
+    // the file.
+    new BroadcastChannel('share-ready').postMessage('share-ready');
```
surma: What about iOS?

jakearchibald: Ah fuck

jakearchibald: fwiw, `BroadcastChannel` is only used where Share Target V2 is also supported, but I'll make it work using the clients API anyway.

jakearchibald:
@surma finally removed BroadcastChannel. I've tested this, and it works. Wanna give it a review?

src/components/App/index.tsx
```diff
@@ -103,15 +114,18 @@ export default class App extends Component<Props, State> {
...
     return (
       <div id="app" class={style.app}>
         <file-drop accept="image/*" onfiledrop={this.onFileDrop} class={style.drop}>
-          {!isEditorOpen
-            ? <Intro onFile={this.onIntroPickFile} showSnack={this.showSnack} />
-            : (Compress)
-              ? <Compress file={file!} showSnack={this.showSnack} onBack={back} />
-              : <loading-spinner class={style.appLoader}/>
+          {
+            awaitingShareTarget
```
surma: This ternary chain is getting a bit outta hand, don’t you think? Maybe we can put it into some helper functions and a switch? `shouldShowSpinner(state)` `shouldShowIntro(state)` etc?

src/sw/util.ts
```diff
@@ -104,3 +124,26 @@ export async function cacheAdditionalProcessors(cacheName: string, buildAssets:
...
+function nextMessage(dataVal: string): Promise<void> {
+  return new Promise((resolve) => {
+    if (!nextMessageResolveMap.has(dataVal)) {
+      nextMessageResolveMap.set(dataVal, []);
+    }
+    nextMessageResolveMap.get(dataVal)!.push(resolve);
+  });
+}
+
+self.addEventListener('message', (event) => {
```
surma: Curious: Why the indirection via a map? Couldn’t each call to `nextMessage()` just add an additional listener that self-removes when a message with the desired string has been received?

surma:
Also seems like a rebase is needed

src/sw/util.ts
jakearchibald: Top level listeners can only be added during initial execution.

jakearchibald:
Hm, I merged master as part of this. I'll double check.

PR #474, opened 2019-02-20
Remove timing difference between 90° and 270° rotations (in the tiling Rust code)

jviide:
When I tested the tiling rotation code from #464 with Chrome 72 (MacOS) there was something funky going on. Clocking the rotations with `console.time`, 270° rotations appeared to be 2-3x slower than 90° rotations. Based on my earlier PR #463 that this shouldn't be the case, and the code itself seemed fine. I ran the included benchmarks with V8 version 7.4.196, and sure enough, the optimizing benchmark (more or less) agreed with manual testing. For the record, Firefox 65 didn't seem to have this quirk. 

This pull request modifies the code in following ways:
 1. Fix one small bug in unwrap_hard when debug_assertions is set.
 1. Apply some optimizations to the base code, using slice iterators when possible.
 1. Divide the different rotation branches into their own functions, marked #[inline(never)]. This was a result of trial, error and guesswork, and is a separate commit from the other modifications in case there's a better solution.

It seems that both modifications 2 and 3 are required to remove the difference between 90 and 270 degree rotations. The resulting code is a bit faster on all benchmarks (YMMV).

Baseline:

| rotation | μ before | σ before | μ after | σ after | delta |
| -------- | -------- | -------- | ------- | ------- | ----- |
| 0° | 54.74 | 7.93 | 43.95 | 8.51 | -19.71% |
| 90° | 269.98 | 4.63 | 222.43 | 3.61 | -17.61% |
| 180° | 68.91 | 1.00 | 46.83 | 0.67 | -32.04% |
| 270° | 264.82 | 4.55 | 222.04 | 3.00 | -16.15% |

Optimizing:

| rotation | μ before | σ before | μ after | σ after | delta |
| -------- | -------- | -------- | ------- | ------- | ----- |
| 0° | 22.68 | 6.79 | 14.87 | 6.49 | -34.44% |
| 90° | 103.15 | 2.26 | 86.73 | 2.92 | -15.92% |
| 180° | 24.34 | 0.73 | 16.78 | 0.70 | -31.06% |
| 270° | 230.27 | 3.39 | 78.58 | 3.65 | -65.87% |

I don't really know what's up when it comes to the code size. Gzipped size for the modified code is 806 bytes vs. the original 680 bytes. However, the *original* code takes 890 bytes (Gzipped) when built with my dockerified toolchain (built with `yarn build:image`). Maybe we have different versions of the Rust compilers?


surma:
Stellar work. I’m gonna merge this! I am as surprised as you are that these two loops were show such contrasting performance characteristics.

cc @ThomasTheDane and @dschuff

PR #476, opened 2019-02-21
Fix potential over-the-bounds slicing in rotate.rs introduced in #474

jviide:
This pull request fixes a potential out-of-bounds slicing problem introduced by Yours Truly in #474. Unlike in e.g. Python slicing over the bounds fails in Rust. Sorry about that.

The code needed some reworking, as the V8 version 7.4.196 benchmarks (as well as manual timing tests in Chrome 72) showed that just slicing until `y * width + (x_start + TILE_SIZE).min(width)` caused a major performance drop. Firefox 65 didn't seem to have this performance drop.

I also bumped the tile size up to 64x64, as it gave a noticeable performance boost in my test environment. YMMV.


jviide:
Related to #464.

surma:
Hm, I should have caught that! Either way, thanks for following up on this.

I’ll merge this. 

FWIW, I did some benchmarks a while ago where a tile size of 16 was just as fast as a tile size of 64 (across iMac, MacBook Pro and Pixel 3), but I’ll do another test run and make a decision in the other PR :)

PR #494, opened 2019-03-04
Don’t use instantiateStreaming for rotate wasm module

### LINKED ISSUE #493: Rotation change doesn't work on Safari.
Since [WebAssembly.instantiateStreaming](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WebAssembly/instantiateStreaming) method has no support on Safari, changing the rotation of an image on Safari causes an error being thrown.

**Steps to reproduce the behavior:**
1. Open the app in Safari browser.
2. Select an image.
3. Click rotation-change button.

**Versions:**
 - OS version: MacOS 10.14.3
 - Browser version: Safari 12.0.3
 - Node version: 10.15.1
 - npm version: 6.4.1

**Screen-shot:**

screen shot 2019-03-04 at 16 13 05

surma:
Fixes #493 

(H/T @alisaitbilgi)

### DIFF
--- a/src/codecs/rotate/processor.ts
+++ b/src/codecs/rotate/processor.ts
@@ -1,19 +1,29 @@
 import wasmUrl from '../../../codecs/rotate/rotate.wasm';
 import { RotateOptions, RotateModuleInstance } from './processor-meta';
 
-const instancePromise = (WebAssembly as any).instantiateStreaming(fetch(wasmUrl));
+// We are loading a 500B module here. Loading the code to feature-detect
+// `instantiateStreaming` probably takes longer to load than the time we save by
+// using `instantiateStreaming` in the first place. So let’s just use
+// `ArrayBuffer`s here.
+const instancePromise = fetch(wasmUrl)
+  .then(r => r.arrayBuffer())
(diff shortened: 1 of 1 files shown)


alisaitbilgi:
I was thinking to try something like that and not sure to have a performance penalty by feature detecting **instantiateStreaming**:

```
const getInstancePromise = async () => {
  const response = await fetch(wasmUrl);
  const buffer = await response.arrayBuffer();
  const module = await WebAssembly.compile(buffer);

  return WebAssembly.instantiate(module);
};

const instancePromise = ('instantiateStreaming' in WebAssembly)
  ? (WebAssembly as any).instantiateStreaming(fetch(wasmUrl))
  : getInstancePromise();
```

surma:
As I say in the comment in the code, I don’t think it’s worth shipping the detection code considering we are loading a 500B wasm module here.

alisaitbilgi:
hoo, cool I thought it was 500MB :D

surma:
That would _not_ have made it to production. Ever. 😂

PR #498, opened 2019-03-06
Better resize methods

jakearchibald:


codecs/resize/pkg/README.md
```diff
@@ -0,0 +1,52 @@
+# 🦀🕸️ `wasm-pack-template`
```
surma: I should `.gitignore` this one

package.json
```diff
@@ -6,7 +6,7 @@
   "scripts": {
     "start": "webpack-dev-server --host 0.0.0.0 --hot",
     "build": "webpack -p",
-    "lint": "tslint -c tslint.json  -p tsconfig.json -t verbose 'src/**/*.{ts,tsx,js,jsx}'",
```
surma: Is this change intentional?

src/codecs/resize/processor.ts
```diff
@@ -1,49 +1,50 @@
...
-    ({ sx, sy, sw, sh } = getContainOffsets(sw, sh, opts.width, opts.height));
+    const { sx, sy, sw, sh } = getContainOffsets(data.width, data.height, opts.width, opts.height);
+    input = crop(input, Math.round(sx), Math.round(sy), Math.round(sw), Math.round(sh));
   }
 
-  return drawableToImageData(data, {
-    sx, sy, sw, sh,
-    width: opts.width, height: opts.height,
-  });
+  await wasm_bindgen(wasmUrl);
```
surma: Does this cause a fetch/initialization every time? Could we keep the instance around?

package.json
jakearchibald: Yeah, this is what was causing it to lint all the `.d.ts` files. It still lints regular files.

src/codecs/resize/processor.ts
jakearchibald: Good call.

jakearchibald:
Cached the module. I'll have a click around in a few browsers & ship it. Yay for achieving something today!

PR #507, opened 2019-03-07
Implement alpha premultiplication

surma:
I’ll do a benchmark later


surma:
Added a benchmark.

Scaling a 2000x2000 to 1500x1500 using Lanczos

- w/o premulti: 604.71ms
- w/ premulti: 711.94ms (+18%)

So yes, of course a negative performance impact, but a bearable one.

And we now allow users to create clean edges on transparent images:
screen shot 2019-03-07 at 9 51 29 pm


H/T @kornelski, thanks for your help :)

codecs/resize/src/lib.rs
```diff
@@ -36,7 +37,17 @@ pub fn resize(
         3 => Type::Lanczos3,
         _ => panic!("Nope"),
     };
+    let num_input_pixels = input_width * input_height;
     let num_output_pixels = output_width * output_height;
+
+    if premultiply {
+      for i in 0..num_input_pixels {
+        for j in 0..3 {
+          input_image[4 * i + j] = ((input_image[4 * i + j] as f32) * (input_image[4 * i + 3] as f32)/255.0) as u8;
```
kornelski: The int->float->int conversion isn't technically necessary (you can use `u16` instead).

codecs/resize/src/lib.rs
```diff
@@ -48,5 +59,15 @@ pub fn resize(
     let mut output_image = Vec::<u8>::with_capacity(num_output_pixels * 4);
     output_image.resize(num_output_pixels * 4, 0);
     resizer.resize(input_image.as_slice(), output_image.as_mut_slice());
+
+    if premultiply {
+      for i in 0..num_output_pixels {
+        for j in 0..3 {
+          output_image[4 * i + j] = ((output_image[4 * i + j] as f32) * 255.0 / (output_image[4 * i + 3] as f32)) as u8;
```
kornelski: division by 0

codecs/resize/pkg/resize_bg.d.ts
```diff
@@ -3,4 +3,4 @@ export const memory: WebAssembly.Memory;
 export function __wbindgen_global_argument_ptr(): number;
 export function __wbindgen_malloc(a: number): number;
 export function __wbindgen_free(a: number, b: number): void;
-export function resize(a: number, b: number, c: number, d: number, e: number, f: number, g: number, h: number): void;
+export function resize(a: number, b: number, c: number, d: number, e: number, f: number, g: number, h: number, i: number): void;
```
jakearchibald: Any idea why the bool is coming out as 'number'?

src/codecs/resize/options.tsx
```diff
@@ -144,6 +147,15 @@ export default class ResizerOptions extends Component<Props, State> {
...
           }
         </Expander>
+        <label class={style.optionInputFirst}>
+          <Checkbox
+            name="premultiply"
+            checked={options.premultiply}
+            onChange={this.onChange}
+          />
+          Premultiply alpha channel
+        </label>
```
jakearchibald: TODO for me: Only show this when one of the worker resize types is picked.

src/codecs/resize/processor-meta.ts
```diff
@@ -6,6 +6,7 @@ export interface ResizeOptions {
   height: number;
   method: 'vector' | BrowserResizeMethods | WorkerResizeMethods;
   fitMethod: 'stretch' | 'contain';
+  premultiply: boolean;
```
jakearchibald: TODO for me: Move this onto the worker options only

jakearchibald:
I'm happy for us to allow premultiplying with lanczos etc. It's what Safari seems to do by default with it's "high quality" resizer.

I've noticed a weird bug though:

With this:

logo_gmail_lockup_default_2x

Resized to 201 width with premultiplying and lanczos:

logo_gmail_lockup_default_2x 1

Notice the unexpected red & yellow pixel in the top-right of the icon. That seems to be caused by the premultiplying.

jakearchibald:
My TODOs are TODONE.

surma:
kapture 2019-03-08 at 10 31 53

I think those stray pixels _might_ be “technically correct”. The image has some surprising transparency.

codecs/resize/pkg/resize_bg.d.ts
surma: For the `resize_bg.d.ts` file that’s actually correct. Wasm doesn’t have a `bool` type. `resize.d.ts`, the types for the wrapper functions, has `boolean`.

jakearchibald:
Hah! Caused by palette reduction no doubt. Fwiw, Safari's high quality resizer (which seems to be doing Lanczos3 with premultiplying) doesn't produce those pixels, but I'm not particularly bothered.

codecs/resize/src/lib.rs
surma: That was part of the reason why I chose `f32`, because then division by zero is not a panic and gets conveniently converted to `0` when cast to an `u8`.

[Playground example.](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=65e6c0d6621461fd15ac372a7dac8dfb)

surma: I should add a comment :D

PR #510, opened 2019-03-08
Implement sRGB color conversoin

surma:
Operating on a 2000x1500 pixel image, using Lanczos3 as the algorithm on my MBP:

<table>
	<tr>
		<td colspan=2></td>
		<th colspan=2>Premultiplication</th>
	<tr>
	<tr>
		<td colspan=2></td>
		<th>false</th>
		<th>true</th>
	</tr>
	<tr>
		<th rowspan=2>Color space conversion</th>
		<th>false</th>
		<td>594ms</td>
		<td>823ms</td>
	</tr>
	<tr>
		<th>true</th>
		<td>840ms</td>
		<td>938ms</td>
	</tr>
</table>

I also generate the LUT at build time. Sadly, `const` functions (that can be executed at compile time) aren’t powerful enough yet, so I had to resort to good old code generation (see `build.rs`).


jakearchibald:
@Kornel so, is `getImageData` in sRGB? We've added the transform here, but just wanting to check we aren't transforming something already transformed.

Demo: https://deploy-preview-510--squoosh.netlify.com/

src/codecs/resize/options.tsx
```diff
@@ -138,6 +141,17 @@ export default class ResizerOptions extends Component<Props, State> {
...
             : null
           }
+          {isWorkerOptions(options) ?
+            <label class={style.optionInputFirst}>
+              <Checkbox
+                name="colorspace"
+                checked={options.colorspace}
+                onChange={this.onChange}
+              />
+              Convert from sRGB to RGB before resizing
```
jakearchibald: Any objections to just "Linear RGB"?

surma: Nope, that works.

surma:
```suggestion
              Linear RGB
```

surma:
CC for the correct Kornel, aka @kornelski ;)

kornelski:
LGTM

Issue #313, opened 2018-11-17
Write tests


kosamari:
something something... testing is good... something.

alisaitbilgi:
Is there an active test branch? If not, which technologies and tools do you think to use for testing?

PR #563, opened 2019-04-23
Tests!

JoelEinbinder:
I come bearing tests!

New dependencies:
* The `build` directory is served via [`serve-handler`](https://www.npmjs.com/package/serve-handler).
* Puppeteer is used to launch Chromium
* `@pptr/testrunner` is used to set up the basic describe/expect pattern

Right now there are 4 basic tests: The index page has a title, a file can be selected, the app works offline, and it doesn't throw a bunch of errors in the console. 

I am using `@pptr/testrunner` just because I am familiar with setting it up. I can switch to mocha or jest or jasmine or whatever if the maintainers have a preference.

The tests currently treat the app as a black box. The build version of the app compresses and optimizes all of the JavaScript code, so it is tricky to call into it from test code.

#313

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
 language: node_js
 cache: npm
-script: npm run build
+script: npm run build && npm run test
 after_success: npm run sizereport
--- a/README.md
+++ b/README.md
@@ -28,4 +28,15 @@ You can run the development server with:
 npm start
 ```
 
+# Tests
+Tests can be run with:
+``sh
+npm test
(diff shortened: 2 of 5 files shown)


surma:
Thanks for the PR @JoelEinbinder!

Do you need this landed urgently? We are currently quite busy with I/O, so unless you need this ASAP, we’d probably get around to this after I/O.

JoelEinbinder:
No urgency, take your time.

PR #586, opened 2019-05-13
Add path aliases and tidy import declarations

alexnoz:
Hey, guys, thanks for the great app! This PR adds path aliases for `codecs` (alias `@codecs`), `src/codecs` (alias `codecs`) and `src/lib` (alias `lib`) so that we don't have to write these dot-dot-slash chains anymore 😃 I also took the liberty to slightly reorder some import declarations and add spaces between import groups so that in each file imports follow the same order and it's easier to read the code. Also, there're a couple of `tslint` fixes related to missing `key` props.

### DIFF
--- a/src/components/checkbox/index.tsx
+++ b/src/components/checkbox/index.tsx
@@ -1,6 +1,8 @@
 import { h, Component } from 'preact';
+
+import { UncheckedIcon, CheckedIcon } from 'lib/icons';
+
 import * as style from './style.scss';
-import { UncheckedIcon, CheckedIcon } from '../../lib/icons';
 
 interface Props extends JSX.HTMLAttributes {}
 interface State {}
--- a/src/components/compress/index.tsx
+++ b/src/components/compress/index.tsx
@@ -1,36 +1,40 @@
 import { h, Component } from 'preact';
 
-import { bind, Fileish } from '../../lib/initial-util';
-import { blobToImg, drawableToImageData, blobToText } from '../../lib/util';
-import * as style from './style.scss';
-import Output from '../Output';
-import Options from '../Options';
-import ResultCache from './result-cache';
-import * as identity from '../../codecs/identity/encoder-meta';
-import * as optiPNG from '../../codecs/optipng/encoder-meta';
-import * as mozJPEG from '../../codecs/mozjpeg/encoder-meta';
-import * as webP from '../../codecs/webp/encoder-meta';
-import * as browserPNG from '../../codecs/browser-png/encoder-meta';
-import * as browserJPEG from '../../codecs/browser-jpeg/encoder-meta';
-import * as browserWebP from '../../codecs/browser-webp/encoder-meta';
-import * as browserGIF from '../../codecs/browser-gif/encoder-meta';
-import * as browserTIFF from '../../codecs/browser-tiff/encoder-meta';
-import * as browserJP2 from '../../codecs/browser-jp2/encoder-meta';
-import * as browserBMP from '../../codecs/browser-bmp/encoder-meta';
-import * as browserPDF from '../../codecs/browser-pdf/encoder-meta';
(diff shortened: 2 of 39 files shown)


surma (commented):
Hm, I’m personally not opposed to the aliases. But deferring to @jakearchibald on this one.

tsconfig.json
```diff
@@ -11,8 +11,13 @@
     "jsx": "react",
     "jsxFactory": "h",
     "allowJs": false,
-    "baseUrl": "."
-  },
+    "baseUrl": ".",
+    "paths": {
+      "codecs/*": ["src/codecs/*"],
+      "lib/*": ["src/lib/*"],
+      "@codecs/*": ["codecs/*"]
```
jakearchibald: We should come up with something better than `"@codecs"` vs `"codecs"`.

Having a prefix for all our aliases seems good though.

tsconfig.json
```diff
@@ -11,8 +11,13 @@
...
+    }
+	},
```
jakearchibald: Weird indenting.

src/components/compress/index.tsx
```diff
@@ -568,6 +572,7 @@ export default class Compress extends Component<Props, State> {
 
     const options = sides.map((side, index) => (
       <Options
+        key={index}
```
jakearchibald: This seems unrelated to this PR

src/components/checkbox/index.tsx
```diff
@@ -1,6 +1,8 @@
 import { h, Component } from 'preact';
+
+import { UncheckedIcon, CheckedIcon } from 'lib/icons';
+
```
jakearchibald: What's the logic behind the reordering and spacing in the changes?

jakearchibald (changes requested):

src/components/compress/index.tsx
jakearchibald: Ah, just saw you mentioned this in the original comment. @developit does `key` make sense here? (and other places in the PR)

src/components/checkbox/index.tsx
alexnoz: Basically, vendors go first, then anything from `src/lib` and `src/codecs`, then everything else, styles are always at the end, the groups are separated by an empty line. I think it makes code slightly easier to read when you don't have to scan every import declaration to find a certain import 🙂

tsconfig.json
alexnoz: Maybe rename `codecs` to `wasm-codecs` and use `@` as a prefix for all the aliases?

src/components/compress/index.tsx
developit: Index keys are never required and often cause performance degradation. `side` could be used as a key here, though there is no reason to do so.

More info:
https://stackoverflow.com/a/42775667
https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-array-index-key.md

```suggestion
```

src/components/compress/index.tsx
```diff
@@ -583,6 +588,7 @@ export default class Compress extends Component<Props, State> {
 
     const results = sides.map((side, index) => (
       <Results
+        key={index}
```
developit, line 591:
```suggestion
```

tsconfig.json
developit: Using `@` seems contentious here, since `@aaa/bbb` is a valid npm package name and bare specifier. I have seen most folks use `~` as a prefix, since it has no other useful meaning within a project.

developit (commented):
My personal preference is to avoid aliases, since they break tooling by default (even if TS and Webpack can be configured with aliases). Most folks I've talked to who have adopted aliases have done so in order to avoid typing long/nested import paths, or to ease refactoring of those paths - both of which are automatically solved by editors like VSCode (via auto-import and automatic import renaming).
Other than that, I don't really have a ton of useful feedback to give here.

PR #596, opened 2019-05-20
Using travis-size-report

jakearchibald:


jakearchibald:
Going to wait until travis works fine before merging.

PR #624, opened 2019-06-16
Add Hq{2,3,4}x

surma:
Did I procrastinate? Yes I did.
Was it worth it? I sure as hell think so!

Shout-out to @torch2424 for (indirectly) introducing me to this Rust library 🎉 

Screenshot 2019-06-16 at 15 23 14


torch2424:
@surma OMG YESSS!!! :smile: :+1: :rocket: :guitar: 

This is so rad! :smile: Glad to see this in squoosh! :tada: :tada: :tada: 

cc @CryZe for making the awesome package! :smile: :clap:

surma:
Aw shit, that reminds me that i forgot to add a readme with a link to the original repo. Will fix tomorrow.

jakearchibald:
If I load the Squoosh logo and select "pixel art", it seems to complete the operation quickly, but then it pegs the CPU at 300%.

We need to think about the UX here a bit. Right now, if I press "resize" before "pixel art", I won't really see the affect of my changes.

Feels like "pixel art" is a form of resize, but I realise that you might want to do regular resizing on top of pixel art resizing.

Some ideas:

* Make "pixel art" a resize option, which picks 2/3/4x automatically based on the output size, and then bilinear resizing to get to the target size.
* Provide a way to add multiple resizings, and if you pick Hq, it only lets you pick 2/3/4x as an output.

surma:
Here are a couple pictures for your amusement :D

340283-bugs-bunny-crazy-castle-3-game-boy-color-screenshot-bugs-take
gfs_28963_2_15
118548-the-legend-of-zelda-link-s-awakening-game-boy-screenshot-at
118544-the-legend-of-zelda-link-s-awakening-game-boy-screenshot-inside
tetris1

CryZe:
Oh wow, that's pretty neat to see my hqx port being used here. However you need to be careful. The original hqx implementation in C is LGPL licensed. Mine is a 1:1 translation of that C library, so I'm assuming LGPL still applies to mine. That would mean in order for it to be compatible with the Apache license here, you'd need to "link" to it, which I'm assuming in the Rust case would be a separate crate. We could potentially publish my crate as its own crate on crates.io and then just use it here.

surma:
@CryZe Yeah, I still have an TODO on the license check.

FWIW, I think your port is great and I’d welcome a separate publish to crates.io. WDYT? 😄 Great job!

surma:
@jakearchibald I integrated the HQX scaling into the resize section and automatically figure out which factor to apply (and the rest is done via nearest-neighbor as defined by [Wikipedia’s article on HQX](https://en.wikipedia.org/wiki/Hqx).

@CryZe For now I’m linking against your GitHub repo in my `Cargo.toml` :)

surma:
The high CPU utilization is still happening (regardless of input image it seems), and I have no idea how to debug it. All the workers have been terminated according to DevTools. The CPU goes back to normal after a minute or so. And ideas?

torch2424:
@CryZe 

Thanks for chiming in! 😄 

@jakearchibald @surma 

If we are going to add a pixel art mode, just know that there are a ton of fun / cool looking upscaling for pixel art that would be fun to try 😄 : 

https://en.wikipedia.org/wiki/Pixel-art_scaling_algorithms

Also, if you find / create more wasm compitble pixel art upscaling packages, let me know 👀 😂

jakearchibald:
I've added a dropdown with some resize presets. It seems generally useful, so I haven't limited it to Hq.

src/codecs/processor-worker/index.ts
```diff
@@ -31,13 +33,36 @@ async function rotate(
...
+      /* webpackChunkName: "process-hqx" */
+      '../hqx/processor');
+
+    const widthRatio = opts.width / data.width;
+    const heightRatio = opts.height / data.height;
+    const ratio = Math.max(widthRatio, heightRatio);
+    const factor = clamp(Math.floor(ratio), { min: 2, max: 4 }) as 2|3|4;
+    if (ratio < 2) {
+      return data;
+    }
```
jakearchibald: Feels like we should only bail if ratio <= 1. This could be moved up a line. Also we're allowed one-liners in this project, so `if (ratio <=1) return data` ftw.

src/codecs/processor-worker/index.ts
```diff
@@ -31,13 +33,36 @@ async function rotate(
...
 }
 
+async function hqx(
+  data: ImageData, opts: import('../hqx/processor-meta').HqxOptions,
+): Promise<ImageData> {
+  const { hqx } = await import(
+    /* webpackChunkName: "process-hqx" */
+    '../hqx/processor');
+  return hqx(data, opts);
+}
```
jakearchibald: Is this still needed?

src/codecs/resize/processor-meta.ts
```diff
@@ -29,10 +47,21 @@ export interface VectorResizeOptions extends ResizeOptionsCommon {
...
 }
 
+/**
+ * Return whether a set of options are from the HQ<n>X set
+ *
+ * @param opts
+ */
+export function isHqx(opts: ResizeOptions): boolean {
+  return opts.method === 'hqx';
+}
```
jakearchibald: It doesn't quite seem worth breaking this out into its own function, but meh

src/components/compress/index.tsx
```diff
@@ -100,12 +100,22 @@ async function preprocessImage(
 ): Promise<ImageData> {
   let result = source.processed;
 
+  if (isWorkerResizeOptions(preprocessData.resize) && isHqx(preprocessData.resize)) {
+    result = await processor.workerResize(result, preprocessData.resize);
```
jakearchibald: Could this be moved into the `else if (isHqx(preprocessData.resize)) {` block below?

src/components/compress/index.tsx
```diff
@@ -100,12 +100,22 @@ async function preprocessImage(
...
     if (preprocessData.resize.method === 'vector' && source.vectorImage) {
       result = processor.vectorResize(
         source.vectorImage,
         preprocessData.resize,
       );
+    } else if (isHqx(preprocessData.resize)) {
+      // Hqx can only do x2, x3 or x4. That part has already been done above.
+      // If the target size is not a clean x2, x3 or x4, use nearest neighbor
+      // for the remaining scaling.
+      const pixelOpts = { ...preprocessData.resize, method: 'browser-pixelated' };
```
jakearchibald: Let's switch to one of the other resize methods. If we're going to be downscaling, it seems like catmull-rom does well.

Also, we can skip resizing if it's already the target size.

src/components/compress/index.tsx
```diff
@@ -100,12 +100,22 @@ async function preprocessImage(
...
+      console.log('Remaining resize', pixelOpts);
```
jakearchibald: Remove log.

src/codecs/resize/options.tsx
```diff
@@ -126,7 +191,7 @@ export default class ResizerOptions extends Component<Props, State> {
           />
         </label>
         <Expander>
-          {isWorkerOptions(options) ?
+          {(isWorkerOptions(options) && !isHqx(options)) ?
```
jakearchibald: If we're switching to catmull-rom, we can keep these options.

src/codecs/resize/options.tsx
```diff
@@ -95,12 +150,22 @@ export default class ResizerOptions extends Component<Props, State> {
             <option value="mitchell">Mitchell</option>
             <option value="catrom">Catmull-Rom</option>
             <option value="triangle">Triangle (bilinear)</option>
+            <option value="hqx">HQX</option>
```
jakearchibald:
```suggestion
            <option value="hqx">hqx (pixel art)</option>
```

It seems stylized as "hqx" lowercase.

jakearchibald:
@surma can you review my changes? https://github.com/GoogleChromeLabs/squoosh/pull/624/commits/83456b4c523e83277db4c863dd0d5e32dfaff2d9

src/components/compress/index.tsx
jakearchibald: Hah, just saw you came to the same conclusion in terms of catmull-rom.

jakearchibald:
I've tried a few different build flags and haven't found any that work around the Chrome bug. I mostly don't know what I'm doing though.

codecs/hqx/package.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "hqx",
+  "scripts": {
+    "build:image": "docker build -t squoosh-hqx .",
+    "build": "docker run --rm -v $(pwd):/src squoosh-hqx ./build.sh"
```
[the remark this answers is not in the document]
surma: Probably.

surma:
@jakearchibald Your stuff LGTM, I added a link to the crbug.

PTAL at my stuff :D

src/codecs/processor-worker/index.ts
```diff
@@ -31,6 +33,18 @@ async function rotate(
...
 ): Promise<ImageData> {
+  if (isHqx(opts)) {
+    const { hqx } = await import(
+      /* webpackChunkName: "process-hqx" */
+      '../hqx/processor');
+
+    const widthRatio = opts.width / data.width;
+    const heightRatio = opts.height / data.height;
+    const ratio = Math.max(widthRatio, heightRatio);
+    if (ratio < 1) return data;
```
jakearchibald:
```suggestion
    if (ratio <= 1) return data;
```

src/components/compress/index.tsx
```diff
@@ -106,6 +106,13 @@ async function preprocessImage(
         source.vectorImage,
         preprocessData.resize,
       );
+    } else if (isHqx(preprocessData.resize)) {
+      // Hqx can only do x2, x3 or x4.
+      result = await processor.workerResize(result, preprocessData.resize);
+      // If the target size is not a clean x2, x3 or x4, use nearest neighbor
+      // for the remaining scaling.
+      const pixelOpts = { ...preprocessData.resize, method: 'catrom' };
+      result = processor.resize(result, pixelOpts as BrowserResizeOptions);
```
jakearchibald: It's `WorkerResizeOptions` now I think.

binji:
Interesting stuff!

As mentioned in the bug, the hq2x implementation ends up producing a huge 150k byte function to perform the filter. I took a look at the generated wasm binary, and generated a control-flow graph of that function, you can see [the resulting SVG here](https://drive.google.com/open?id=1CGe7uKqOLORkfDKkNJcwfnGQHYxNiiO2). It gives you a pretty good idea of the structure; there is a huge fan-out with very little shared code between the branches.

Looking a bit closer, I notice that this pattern shows up many, many times:

```wasm
local.get $N
i32.ge_u
br_if $L
...
i32.const $M
local.get 5
local.get 4
call 40
unreachable
```

This is almost certainly a bounds-check. I thought rust removed these bounds checks when producing optimized code, so I was a bit surprised to see them here.

I found a version of the (almost) original [C source here](https://github.com/grom358/hqx/blob/master/src/hq2x.c). It uses macros to generate many of the various pixel operations, but otherwise is completely unrolled by hand. As an experiment, I took this version and compiled it with `-O3` to wasm. It produces a function that is still very large (120k), but doesn't have any of the extra bounds checks. You can see the [generated control-flow graph here](https://drive.google.com/open?id=1xckSP2jcUIyeYmst4OZqTDStjRm56h_Z). 

It's worth noting that hq2x has been around for a while, so folks have tried to improve the code over time. In particular, I found [this article](http://blog.pkh.me/p/19-butchering-hqx-scaling-filters.html) by Clément Bœsch describing the FFMPEG implementation where the author reduces the amount of code from 12k lines to 560 lines for hq{2,3,4}x. You can find [the code here](http://git.videolan.org/?p=ffmpeg.git;a=blob;f=libavfilter/vf_hqx.c;hb=HEAD).

There's also an [associated HN thread here](https://news.ycombinator.com/item?id=7925671). In it, [byuu describes their version](https://news.ycombinator.com/item?id=7925820) that is much shorter as well (188 lines). See [the pastebin here](http://pastebin.com/raw/DsNupdbc).

This is all to say that the current implementation can almost certainly be improved (while the Chrome bug is being fixed :smiley:)

CryZe:
Rust always does bounds checks as they are necessary for safety. However LLVM can remove them if it can prove that they are not out of bounds. However I could also just remove the bounds checks in the rust code by switching to unsafe code, which was not necessary for the wasmboy use case I was using it with initially, so I never tried optimizing its performance, as it was always fast enough.

mathiasbynens:
#629 works around the issue by using `wasm-opt`.

PR #626, opened 2019-06-17
Display filename on desktop.

pewstiepoll:
This is an attempt to resolve: https://github.com/GoogleChromeLabs/squoosh/issues/244
The changes add filename into UI for desktop.

1. Display filename in the top right corner:
regular_name_usecase

2. If the filename is longer than 300px it's truncated with ellipses:
long_name_usecase

3. If control buttons collapsed (screen_width < 860px) filename is hidden to prevent overlapping:
controls_collapsed_usecase

@jakearchibald @surma @kosamari  please review =)

P.S. I'm not much of a designer, so I didn't figure out where to put filename on mobiles.


jakearchibald:
Design is the hardest part here 😢.

However, we already have a desktop-only solution (the tab title).

PR #672, opened 2019-08-05
Use native Wasm+Webpack support for Rust codecs

RReverser:
This delegates loading of Wasm modules to Webpack itself, making wrapper code simpler.

Emscripten-generated modules are still using custom loading glue as they're not compatible with Webpack.


RReverser:
Ugh, more files that worked locally but ignored by Git. Fixing...

Issue #644, opened 2019-07-01
Use favicon to indicate when work is happening/is done


Rich-Harris:
**Is your feature request related to a problem? Please describe.**

When dealing with large images, it can take a little while for things to happen. I can either stare at the tab while it's crunching, or switch over to a different tab with a plan to return a bit later, which I will often forget to do for minutes at a time.

**Describe the solution you'd like**

It'd be nice if the favicon changed to something that indicated that it was working, then later changed to something indicating it was done (a checkmark?) so i can goof around on twitter but still see when it's finished.

**Does other service/app have this feature?**

None spring to mind

jakearchibald:
Would it be enough to prepend ⏳ to the document title? That would also work for Safari.

Rich-Harris:
Oh, Safari, you really are the runt of the litter.

Changing the title definitely gets us most of the way there. Personally I tend to have 30+ tabs open so that I can't see titles, only favicons, but that's on me. I'd say that's a decent plan.

Rich-Harris:
(would be happy to try and rustle up a PR given a nudge in the right direction, btw)

surma:
Go for it! I think you’d go about it by changing `document.title` in our (pretty huge) `<Compress>` component in the area where we handle the compression jobs and workers.

Let me know if you need more details than that :)

jakearchibald:
https://github.com/GoogleChromeLabs/squoosh/blob/master/src/components/compress/index.tsx#L55

Unfortunately we currently track loading in three places. One for each side, and another overall "loading" for when we're loading/decoding/processing the image.

I guess you'll want to check all three of these in one of the post-update callbacks, and include the ⏳ in the title if any are true.

Avi98:
Hi, is it still open can I work on this?

PR #680, opened 2019-08-14
Use document.title and ⏳ emoji to indicate when work is being done 

WilhelmWillie:
Solution for #644 

Listens to changes in the loading status within the `Compress` component. If a change in loading status is detected and something is loading, append a ⏳ emoji to the `document.title`

*GIFs*

Classic example w/ the famous Lenna  test image. See that the `document.title` updates as the image compresses 
classic_lenna_example

Another example w/ the provided `photo.jpg`. Notice that at a point, both sides are loading/doing work (apologies, had to resize GIF to meet Github's size requirements and Squoosh doesn't support animations 😉) 

another_photo_example

---

Starting  to get into open source and thought this was a neat issue. Also think it would be neat to explore using the `document.title` to output an error emoji in the title like `🚨`

### DIFF
--- a/src/components/compress/index.tsx
+++ b/src/components/compress/index.tsx
@@ -295,9 +295,13 @@ export default class Compress extends Component<Props, State> {
   private updateLoadingIndicator(isLoading: boolean = false): void {
     // Update the title if the editor loading state changes
     if (isLoading) {
-      document.title = loadingIndicator + document.title;
-    } else if (document.title.startsWith(loadingIndicator)){
-      document.title = document.title.slice(loadingIndicator.length);
+      if (!document.title.startsWith(loadingIndicator)) {
+        document.title = loadingIndicator + document.title;
+      }
+    } else {
+      if (document.title.startsWith(loadingIndicator)) {
+        document.title = document.title.slice(loadingIndicator.length);
+      }
     }
   }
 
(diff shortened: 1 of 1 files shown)


WilhelmWillie:
@googlebot I signed it!

src/components/compress/index.tsx
```diff
@@ -289,6 +292,19 @@ export default class Compress extends Component<Props, State> {
     document.title = filename ? `${filename} - ${originalDocumentTitle}` : originalDocumentTitle;
   }
 
+  private updateLoadingIndicator(isLoading: boolean = false): void {
+    // Update the title if the editor loading state changes
+    if (isLoading) {
+      if (!document.title.startsWith(loadingIndicator)) {
```
jakearchibald: This feels a bit fragile. Eg, if the `filename` starts with "⏳ " then this code is going to think there's a loading indicator.

Ideally we should avoid relying on reading state from `document.title`, and instead generate it from app state.

jakearchibald (changes requested):
Thanks for digging into this! Just one change needed I think.

surma:
This has ben reviewed, but the review has not been addressed. @WilhelmWillie, do you still want to work on this or should we close it out?

WilhelmWillie:
@surma Hi! I apologize, got busy over the past year and never got around back to this. I'm still down to work on this, I can try over the weekend and address those previous comments

PR #689, opened 2019-08-28
Swap OptiPNG with OxiPNG

RReverser:
This makes building simpler and allows us to potentially use multithreading version in the future.

For now points to a custom fork of OxiPNG that enables WebAssembly support, as PR is still pending review.


surma:
> docker: Error response from daemon: OCI runtime create failed: container_linux.go:345: starting container process caused "exec: "./build.sh": permission denied": unknown.

This should be fixed by `chmod +x build.sh`.

RReverser:
> It's bigger in file size.

Yes, I mentioned this concern earlier myself in the chat, but got a green light since regression isn't that big after Gzip / Brotli given the other wins. But let me know if that changed :)

jakearchibald:
Yeah, I listed the three pros/cons for comparison. I think the increase in code size vs the performance gain is an acceptable trade-off. My main concern is the regression in squooshing.

surma:
> ✅ It feels much faster than optipng.
> ❌ It's bigger in file size.
> ❌ It produces files that are 2%-4% bigger than optipng.

Interesting. I find the +40KiB gzip’d a bigger drawback than minor file size increase (I wasn’t actually able to reproduce that big of a difference in my tests, but the files of OxiPNG where ever so slightly larger, indeed).

I tried running `wasm-opt` to reduce the wasm module size, but couldn’t get anywhere close to our original size. 

I took a look at an non-stripped build and saw the following:
<details>
<pre>
Shallow Bytes │ Shallow % │ Item
───────────────┼───────────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
─────────────────────────────────────────────────────────────────────────────────────────────────────
         51857 ┊    12.40% ┊ data[0]
         37601 ┊     8.99% ┊ inflate::InflateStream::next_state::hde6ea2e6bb66aa66
         36533 ┊     8.74% ┊ oxipng::optimize_png::h2f1cce12b50d3670
         30260 ┊     7.24% ┊ "function names" subsection
         10723 ┊     2.56% ┊ png::decoder::ReadDecoder<R>::decode_next::h4d05d55b5a36242c
         10368 ┊     2.48% ┊ oxipng::deflate::inflate::h7d04b9aed3b501cf
          9415 ┊     2.25% ┊ oxipng::reduction::alpha::try_alpha_reductions::h4f83721926eaf4df
          9244 ┊     2.21% ┊ optimise
          8111 ┊     1.94% ┊ png::decoder::Reader<R>::next_interlaced_row::hbef92f8f6502258f
          7018 ┊     1.68% ┊ oxipng::deflate::miniz_stream::compress_to_vec_oxipng::h3c5a057c5a81ca0c
          6963 ┊     1.67% ┊ core::fmt::float::float_to_decimal_common_shortest::h6d9589e2e50d285b
          6258 ┊     1.50% ┊ image::dynimage::load::h14e6fa54b94e5977
          6145 ┊     1.47% ┊ core::fmt::float::float_to_decimal_common_exact::h595ea76df31e010d
          5770 ┊     1.38% ┊ oxipng::optimize_png::{{closure}}::h8e8641e07ecd1ffa
          5665 ┊     1.35% ┊ zopfli::squeeze::lz77_optimal::h7705bef4b0ca1a4d
          5661 ┊     1.35% ┊ zopfli::deflate::add_lz77_block::hd915899813893e19
          4778 ┊     1.14% ┊ core::num::flt2dec::strategy::dragon::mul_pow10::h09d7abc17292b664
          4341 ┊     1.04% ┊ zopfli::katajainen::length_limited_code_lengths::ha69507b1f314a765
          3865 ┊     0.92% ┊ zopfli::blocksplitter::blocksplit_lz77::h223a965f0972495e
          3338 ┊     0.80% ┊ miniz_oxide::deflate::core::compress_block::hfd07e5ef93a17d48
          3317 ┊     0.79% ┊ dlmalloc::dlmalloc::Dlmalloc::malloc::haa66310b00a7bd6b
          3221 ┊     0.77% ┊ miniz_oxide::deflate::core::HuffmanOxide::optimize_table::h07fde2a05b3efea4
</pre>
</details>

- It is using `dlmalloc`, not `wee_alloc`
- There is _a lot_ of data that looks like assertions. I wonder if we can remove that? Not sure where it’s coming from tbh.
- The formatter code is still pulled in, which [we might be able to avoid](https://rustwasm.github.io/docs/book/reference/code-size.html#avoid-panicking).

@RReverser wdyt? Wanna take a stab at this?

RReverser:
@surma Are you taking JS size into account? If you remember from chat, when comparing, I was also initially taken aback by the increased Wasm size, but then realised that the gzipped JS wrapper got so much smaller than the total size has actually *decreased*.

RReverser:
(also, 40 KB gzipped doesn't sound right, I had ~16 KB difference, but would need to recheck)

surma:
I did not. You are correct, that makes a different. It’s not a net gain tho. JS size is -27KiB, Wasm size is +45KiB, so +18KiB in total. Which is definitely much less of a big deal.

(That being said, switching to `wee_alloc` should remove another few KiB.)

Screenshot 2019-09-03 at 17 05 28

As I said, I don’t think the slightly bigger files by OxiPNG are that big a deal, but I see where Jake is coming from. I am leaning towards merging this.

RReverser:
Just checked, switching to `wee_alloc` saves ~1.5 KB (<1%) when gzipped. Given the potential performance implications (it's usually somewhat less efficient, and OxiPNG does a lot of allocations), I don't think it's worth it and would rather keep the default allocator.

RReverser:
As for sizes - not much I can do here, but just out of curiosity - are you comparing both codecs with maximum levels? I'm not sure the levels map 1:1 (at the very least, OptiPNG had 8 levels and OxiPNG has 7), so comparing the same ones might be not very useful.

jakearchibald:
I don't think the size of the codec is such a big deal. As @RReverser says, it's a minor increase, and we wouldn't really think twice about throwing another 100k codec in there.

> As I said, I don’t think the slightly bigger files by OxiPNG are that big a deal, but I see where Jake is coming from. I am leaning towards merging this.

But, the whole point of Squoosh is to make images smaller, and this PR regresses on that feature. At maximum effort:

Squoosh logo: 33.9k vs 34.3k (1.2% bigger)

https://upload.wikimedia.org/wikipedia/commons/thumb/1/1a/Su25-kompo-vers2.svg/2880px-Su25-kompo-vers2.svg.png 472k vs 573k (21% bigger)

https://upload.wikimedia.org/wikipedia/commons/thumb/2/2f/Google_2015_logo.svg/1000px-Google_2015_logo.svg.png No difference.

The above image, but reduced to 53 colours: 10.3k vs 11k (6.8% bigger).

https://upload.wikimedia.org/wikipedia/commons/thumb/3/3b/Android_new_logo_2019.svg/1000px-Android_new_logo_2019.svg.png also reduced to 53 colours. 7.18k vs 7.66k (6.7% bigger).

https://upload.wikimedia.org/wikipedia/en/thumb/a/a4/Flag_of_the_United_States.svg/1024px-Flag_of_the_United_States.svg.png also reduced to 53 colours. 12.2k vs 12.7k (4% bigger).

https://upload.wikimedia.org/wikipedia/en/thumb/a/ae/Flag_of_the_United_Kingdom.svg/1024px-Flag_of_the_United_Kingdom.svg.png reduced to 53 colours. 3.48k vs 4.1k (17.8% bigger).

I'm yet to find an example where Oxi performs better, and it seems to perform significantly worse on paletted images.

I guess my problem is, if we land this, what do we announce. New version of Squoosh! Now Squooshes things worse than before! 😄

surma:
Yeah I hear ya.

On that note: cc @shssoichiro and @kornelski. Are these results useful / interesting for you?

Considering both are CLIs potentially used in the wild — what if we just offered both to allow people to use the tools with Squoosh as the GUI 🤔

jakearchibald:
@surma 

> Considering both are CLIs potentially used in the wild — what if we just offered both to allow people to use the tools with Squoosh as the GUI

Yeah, that's what I suggested in https://github.com/GoogleChromeLabs/squoosh/pull/689#pullrequestreview-283059578.

I think it'd be a solid 'yes' if Oxi was sometimes producing smaller files. I'd like to get to a point where we ship multiple crushers, run them in parallel, and pick the smallest file (like ImageOptim). But, if Oxi never produces smaller files, isn't it just a waste of bandwidth and CPU?

surma:
> Yeah, that's what I suggested in #689 (review).

Apologies. I didn’t see that. I think we should go with that for now.

> I think it'd be a solid 'yes' if Oxi was sometimes producing smaller files. I'd like to get to a point where we ship multiple crushers, run them in parallel, and pick the smallest file (like ImageOptim). But, if Oxi never produces smaller files, isn't it just a waste of bandwidth and CPU?

In its current incarnation, that would be a waste, yes. 

Being able to select OxiPNG for faster and slightly suboptimal crushing might be nice, especially for mobile, though.

jakearchibald:
I wonder if we're holding it wrong. Using ImageOptim, in "extreme" mode:

Squoosh logo:
Our optipng: 33.9k
Our oxipng: 34.3k
ImageOptim oxi: 33.9k

https://upload.wikimedia.org/wikipedia/commons/thumb/1/1a/Su25-kompo-vers2.svg/2880px-Su25-kompo-vers2.svg.png
Our opti: 472k
Our oxi: 573k
ImageOptim oxi: 574k

https://upload.wikimedia.org/wikipedia/commons/thumb/2/2f/Google_2015_logo.svg/1000px-Google_2015_logo.svg.png reduced to 53 colours
Our opti: 10.3k
Our oxi: 11k
ImageOptim oxi: 10.3k

https://upload.wikimedia.org/wikipedia/en/thumb/a/a4/Flag_of_the_United_States.svg/1024px-Flag_of_the_United_States.svg.png also reduced to 53 colours
Our opti: 12.2k
Our oxi: 12.7k
ImageOptim oxi: 12.2k.

https://upload.wikimedia.org/wikipedia/en/thumb/a/ae/Flag_of_the_United_Kingdom.svg/1024px-Flag_of_the_United_Kingdom.svg.png reduced to 53 colours.
Our opti: 3.48k
Our oxi: 4.1k
ImageOptim oxi: 3.5k

ImageOptim's version/integration of oxi is better than our's. Although there's the odd output size regression, the only major regression in that large airplane image.

With those numbers, the performance improvement is worth the very slight file size regression.

RReverser:
@jakearchibald It could be that ImageOptim is just stripping headers too (and maybe other post-processing stuff). OxiPNG has an option for that - [`Options::strip`](https://docs.rs/oxipng/2.2.2/oxipng/struct.Options.html#structfield.strip), but I didn't enable it because we didn't have it enabled in OptiPNG either.

Do you want to try to enable it and see if results look better on these test files? (If not, I can try to look at it later.)

RReverser:
Oh, also you can choose Zopfli for compression in OxiPNG - https://docs.rs/oxipng/2.2.2/oxipng/enum.Deflaters.html - but that would make it slower, maybe even as slow as OptiPNG. I guess we could expose these as options.

jakearchibald:
I'm pretty sure our OptiPNG is using Zopfli - @surma is that right?

But yeah, we should enable the options that make the file smallest.

I'm not sure header stripping will help us, but I'll take a look at which headers the output files have.

jakearchibald:
I'm not seeing any superfluous headers in Squoosh's oxi output.

RReverser:
> I'm pretty sure our OptiPNG is using Zopfli

Huh, I didn't hear that, but I guess it could explain both speed (good) and size (bad) differences.

jakearchibald:
Judging by https://github.com/ImageOptim/ImageOptim/blob/master/imageoptim/Backend/Workers/OxiPngWorker.m, it looks like ImageOptim uses these args:

`-o` followed by 0-6 for the level.
`-i0` to prevent interlacing.
`-a` to zero any transparent pixels.
`--strip=safe`.

Are we doing something different? Since we're getting the pixels from a canvas, I don't think header stripping will benefit us, but maybe we're missing something else?

It doesn't look like ImageOptim is using oxi's zopfli (it uses zopflipng as an additional pass, which I had disabled in my tests above).

Feels like we should figure out what we're missing vs imageoptim, then see if it's worth adding zopfli as an option.

In fact, there are lots of interesting options in https://github.com/shssoichiro/oxipng/blob/master/src/main.rs. If you expose them in the wasm, I'll happily do the UI work in Squoosh.

kornelski:
I don't recommend using Zopfli from OxiPNG. It's really slow, and due to slowness it makes OxiPNG dumber, because it checks fewer combinations.

OxiPNG native uses Cloudflare's zlib fork, which has pretty good performance/speed ratio. OxiPNG compiled for WASM will fall back to whatever else is out there that compiles. This is probably the biggest difference in compression quality.

jakearchibald:
@kornelski ohh, great feedback!

@RReverser do you think it's possible to use the cloudflare fork in our wasm version?

kornelski:
It's not possible to use Cloudflare's fork, because the fork is for 64-bit CPUs with SIMD, and WASM currently has neither.

jakearchibald:
[long fart noise]

Well that sucks.

RReverser:
I mean... We could use this opportunity to test native CompressStream when it comes, but not much ideas for something right now. 

Although I suppose OptiPNG uses the same zlib on Wasm as OxiPNG, so it doesn't explain the size differences?

jakearchibald:
> Although I suppose OptiPNG uses the same zlib on Wasm as OxiPNG, so it doesn't explain the size differences?

That's a very good point. Or if it *is* using a different zlib implementation, it's one we would be able to use.

What about the other options like zeroing transparent pixels? Any of those we're missing? Although, some of the images I tested didn't have any transparency, so I guess it isn't that.

RReverser:
@jakearchibald I've tried to play with same options, but still getting same size results. Either ImageOptim is doing some extra post-processing to reduce size further, or it's a regression in OxiPNG itself (between different versions) or... idk, I'm not sure what else is different.

RReverser:
@jakearchibald It could be as well that OptiPNG is simply providing better compression [on these assets] and the difference between ImageOptim+OxiPNG and ours is indeed just zlib itself...

jakearchibald:
Yeah, could be. I thought it might be a simple option because ImageOptim's oxi is producing sizes almost identical to our current imageoptim.

Possible next step:

* Find a better zlib to plug into OxiPNG (as @kornelski suggests)
* Or wasm https://github.com/google/zopfli/blob/master/README.zopflipng, and we'll run oxi + zopflipng in sequence.

Do you have time to take that on? I'm happy to do the Squoosh integration if you can do the wasm side.

RReverser:
> Or wasm google/zopfli:README.zopflipng@master , and we'll run oxi + zopflipng in sequence.

As mentioned above, Zopfli is already directly supported in OxiPNG, I'm not sure what's the point is in compiling ZopfliPNG instead.

For what it's worth, I just tried using Zopfli on the UK flag with 53 colours with the default effort (`2`), and it took 1 minute 25 seconds and produced 3.59KB file.

I'm sure it would beat OptiPNG on a higher level, but I don't have the patience to wait for it to finish and not sure whether it's worth adding even as an option given how slow it is. (At least not until we have multi-thread support.)

kornelski:
@fhanau has looked at faster Zopfli compression and improved heuristics in OxiPNG, so he may be able to help here.

fhanau:
I don't think there's a perfect solution here.

My efforts to improve zopfli resulted in [ECT](https://github.com/fhanau/Efficient-Compression-Tool), but it is probably not stable enough to be used here. Using stock zlib with oxipng would work, that should eliminate the file size regression but eat up a large chunk of the performance improvements. [Libdeflate](https://github.com/ebiggers/libdeflate) also performs very well, but I am not sure if it is as stable as zlib.

jakearchibald:
The nice thing about using oxi is it prepares us for multi-threading as that stabilises in browsers and the rust-to-wasm toolchain.

So, even if it's only a slight improvement performance wise, it feels like a step in the right direction.

jakearchibald:
I don't think we should land this as-is, as it results in a significant regression in Squooshing. I'd happily land something that was close (like we see with Imageoptim oxi) or better.

RReverser:
Okay, I guess I misunderstood the "it feels like a step in the right direction" comment.

jakearchibald:
Ah sorry, that was in response to https://github.com/GoogleChromeLabs/squoosh/pull/689#issuecomment-529264610.

*If we eliminated the file size regression*, but the performance ended up around the same as optipng, that'd still be a step in the right direction.

RReverser:
@jakearchibald Fair enough. I don't think getting same optimisation is a quick change on OxiPNG side, so I guess let's provide these codecs alongside each other.

jakearchibald:
@surma what do you think here?

If we had a few PNG crushers, I wouldn't have a separate item in the dropdown for each. I'd have one "PNG" option, which had a "crush" option, which tried all crushers in parallel (some in series where there's benefit) to produce the smallest file. There might be an advanced option where you can dig into the settings of each crusher.

My worry about this oxi implementation is it's never better than opti when it comes to squooshing. Although it's faster at doing the squooshing, the result is always worse. We have some of that already with the "browser jpeg" encoder, but that adds very little to the app size, whereas oxi is pretty heavy.

RReverser:
> which tried all crushers in parallel

I don't really see the point in it, because, as you said, OxiPNG is worse in terms of size anyway, so this would just make the overall squooshing smaller.

> whereas oxi is pretty heavy

Isn't it lazy-loaded (as in, shouldn't affect app size unless explicitly chosen)?

---

My suggestion was mostly in line with

> Considering both are CLIs potentially used in the wild — what if we just offered both to allow people to use the tools with Squoosh as the GUI 🤔

so that users could explicitly try different tools and compare trade-offs.

RReverser:
Update: I've found a promising 3rd-party DEFLATE library with usually even better and faster compression than zlib / cloudflare-zlib - https://github.com/ebiggers/libdeflate - and decided to integrate it into OxiPNG.

You can find the detailed results, distribution histograms and a spreadsheet with raw numbers here: https://github.com/shssoichiro/oxipng/pull/203.

The TL;DR is that while the results are mixed, in majority of cases it outperforms zlib both in terms of compression ration and in speed, although for some cases it can be slightly worse.

Now that the upstream part of this integration has been merged into OxiPNG, I've updaded this PR to use `libdeflate` as well, and you can play with it on Netlify preview [here](https://deploy-preview-689--squoosh.netlify.com/).

I haven't added threads support to the PR, but it's something we can iterate on separately. For now I'd like all of us to agree on next steps whether this should replace OptiPNG completely, or be a separate codec in addition to it (cc @surma @jakearchibald as you had different opinions on this when we chatted about this 1:1).

jakearchibald:
I can't remember what my previous view was, so I'm sorry if this is a contradiction:

If the new oxi is better in almost all cases, I'm happy dropping optipng. If opti still wins in 20% of cases (kinda making the number up) when we should keep both.

However, they should both exist as one 'Optimised PNG' output type. We'd just run both crunchers, and pick the best result.

RReverser:
Looking at the total size spreadsheet I've made based on tests cases in OxiPNG repo.

If I exclude some edge cases (e.g. anything with abs(diff) < 20 bytes), Opti wins in 11 out of 178 cases ~= 6%.

The size wins in these cases are between 1%-3%.

I don't know how representative that test suite of images is, as many of these might be edge cases for testing, but overall the OptiPNG wins don't seem significant enough.

---

As another data point, the "device screen" Squoosh demo image we have on the main page gets compressed to 1.49 MB with OptiPNG at max level and 1.42 MB with OxiPNG + Libdeflate at max level now.

---

As some more data points, new results for images from above:

> https://upload.wikimedia.org/wikipedia/commons/thumb/2/2f/Google_2015_logo.svg/1000px-Google_2015_logo.svg.png No difference.

27.4KB -> 27.8KB (now 1.4% bigger)

> The above image, but reduced to 53 colours: 10.3k vs 11k (6.8% bigger).

10.3KB -> 9.37KB (now 9% smaller)

> https://upload.wikimedia.org/wikipedia/commons/thumb/3/3b/Android_new_logo_2019.svg/1000px-Android_new_logo_2019.svg.png also reduced to 53 colours. 7.18k vs 7.66k (6.7% bigger).

7.22KB -> 6.75KB (now 6.5% smaller)

> https://upload.wikimedia.org/wikipedia/en/thumb/a/a4/Flag_of_the_United_States.svg/1024px-Flag_of_the_United_States.svg.png also reduced to 53 colours. 12.2k vs 12.7k (4% bigger).

12.3KB -> 10.9KB (now 11.4% smaller)

> https://upload.wikimedia.org/wikipedia/en/thumb/a/ae/Flag_of_the_United_Kingdom.svg/1024px-Flag_of_the_United_Kingdom.svg.png reduced to 53 colours. 3.48k vs 4.1k (17.8% bigger).

3.48KB -> 3.72KB (now 6.9% bigger)

---

All in all, this confirms the trend that there are some regressions, but IMO they don't warrant the complexity of keeping and running both (especially given the way we have to wrap OptiPNG CLI).

It's also worth keeping in mind that OxiPNG is more actively developed and easier to contribute to to improve optimisations further - e.g. I still could and probably will try to add regular zlib support to Wasm version, and then we could run it using both zlib and libdeflate in parallel internally without running two completely different optimizers.

If you still think it's still not good enough yet, we can keep this PR open until such further optimisations happen, or I can contribute them later as separate PRs, but I just don't want it to get blocked forever.

surma:
All in all, with this data, I vote for dropping OptiPNG (which has always been a hacky CLI port by me anyway) and shipping OxiPNG instead.

package.json
```diff
@@ -44,7 +44,7 @@
     "mini-css-extract-plugin": "0.8.0",
     "minimatch": "3.0.4",
     "node-fetch": "2.6.0",
-    "node-sass": "4.12.0",
+    "node-sass": "^4.13.0",
```
nucliweb, line 47: I think that the convention is to use the fixed version for the dependencies

RReverser: We use package-lock anyway, so it doesn't matter much on practice (and `^` would make updates a bit easier), but given that all other packages are specified without `^`, I can change back to match it.

nucliweb: As you want, I only saw something that caught my attention ... it is a small detail without importance.

Btw, nice job 😊

RReverser:
Some timing comparisons using #743, both encoders at highest levels for files above:

| Input | OptiPNG | OxiPNG | Faster by |
| ---- | ---------- | --------- | ---------- |
|Device screen | 76909ms | 6991ms | 11x |
|Google logo, 53 colours | 1975ms | 832ms | 2.3x |
|Android logo, 53 colours | 1191ms | 412ms | 2.9x |
|US flag, 53 colours| 2841ms | 963ms | 3x |
|UK flag, 53 colours| 2438ms | 1028ms | 2.4x |

Note that Libdeflate doesn't support same fine-tuned compression params as Zlib does and instead simply does its best job for the given compression level.

What this means is that while performance of the libraries plays some role at lower levels, most of the improvements at the highest levels come from the simple fact that my OxiPNG + Libdeflate integration ends up trying far fewer iterations than OptiPNG.

Also, OptiPNG has a potential of using Wasm threads which will improve performance even further, although even now getting a slightly smaller result *and* 11x faster on images like the first one seems like a huge improvement.

surma:
Boom.

PR #698, opened 2019-09-15
introduce ESLint

43081j:
hello 👋 

TSLint was essentially deprecated some time back and was made redundant long before that as ESLint has been able to support typescript for quite a while now.

So here i've added ESLint and done the first run through of lint fixes.

There's still a couple hundred lint errors that have to be manually fixed so ill wait and see if you're interested in this change before I go do those. 

Interesting stuff I noticed:
* Your dependencies are all in devDependencies (even things like preact), was this on purpose?
* You use the airbnb guide, would you rather use the google one?
* `noUnusedParameters` isn't turned on in tsconfig, it would be nice if it was
* Stylistic rules are always a bit flaky, it may be worth trying out prettier or clang-format at some point and turning eslint's stylistic rules off

cc @surma


surma:
Hey @43081j, thanks for your PR!

> Your dependencies are all in devDependencies (even things like preact), was this on purpose?

Yes 

> You use the airbnb guide, would you rather use the google one?

No. We don't actually like parts of the Google style guide :P

> noUnusedParameters isn't turned on in tsconfig, it would be nice if it was
> Stylistic rules are always a bit flaky, it may be worth trying out prettier or clang-format at some point and turning eslint's stylistic rules off

Let's scope this PR just to a change in tooling. We can discuss style changes in a separate issue. Does that work for you?

43081j:
thats fair enough, ill fix up the lint errors and what not this week some time.

can do a separate PR for `noUnusedParameters` as it seems to be an easy one.

and an issue for the formatting if i get chance 👍

PR #704, opened 2019-10-01
enable stricter compilation options

43081j:
This was a quick one so if you don't actually enjoy the whole `_foo` style typescript uses, feel free to abandon the PR.

I enabled any remaining strict compilation options in the tsconfig:

* `noUnusedParameters`
* `noFallthroughCasesInSwitch`
* `noImplicitReturns`

Worthwhile IMO seeing as the code almost already followed these three rules.


surma:
I disagree with `noImplicitReturns`. It’s very much idiomatic to have a function without an implicit return value.

I personally dislike `noFallthroughCasesInSwitch` (i.e. having to put a leading underscore on unused function parameters. But curious what others think here.

Definitely in favor of enabling `noFallthroughCasesInSwitch`.

43081j:
That's fair enough. `noUnusedParameters` is pretty useful for keeping things clean and concise. But it's does pollute public interfaces a bit, especially if you have matching jsdoc so your parameters need the underscore in them. 

though that doesn't really apply here. i'd recommend keeping it but your call

PR #709, opened 2019-10-30
add maskable icon

myakura:
Support for maskable icon is landed on WebAPK and Fenix Nightly. I see Proxx already supporting maskable icon and think Squoosh should do the same.


surma:
Thanks a bunch, @myakura!

PR #710, opened 2019-10-30
Deploy

jakearchibald:
There's been a lot of changes, so we should have a click around and check everything works.


surma:
Did some clickies. All codecs and resize algos seem to work.

jakearchibald:
Same.

PR #720, opened 2019-12-17
Set loading state to false on error

bojan88:
I noticed that loading state is only set to false after successful image processing, and in case of error the spinner never stops spinning.

Steps to reproduce (might not work for everybody):
* Load the "device screen" sample photo
* Click resize
* Click up arrow to increase width

Noticed on Chrome 79 and Firefox 72 on Linux


bojan88:
@googlebot I signed it!

Issue #721, opened 2020-01-07
Codec to consider: JPEG XL


mathiasbynens:
I searched within this repo and GitHub returned no results, so... here's an issue about the JPEG XL codec :)

More info about this codec: https://jpeg.org/jpegxl/

JS/Wasm demo: https://libwebpjs.appspot.com/jpegxl/

jakearchibald:
Isn't this really licencey and patenty?

surma:
I am working on this now. To be clear, JPEG X**R** is patenty, JPEG X**L** is royalty-free and open source.

cctse:
I found a cool project, https://github.com/imager-io/imager，it's really impressive in some cases

surma:
Shipped!

PR #722, opened 2020-02-06
AVIF support

surma:
This PR lands support for AVIF.

So far, the codec is ported and integrated into Squoosh. Options UI needs polish, which @jakearchibald will do.


surma:
Closes #721

codecs/avif_dec/build.sh
```diff
@@ -0,0 +1,84 @@
...
+echo "============================================="
+echo "Compiling libavif"
+echo "============================================="
+test -n "$SKIP_LIBAVIF" || (
+  cd node_modules/libavif
+  mkdir -p build
+  cd build
+  emcmake cmake \
+    DCMAKE_BUILD_TYPE=Release \
+    -DAVIF_CODEC_AOM=1 \
```
RReverser: Out of curiosity - why this codec and not some other(s)? (I don't know anything about AVIF, so this might be a stupid question.)

codecs/avif_dec/build.sh
```diff
@@ -0,0 +1,84 @@
...
+   emcc \
+    ${OPTIMIZE} \
+    --bind \
+    -s ALLOW_MEMORY_GROWTH=1 \
+    -s MODULARIZE=1 \
+    -s 'EXPORT_NAME="avif_dec"' \
+    --std=c++11 \
+    -I ./node_modules/libavif/include \
+    -o ./avif_dec.js \
+    -x c++ \
```
RReverser: I've noticed this repeated in other C++ scripts and been meaning to clean up. There is no need to pass this explicitly when extension is `.cpp`.

codecs/avif_enc/avif_enc.cpp
```diff
@@ -0,0 +1,88 @@
...
+  int subsample;
+};
+
+avifRWData output = AVIF_DATA_EMPTY;
+val encode(std::string buffer, int width, int height, AvifOptions options) {
+  int depth = 8;
+  avifPixelFormat format;
+  switch (options.subsample) {
+  case 0:
+    format = AVIF_PIXEL_FORMAT_YUV420;
```
RReverser: Let's expose the actual underlying enum instead.

codecs/avif_enc/avif_enc.cpp
```diff
@@ -0,0 +1,88 @@
...
+    break;
+  }
+  avifImage *image = avifImageCreate(width, height, depth, format);
+
+  uint8_t *rgba = (uint8_t *)buffer.c_str();
+
+  avifImageAllocatePlanes(image, AVIF_PLANES_RGB);
+  avifImageAllocatePlanes(image, AVIF_PLANES_A);
+
+  for (int y = 0; y < height; y++) {
```
RReverser: As far as I can tell, this is not necessary and we can assign `pixels` directly to our data without allocating an extra copy and copying everything byte-by-byte. https://github.com/AOMediaCodec/libavif/blob/master/include/avif/avif.h#L371-L410

codecs/avif_dec/example.html
```diff
@@ -0,0 +1,24 @@
...
+    const result = Module.decode(image);
+    console.log(result.width, result.height, result.buffer);
+    const imageData = new ImageData(new Uint8ClampedArray(result.buffer), result.width, result.height);
+    Module.free_result();
+    const canvas = document.createElement('canvas');
+    canvas.width = result.width;
+    canvas.height = result.height;
+    document.body.appendChild(canvas);
+    const ctx = canvas.getContext('2d');
+    ctx.putImageData(imageData, 0, 0);
```
RReverser: Don't we use `ImageData` that still points to a freed memory? I know this currently works because we don't call into Wasm again, but this is very unsafe / UB.

src/codecs/avif/decoder.ts
```diff
@@ -0,0 +1,20 @@
...
+  const module = await emscriptenModule;
+  const rawImage = module.decode(data);
+  const result = new ImageData(
+    new Uint8ClampedArray(rawImage.buffer),
+    rawImage.width,
+    rawImage.height,
+  );
+
+  module.free_result();
+  return result;
```
RReverser, line 15: Same here as above - we're returning an `ImageData` that points to a freed memory and that can be potentially overridden. This is very unsafe.

src/codecs/avif/encoder.ts
```diff
@@ -0,0 +1,18 @@
...
+export async function encode(data: ImageData, options: EncodeOptions): Promise<ArrayBuffer> {
+  if (!emscriptenModule) emscriptenModule = initEmscriptenModule(avif_enc, wasmUrl);
+
+  const module = await emscriptenModule;
+  const resultView = module.encode(data.data, data.width, data.height, options);
+  const result = new Uint8Array(resultView);
+  module.free_result();
+
+  // wasm can’t run on SharedArrayBuffers, so we hard-cast to ArrayBuffer.
+  return result.buffer as ArrayBuffer;
```
RReverser, line 19: Same (although now I've looked through other codecs and I see that we do this everywhere 😶).

RReverser: Actually no, this one is fine because we create a new copy of `Uint8Array` unlike in `ImageData` cases.

src/codecs/avif/decoder.ts
RReverser: We do this in few other codecs too, sigh.

codecs/avif_dec/build.sh
surma: No deeper reasoning. I just wanted to get _one_ to work to get started (and I do not know if they provide different capabilities).

Seems like since I started more impls have been added, as well. Feel free to experiment.

codecs/avif_dec/build.sh
surma: My C++ days are pretty long behind me, so I think I just poked around the CLI flags until it worked with MozJPEG and copy-pasted the same invocation ever since. Any cleanup is appreciated!

codecs/avif_dec/example.html
surma: As you noticed, I did this a lot 😅  
`free_result()` just uses `free()` which, AFAIU, doesn’t immediately overwrite the freed memory region, so this was fine-ish. I agree that this should probably be done correctly.

RReverser: > doesn’t immediately overwrite the freed memory region

It's technically implementation-dependant. While it doesn't do that in Emscripten, it makes me quite uncomfortable (especially having run into actual use-after-free bugs in the past in the *wild*).

codecs/avif_enc/avif_enc.cpp
jakearchibald: Yeah, I'll pick this up (and as many other options as I can find) when I do the UI

src/codecs/avif/decoder.ts
jakearchibald: This might explain some of the errors we've seen when lots of quick changes are made.

RReverser: Lol totally.

codecs/avif_dec/build.sh
```diff
@@ -0,0 +1,84 @@
...
+export LDFLAGS="${OPTIMIZE}"
+export CFLAGS="${OPTIMIZE}"
+export CPPFLAGS="${OPTIMIZE}"
+
+echo "============================================="
+echo "Compiling libaom"
+echo "============================================="
+test -n "$SKIP_LIBAOM" || (
+  cd node_modules/libavif/ext  
+  test -d aom || git clone -b v1.0.0-errata1-avif --depth 1 https://aomedia.googlesource.com/aom aom
```
RReverser: I've noticed we use `napa` in other codecs, but manual `git clone` here. I don't have preference either way, but probably best to stick to one method for all? Which one would it be?

surma: I prefer napa. I _think_ I went with a manual clone here for `depth`. Not sure. Feel free to switch.

RReverser:
@surma Just copying some of my unresolved comments from resolved conversation on the build system PR, just so that they're not lost:

> I suspect you'll need to free `output` too, like the official example does, using `avifRWDataFree`.

> You'll probably also want to perform free/destroy above the `encodeResult` condition, otherwise those structures will get leaked on encoding errors.

> I'd probably recommend trying this out in Squoosh with sanitizer enabled once we are done, but this can be left as part of the original PR.

surma:
I (hopefully) fixed the leaky bits now :)

RReverser:
> I (hopefully) fixed the leaky bits now :)

The code looks good to me now, but I'd still double-check with sanitizers enabled and the manual "run leak checks" function call like in the blog post to be sure :)

surma:
Oh yeah, I’m definitely looking forward to you unleashing the sanitizer on the code base :D

src/lib/util.ts
```diff
@@ -104,6 +104,7 @@ const magicNumberToMimeType = new Map<RegExp, string>([
   [/^II*/, 'image/tiff'],
   [/^MM\x00*/, 'image/tiff'],
   [/^RIFF....WEBPVP8[LX ]/, 'image/webp'],
+  [/^\x00\x00\x00\x20\x66\x74\x79\x70\x61\x76\x69\x66\x00\x00\x00\x00/, 'image/avif'],
```
RReverser:
```suggestion
  [/^\x00\x00\x00 ftypavif\x00\x00\x00\x00/, 'image/avif'],
```

RReverser:
> Oh yeah, I’m definitely looking forward to you unleashing the sanitizer on the code base :D

I already did on the existing codecs, now it needs to be done on any new ones by whoever adds them :P

surma:
Thanks for volunteering.

codecs/avif/dec/avif_dec.cpp
```diff
@@ -0,0 +1,59 @@
+#include <emscripten/bind.h>
+#include <emscripten/val.h>
+#include "avif/avif.h"
+
+using namespace emscripten;
+
+class RawImage {
```
RReverser: @surma Please remove this in favour of `ImageData` using other codecs as examples.

RReverser:
> Thanks for volunteering.

Nice try.

codecs/avif/enc/avif_enc.d.ts
```diff
@@ -0,0 +1,9 @@
+import { EncodeOptions } from '../../src/codecs/avif/encoder-meta';
+
+interface AVIFModule extends EmscriptenWasm.Module {
+  encode(data: BufferSource, width: number, height: number, options: EncodeOptions): Uint8Array;
+  free_result(): void;
```
RReverser: TODO: remove this.

surma:
@RReverser The current code passed the sanitizer test :)

<details>

Screenshot 2020-08-11 at 18 41 16

</details>


I made a commit and reverted it to keep the code available. I didn’t want to do all the macro juggling at this point without making a plan with you.

RReverser:
Awesome! I see you also temporarily fixed indentation to tabs in the Makefile (my fault 😔), could you please re-introduce those fixes back?

image

surma:
Done

RReverser (approved):

nucliweb:
Congrats @surma 🤘🏼😊🤘🏼

PR #723, opened 2020-02-07
JPEG XL support

### LINKED ISSUE #721: Codec to consider: JPEG XL

surma:
This PR lands support for JPEG XL (or JXL).

The encoder currently crashes for some speed settings, but not for others. Not quite sure what’s happening there. 

Options UI will need polish, but the encoder options are bit mysterious to me as of yet. 
@jakearchibald said he’ll do the options UI once it’s good to go.

Closes #721


surma:
Related #721

surma:
Using the Squoosh logo as an example image, I see the following behavior:

- Speed=7: Success
- Speed=6: Success
- Speed=5: Success
- Speed=4: Success
- Speed=3: Success
- Speed=2: `/src/node_modules/jxl/jxl/enc_adaptive_quantization.cc:726: JXL_ASSERT: qf_higher / qf_lower < 253`
- Speed=1: Success

In all results, transparent pixels are replaced with black. 

Using the red fox picture:

- Speed=7: Success
- Speed=6: `Cannot enlarge memory, asked to go up to 2159349760 bytes, but the limit is 2147483648 bytes!`
- Speed=5: Success
- Speed=4: Success
- Speed=3: Success
- Speed=2: `/src/node_modules/jxl/jxl/enc_adaptive_quantization.cc:726: JXL_ASSERT: qf_higher / qf_lower < 253`
- Speed=1: `jxl_enc.js:5333 Cannot enlarge memory, asked to go up to 2170884096 bytes, but the limit is 2147483648 bytes!`

It is surprising to me that we are hitting the 2GB limit, but I guess some parameters might still be wrong.

codecs/jxl/dec/jxl_dec.cpp
```diff
@@ -0,0 +1,73 @@
...
+  for (int y = 0; y < height; y++) {
+    const float *red = buffer->PlaneRow(0, y);
+    const float *green = buffer->PlaneRow(1, y);
+    const float *blue = buffer->PlaneRow(2, y);
+    for (int x = 0; x < width; x++) {
+      int pixelOffset = width * y + x;
+      result[pixelOffset * 4 + 0] = clamp(red[x], 0, 255);
+      result[pixelOffset * 4 + 1] = clamp(green[x], 0, 255);
+      result[pixelOffset * 4 + 2] = clamp(blue[x], 0, 255);
+      result[pixelOffset * 4 + 3] = 255;
```
jonsneyers: You can find the alpha channel by testing for `main->HasAlpha()` and if true `main->alpha()` gives you the alpha channel.

codecs/jxl/enc/jxl_enc.cpp
```diff
@@ -0,0 +1,57 @@
...
+
+val encode(std::string image, int width, int height, JXLOptions options) {
+  jxl::CompressParams cparams;
+  jxl::PassesEncoderState passes_enc_state;
+  jxl::CodecInOut io;
+  jxl::PaddedBytes bytes;
+  jxl::ImageBundle* main = &io.Main();
+
+  cparams.speed_tier = static_cast<jxl::SpeedTier>(options.speed);
+  cparams.color_transform = jxl::ColorTransform::kNone;
```
jonsneyers: Better to leave color_transform at its default.

codecs/jxl/enc/jxl_enc.cpp
```diff
@@ -0,0 +1,57 @@
...
+};
+
+val encode(std::string image, int width, int height, JXLOptions options) {
+  jxl::CompressParams cparams;
+  jxl::PassesEncoderState passes_enc_state;
+  jxl::CodecInOut io;
+  jxl::PaddedBytes bytes;
+  jxl::ImageBundle* main = &io.Main();
+
+  cparams.speed_tier = static_cast<jxl::SpeedTier>(options.speed);
```
jonsneyers, line 25: Maybe do `10-options.speed` so it behaves the same as in `cjpegxl`(and set the default somewhere in the middle instead of very-slow or very-fast).

codecs/jxl/enc/jxl_enc.cpp
```diff
@@ -0,0 +1,57 @@
...
+  cparams.color_transform = jxl::ColorTransform::kNone;
+
```
jonsneyers, line 27: You'll also want an option to set the target quality (`cparams.butteraugli_distance`) – it has a useful range from say 0.5 (near-lossless) to 8 (very lossy); the default is 1 which is one unit of just-noticeable-difference.

codecs/jxl/enc/jxl_enc.cpp
jonsneyers: Setting it to kNone is very much untested and will likely completely break things in lossy mode – this is only intended for lossless encoding, where you don't want to do irreversible color transforms. The default is kXYB which is what the lossy encoder is expecting to work with. The only reason we even allow kNone (and kYCbCr) in lossy mode is that we need it for lossless JPEG recompression.

codecs/jxl/dec/jxl_dec.cpp
surma: Yeah I was skipping alpha until I got the basics working. Thanks for this explanation!

codecs/jxl/enc/jxl_enc.cpp
surma: Ah, so that seems to fix the assertion error at speed=2. However, now the colors are wrong. I guess I need to specify sRGB somewhere?
Screenshot 2020-08-28 at 12 18 15

surma: Actually, no. I use `SetFromSRGB()`, so it’s already marked as sRGB input. I think that’s why I originally ended up using `kNone` because it fixed the colors 😅

codecs/jxl/enc/jxl_enc.cpp
surma: Added!

codecs/jxl/dec/jxl_dec.cpp
surma: I just added alpha support :)

codecs/jxl/enc/jxl_enc.cpp
surma: After adding alpha support just now, I wonder if the discolorization is a _decoder_ issue? I am doing this clamping on an `Image3F` which might be wrong? 

https://github.com/GoogleChromeLabs/squoosh/blob/321996807e7caefb8d72316d40d6e830fb2dc3c8/codecs/jxl/dec/jxl_dec.cpp#L42-L62

esx2ve: Digging into the same issue here. [This issue](https://gitlab.com/wg1/jpeg-xl/-/issues/51) from the reference implementation has some more info.

RReverser:
Ugh come on everyone has signed it...

RReverser:
@googlebot I consent.

surma:
LGTM

PR #728, opened 2020-02-27
Use closure in optipng build, which shrinks the JS to less than half

kripken:
Specifically the size shrinks from 108K to 46K :smiley: 

I'm hopeful this just works, as the project already uses MODULARIZE (so nothing can access internals from outside, which is dangerous with closure). Is there a test suite I should run locally? Or will CI here run tests?

### DIFF
--- a/codecs/optipng/build.sh
+++ b/codecs/optipng/build.sh
@@ -27,6 +27,7 @@ echo "============================================="
   emcc \
     --bind \
     ${OPTIMIZE} \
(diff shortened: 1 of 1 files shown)


surma:
We don’t have tests 🙄 
But we do have [branch previews](https://deploy-preview-728--squoosh.netlify.com/)! 


And _holy crap_ at that size reduction! Thank you.

PR #743, opened 2020-04-09
Add encoding/decoding times to console

RReverser:
This intentionally excludes time of loading corresponding modules, and only measures actual processing.

While this is not perfect as it's not integrated in the UI (cc @jakearchibald), it gives at least some way to measure performance of different codecs and their integrations on particular files.

### DIFF
--- a/src/codecs/processor-worker/index.ts
+++ b/src/codecs/processor-worker/index.ts
@@ -2,13 +2,18 @@ import { expose } from 'comlink';
 import { isHqx } from '../resize/processor-meta';
 import { clamp } from '../util';
 
+function timed<T>(name: string, func: () => Promise<T>) {
+  console.time(name);
+  return func().finally(() => console.timeEnd(name));
+}
+
 async function mozjpegEncode(
   data: ImageData, options: import('../mozjpeg/encoder-meta').EncodeOptions,
 ): Promise<ArrayBuffer> {
   const { encode } = await import(
     /* webpackChunkName: "process-mozjpeg-enc" */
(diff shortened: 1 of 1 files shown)


src/codecs/processor-worker/index.ts
```diff
@@ -2,13 +2,18 @@ import { expose } from 'comlink';
...
+}
+
 async function mozjpegEncode(
   data: ImageData, options: import('../mozjpeg/encoder-meta').EncodeOptions,
 ): Promise<ArrayBuffer> {
   const { encode } = await import(
     /* webpackChunkName: "process-mozjpeg-enc" */
     '../mozjpeg/encoder');
-  return encode(data, options);
+  return timed('mozjpegEncode', () => encode(data as any, options));
```
RReverser, line 16: This one was an accident, reverting.

RReverser: Done.

RReverser:
@surma Was thinking about it, but AFAIK for those you need to explicitly start Performance measurement in DevTools, which would also instrument and measure all the other functions? If so, seems a bit like overkill if so.

RReverser:
In the long run, I'd like this to somehow be part of UI anyway, but @jakearchibald said it might be complicated as we don't have much free space left.

For now I'll merge with at least console timings, but open to any improvements in future.

nucliweb:
What points does it show on a layer, like in debug mode?

nucliweb:
What points does it show on a layer, like in debug mode?

PR #750, opened 2020-05-01
Switch to Emscripten upstream

RReverser:
Looks like we've been stuck on Emscripten fastcomp backend, misssing out on all new optimisations between LLVM 6 and LLVM 11 and going via slow asm.js pipeline when building.

This changes Docker images for Emscripten projects to point to emscripten-upstream instead and commits the updated artifacts.

---

**Note**: it looks like most of the Wasm outputs got reduced by 7-13%, but mozjpeg_enc.wasm grew by 30% (61.8KB) instead. In principle, it's not critical for us, especially after compression (e.g. gzip is larger only by 10% - 5.3 KB), but maybe worth looking into by someone on Emscripten side? (cc @kripken)


codecs/imagequant/imagequant.cpp
```diff
@@ -61,21 +61,21 @@ RawImage quantize(std::string rawimage, int image_width, int image_height, int n
...
-  {.a = 255, .r = 215, .g = 215, .b = 0},   // regular yellow
-  {.a = 255, .r = 215, .g = 215, .b = 215}, // regular white
-  {.a = 255, .r = 0, .g = 0, .b = 255},     // bright blue
-  {.a = 255, .r = 255, .g = 0, .b = 0},     // bright red
-  {.a = 255, .r = 255, .g = 0, .b = 255},   // bright magenta
-  {.a = 255, .r = 0, .g = 255, .b = 0},     // bright green
-  {.a = 255, .r = 0, .g = 255, .b = 255},   // bright cyan
-  {.a = 255, .r = 255, .g = 255, .b = 0},   // bright yellow
-  {.a = 255, .r = 255, .g = 255, .b = 255}  // bright white
+  {.r = 0, .g = 0, .b = 0, .a = 255},       // regular black
```
RReverser, line 64: Note: I made this change to accomodate for a new C++ warning about strict order of initialization.

PR #755, opened 2020-05-04
Add LTO for C++ builds

### LINKED ISSUE #754: Enable LTO for C++ codecs
Follow-up from discussion at https://github.com/emscripten-core/emscripten/issues/11055 which tried to investigate why switch from Fastcomp to LLVM backend in #750 made MozJPEG larger than before:

> @kripken: Ah, no need, I figured this out. The difference vanishes if building with LTO. The wasm is then 2.5% smaller than fastcomp in fact.
> 
> Fastcomp runs `-globaldce` by default, and it must be doing quite a lot on this project. Upstream uses wasm object files by default, and so we don't have that.
> 
> Would be good to build that codec with `-flto` on all compile commands.
>
> @RReverser: Ah sounds good. I've actually tried -flto in past, but didn't see any improvement, so decided to leave builds without it.
>
> Back then I didn't know that we're still using Fastcomp backend (it would probably be good to raise an issue on that Docker container's repo, as it's not very obvious), so that probably explains it all.

RReverser:
This didn't have much effect on fastcomp builds, but provides further size savings with new LLVM backend we switched to in #750 (and fixes the MozJPEG size regression from the same PR).

In the future we won't need to pass `--llvm-lto 1` explicitly, but latest Emscripten Docker image doesn't contain the Emscripten version with the necessary fixes for this.

Fixes #754.


RReverser:
> (Do you know why LTO isn’t enabled by default?)

I think it reflects default choice in most other compilers, which is traditionally preferring slightly faster builds over slightly better optimised ones. Especially because in the past LTO has been much slower than it is nowadays.

PR #756, opened 2020-05-12
Config improvements for working on C / C++

RReverser:
Various config improvements for easier continuous work on C / C++:

 - Autoformatted C / C++ with `clang-format` for consistent [Chromium] style.
 - Removed unnecessary `-x c++` from build scripts.
 - Improved rebuild speed by caching the safe bits:
   - Removed `-f` from `autoreconf` to reuse generated configure scripts.
   - Added `-C` to `./configure` to reuse stored information about `emcc`.
   - Removed `rm -rf build` from WebP encoder/decoder.
   - Stored Emscripten cache inside node_modules/.em_cache. Docker image ships without LTO libs, so Emscripten had to rebuild stdlibs on every build otherwise.
 - Merged webp_enc + webp_dec build scripts. Core libwebp library is same in both cases, so there's no point in storing and building two copies of it.


surma:
Haven’t taken a look at the details, but I’m not sure about removing `--rm` from Docker as it will create a new docker VM every time the command runs.

RReverser:
Yeah I suppose you're right; I thought it would help me preserve Emscripten cache, but in the end had to move it to a mounted folder anyway. I'll restore that one.

RReverser:
Updated, please take another look.

PR #764, opened 2020-06-23
Adds install button and analytics to track how often install is available.

petele:
- Adds an 'Install Squoosh' button the the intro element
- Adds analytics: if install is available, when Squoosh is installed, and what method was used to install Squoosh.

Screen Shot 2020-06-23 at 4 13 12 PM

### DIFF
--- a/src/components/intro/index.tsx
+++ b/src/components/intro/index.tsx
@@ -47,11 +47,27 @@ interface Props {
 }
 interface State {
   fetchingDemoIndex?: number;
+  deferredPrompt?: BeforeInstallPromptEvent;
+  installSource?: String;
 }
 
 export default class Intro extends Component<Props, State> {
-  state: State = {};
+  state: State = {
+    deferredPrompt: undefined,
+    installSource: undefined,
+  };
   private fileInput?: HTMLInputElement;
+  private installButton?: HTMLButtonElement;
+
+  constructor() {
+    super();
+
+    // Listen for beforeinstallprompt events, indicating Squoosh is installable.
+    window.addEventListener('beforeinstallprompt', this.onBeforeInstallPromptEvent);
+
+    // Listen for the appinstalled event, indicating Squoosh has been installed.
+    window.addEventListener('appinstalled', this.onAppInstalled);
+  }
 
   @bind
   private resetFileInput() {
@@ -88,6 +104,62 @@ export default class Intro extends Component<Props, State> {
       this.setState({ fetchingDemoIndex: undefined });
       this.props.showSnack("Couldn't fetch demo image");
     }
+  }
+
+  @bind
+  private onBeforeInstallPromptEvent(event: BeforeInstallPromptEvent) {
+    // Don't show the mini-infobar on mobile
+    event.preventDefault();
+
+    // Save the beforeinstallprompt event so it can be called later.
+    this.setState({ deferredPrompt: event });
+
+    // Log the event.
+    ga('send', 'event', 'pwa-install', 'available');
+
+    // Make the install button visible
+    this.installButton!.style.display = 'inline-block';
+  }
+
+  @bind
+  private async onInstallClick(event: Event) {
+    // Get the deferred beforeinstallprompt event
+    const deferredPrompt = this.state.deferredPrompt;
+
+    // If there's no deferred prompt, bail.
+    if (!deferredPrompt) return;
+
+    // Set the install source as the intro install button.
+    const installSource = 'introInstallButton';
+    this.setState({ installSource });
+
+    // Show the browser install prompt
+    deferredPrompt.prompt();
+
+    // Wait for the user to accept or dismiss the install prompt
+    const response = await deferredPrompt.userChoice;
+
+    // Get the outcome and log it
+    const outcome = response.outcome;
+    ga('send', 'event', 'pwa-install', installSource, outcome);
+
+    // If the prompt was dismissed, clear the installSource.
+    if (outcome === 'dismissed') {
+      this.setState({ installSource: undefined });
+    }
+  }
+
+  @bind
+  private onAppInstalled() {
+    // If install button is visible, hide it.
+    const installButton = this.installButton;
+    if (installButton) {
+      installButton.style.display = 'none';
+    }
+
+    // Try to get the install, if it's not set, use 'browser'
+    const source = this.state.installSource || 'browser';
+    ga('send', 'event', 'pwa-install', 'installed', source);
   }
 
   render({ }: Props, { fetchingDemoIndex }: State) {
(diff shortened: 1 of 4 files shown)


src/components/intro/index.tsx
```diff
@@ -90,6 +106,62 @@ export default class Intro extends Component<Props, State> {
...
+    this.setState({ installSource });
+
+    // Show the browser install prompt
+    deferredPrompt.prompt();
+
+    // Wait for the user to accept or dismiss the install prompt
+    const response = await deferredPrompt.userChoice;
+
+    // Get the outcome and log it
+    const outcome = response.outcome;
```
surma:
```suggestion
    const {outcome} = await deferredPrompt.userChoice;
```

src/components/intro/index.tsx
```diff
@@ -47,11 +47,27 @@ interface Props {
...
+  deferredPrompt?: BeforeInstallPromptEvent;
+  installSource?: String;
 }
 
 export default class Intro extends Component<Props, State> {
-  state: State = {};
+  state: State = {
+    deferredPrompt: undefined,
+    installSource: undefined,
+  };
```
surma:
```suggestion
  state: State = {};
```

`undefined` is implicit :D

src/components/intro/style.scss
```diff
@@ -170,6 +170,26 @@
...
+  border: 1px solid #e8e8e8;
+
+  margin-top: 1em;
+
+  align-items: center;
+  padding: 14px;
+  font-size: 1.3rem;
+
+  display: none;
+}
```
surma:
```suggestion
.install-button {
  composes: unbutton from '../../lib/util.scss';

  &:hover,
  &:focus {
    background: #f5f5f5;
  }

  background: #fff;
  border: 1px solid #e8e8e8;

  align-items: center;
  padding: 14px;
  font-size: 1.3rem;

  position: absolute;
  top: 1rem;
  right: 1rem;
  display: none;
  opacity: 0;

  animation: fade-in .3s linear 1s forwards;
}

@keyframes fade-in {
	to {opacity: 1;}
}
```

This places the button in the top right-hand corner with a quick fade-in. Because the `beforeinstallprompt` event can be quite delayed, the previous placement can make the button plop into view and be quite jarring and also cause a layout shift.

<details>

Screenshot 2020-06-24 at 13 13 16

</details>

Not sure this is what we want to ship, but just something to start the convo.

surma (changes requested):
Overall this LGTM. Couple of nits. I also wonder if we can avoid the style duplication for the button. I tried a couple of `compose` combinations, but couldn’t get it to look right.

@jakearchibald do you mind taking a look at this as well?

src/components/intro/index.tsx
petele: I've never seen this syntax before, any pointer to a quick explanation?

src/components/intro/style.scss
petele: I'm OK with this design, I primarily wanted to get something in that would have both an in-app promo, plus the browser promo. Potentially we can A/B test different locations later.

src/components/intro/index.tsx
```diff
@@ -90,6 +103,59 @@ export default class Intro extends Component<Props, State> {
...
+
+    // Set the install source as the intro install button.
+    const installSource = 'introInstallButton';
+    this.setState({ installSource });
+
+    // Show the browser install prompt
+    deferredPrompt.prompt();
+
+    // Wait for the user to accept or dismiss the install prompt
+    const {outcome} = await deferredPrompt.userChoice;
```
petele: @surma - I haven't seen this syntax before: `const {outcome} = await deferredPrompt.userChoice;` Any pointers to docs about it so I can understand it better and use it in the future?

src/components/intro/index.tsx
surma: It’s [destructuring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Object_destructuring)

src/components/intro/index.tsx
jakearchibald: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Object_destructuring

petele: Sweet - thanks.

src/components/intro/index.tsx
```diff
@@ -41,17 +41,31 @@ const demos = [
...
 }
 interface State {
   fetchingDemoIndex?: number;
+  installEvent?: BeforeInstallPromptEvent;
 }
 
 export default class Intro extends Component<Props, State> {
   state: State = {};
   private fileInput?: HTMLInputElement;
+  private installingViaButton = false;
```
jakearchibald, line 58: I made this an instance member, rather than state, since it doesn't impact rendering.

src/components/intro/index.tsx
```diff
@@ -132,6 +191,15 @@ export default class Intro extends Component<Props, State> {
...
         </div>
+        {beforeInstallEvent &&
+          <button
+            type="button"
+            class={style.installButton}
+            onClick={this.onInstallClick}
+          >
+            Install
+          </button>
+        }
```
jakearchibald, line 202: Using JSX to control the rendering of the button

jakearchibald:
I've made some tweaks. @surma could you double check? I'm fine with the style duplication - we can look at this in more detail when we work on this again.

src/components/intro/index.tsx
petele: Squoosh only has one in-app install button, so this works fine. 

The code I used was written to support multiple install buttons on a page, and make it possible to track which button the initiated the install prompt. The idea being each button click handler would set a different identifier in `installSource`. Then when we'd be able to tell the CTR and acceptance of each button, and what specific button (or browser interface) was use used to install the PWA

surma (approved):
Reviewed and tested. LGTM.

RReverser:
This change seems to fail the build for me:

```
[tsl] ERROR in C:\src\squoosh\src\components\intro\index.tsx(197,26)
      TS2339: Property 'installButton' does not exist on type '{ readonly "intro": string; readonly "logoContainer": string; readonly "logoSizer": string; readonly "logo": string; readonly "openImageGuide": string; readonly "selectButton": string; readonly "hide": string; ... 11 more ...; readonly "relatedLinks": string; }'.
```

Do I need to regenerate types somehow?

RReverser:
FWIW I can also see that `src\components\intro\style.scss.d.ts`, which is git-ignored and so, I assume, autogenerated, doesn't contain that property.

@jakearchibald Is there some command I need to run to regenerate it?

RReverser:
Hmm I wonder if this is yet another "Webpack regexes and Windows" problem like a bunch of others I've seen before. Could be some pattern in config still not matching the styles due to `\` or something like that...

RReverser:
Found https://github.com/GoogleChromeLabs/squoosh/blob/dev/config/add-css-types.js, although still not sure how to force regeneration...

PR #771, opened 2020-07-07
Update privacy link

atjn:
Hi, thanks for this awesome app!

The privacy link on the front page is currently broken, as the default branch has been renamed. This PR fixes that.

I also believe `branch` should be updated here, but i'm not 100% sure:
https://github.com/GoogleChromeLabs/squoosh/blob/d94835402fe33b449d8c5bff09334e2f3e6ac175/sizereport.config.js#L3-L6


jakearchibald (approved):
Good catch, thank you!

PR #765, opened 2020-06-23
Use a dimension to note how the user opened squoosh

petele:
- Checks how Squoosh was started and uses a dimension to segment pageviews between standalone and browser.

PR #772, opened 2020-07-07
Updates to analytics for PWAs

petele:
This PR updates the behaviour in #764 and #765.

* Adds query string to the manifest `start_url` to track launches from the PWA 
  (`?utm_medium=PWA&utm_source=launcher`), ideally this will allow us to remove the dimension tracking we're 
  currently using.
* Changes the install button to the same color as the purple text to see if it has a higher CTR, hover color is the 
  same, but darkens the color by lowering the lightness value.
* Adds [`nonInteraction`](https://developers.google.com/analytics/devguides/collection/analyticsjs/events#non-interaction_events) 
  parameter to the `beforeinstallprompt` event analytics logging, since it's not user initiated.
* Standardizes the `action` and `label` names for events fired on install available, and when clicked. 
* Uses an event `value` instead of a label to track if the install was accepted from the install prompt.

These changes will help validate our best practices for analytics in PWAs.

### DIFF
--- a/src/components/intro/index.tsx
+++ b/src/components/intro/index.tsx
@@ -41,7 +41,7 @@ const demos = [
   },
 ];
 
-const installButtonSource = 'introInstallButton';
+const installButtonSource = 'introInstallButton-Purple';
 
 interface Props {
   onFile: (file: File | Fileish) => void;
@@ -113,7 +113,12 @@ export default class Intro extends Component<Props, State> {
     this.setState({ beforeInstallEvent: event });
 
     // Log the event.
-    ga('send', 'event', 'pwa-install', 'available');
+    const gaEventInfo = {
+      eventCategory: 'pwa-install',
+      eventAction: 'shown',
+      nonInteraction: true,
+    };
+    ga('send', 'event', gaEventInfo);
   }
 
   @bind
@@ -130,7 +135,14 @@ export default class Intro extends Component<Props, State> {
 
     // Wait for the user to accept or dismiss the install prompt
     const { outcome } = await beforeInstallEvent.userChoice;
-    ga('send', 'event', 'pwa-install', installButtonSource, outcome);
+    // Send the analytics data
+    const gaEventInfo = {
+      eventCategory: 'pwa-install',
+      eventAction: 'custom-install',
+      eventLabel: installButtonSource,
+      eventValue: outcome === 'accepted' ? 1 : 0,
+    };
+    ga('send', 'event', gaEventInfo);
 
     // If the prompt was dismissed, we aren't going to install via the button.
     if (outcome === 'dismissed') {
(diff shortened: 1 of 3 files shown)


src/components/intro/index.tsx
```diff
@@ -113,7 +113,12 @@ export default class Intro extends Component<Props, State> {
     this.setState({ beforeInstallEvent: event });
 
     // Log the event.
-    ga('send', 'event', 'pwa-install', 'available');
+    const gaEventInfo = {
+      eventCategory: 'pwa-install',
+      eventAction: 'promo-shown',
+      nonInteraction: true,
+    };
+    ga('send', 'event', gaEventInfo);
```
petele, line 121: Big change here is that I've set `nonInteraction`, this doesn't have a big effect, but if it's not set, the bounce rates are incorrectly calculated, because Analytics *thinks* the user did something, but they haven't.

src/components/intro/index.tsx
```diff
@@ -130,7 +135,14 @@ export default class Intro extends Component<Props, State> {
...
     const { outcome } = await beforeInstallEvent.userChoice;
-    ga('send', 'event', 'pwa-install', installButtonSource, outcome);
+    // Send the analytics data
+    const gaEventInfo = {
+      eventCategory: 'pwa-install',
+      eventAction: 'promo-clicked',
+      eventLabel: installButtonSource,
+      eventValue: outcome === 'accepted' ? 1 : 0,
+    };
+    ga('send', 'event', gaEventInfo);
```
petele, line 145: This allows you to track several different install promos (each would be identified by `installButtonSource`, and uses the `value` to track if the user accepted or dismissed the install dialog (CTR). Before, we were putting the outcome in the label, which required looking at analytics and doing some manual calculations to get the CTR.

src/components/intro/style.scss
```diff
@@ -175,11 +175,12 @@
 
   &:hover,
   &:focus {
-    background: #f5f5f5;
+    background: #504488;
   }
 
-  background: #fff;
+  background: #5D509E;
```
petele, line 181: Pulled from the color of `.select-button`

src/manifest.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "Squoosh",
   "short_name": "Squoosh",
-  "start_url": "/",
+  "start_url": "/?utm_medium=PWA&utm_source=launcher",
```
petele, line 4: We're currently using a dimension and some custom code to track how the app was launched, but Analytics has a built in way for doing this with `utm_medium` and `utm_source`.

src/manifest.json
```diff
@@ -20,7 +20,7 @@
     }
   ],
   "share_target": {
-    "action": "/?share-target",
+    "action": "/?utm_medium=PWA&utm_source=share-target&share-target",
```
petele, line 23: The `utm_source` here is different because it's a different style of launch. In this case it's coming from a share target.

jakearchibald (approved):

PR #777, opened 2020-07-13
Consolidate Rust builds

RReverser:
This consolidates Rust build process for various codecs into a single top-level image that is built once and reused.

This ensures that we use same version of tools across codecs (now controlled from a single Dockerfile + project-specific Cargo.lock), simplifies build configs and commands, speeds up common builds and reduces disk space taken by Docker images by reusing same one.

Additionally, this PR renames all codecs to squoosh-* to work around the https://github.com/rustwasm/wasm-pack/issues/829 (which has been already fixed on master of wasm-pack but not released in a while), as well as adds `publish = false` to Cargo.toml to avoid accidental publishing for now.

I'm planning to do similar for Emscripten in a separate PR, although abilities to share configs there are much more limited due to lack of package manager in C++.


jakearchibald:
That's a pretty big jump in the size of hqx. Any idea why?

RReverser:
Huh, it is odd, all the others got smaller... I'll check.

RReverser:
Okay, I think the original Wasm wasn't up to date or built using the same script as was in the repo. It clearly didn't include console-error panic hook, even though it's enabled by default in `features` section.

In fact, panic hook re-exports seem to be completely unused in both codec wrappers where they're included, so I'll just remove them.

**UPD:** Hmm, there might be still more to this difference...

RReverser:
Okay, I don't get it. Even if I build HQX on the `dev` branch, I'm getting the same (larger) size.

Could the original Wasm be post-processed by something else after the `build.sh`?

jakearchibald:
@surma WHAT DID YOU DO???

RReverser:
@jakearchibald Other than that, does anything else stand out?

jakearchibald:
Nah, the rest looks good to me. Although @surma's done more work in this part of the project, so it'll be good for him to review.

surma (approved):
I don’t think I did anything to the HQX codec 😅 

LGTM!

RReverser:
> I don’t think I did anything to the HQX codec 😅

But... how is it so much larger then? :D

jakearchibald (changes requested):
Yeah, I don't think we should land this with such a bug size bump unless there's a good reason 😀

RReverser:
@jakearchibald I see your change in https://github.com/GoogleChromeLabs/squoosh/commit/b6fae0eb4c7ef68ccb2a2f5c19e5a54456f3765d made this file 83KB smaller, but in this case we already run `wasm-opt -Os ...` as part of wasm-pack, as well as strip debug data (except for the few bytes in `producers` section which, I think, are worth keeping anyway), so there must be something else I'm missing...

jakearchibald:
@surma are you able to build hqx at the smaller size we're seeing on live currently?

surma:
I just did a local build and I also get a ~390KB Wasm binary after `wasm-opt`.

That being said, since all other binaries got smaller, all codecs are lazyloaded and HQX is probably not used very often, I’d be okay not blocking this PR on this one codec (we should probably open an issue to investigate further at some point, tho).

RReverser:
> I just did a local build and I also get a ~390KB Wasm binary after `wasm-opt`.

Do you mean on this PR or before? My problem with investigating this size difference is that I couldn't get the smaller size even on the `dev` branch, without this PR:

> Okay, I don't get it. Even if I build HQX on the `dev` branch, I'm getting the same (larger) size.
> 
> Could the original Wasm be post-processed by something else after the `build.sh`?

Which makes me wonder whether either one of you did some manual post-processing, or if there is a serious regression in `wasm-opt` or Rust toolchains (our versions weren't pinned until this PR).

jakearchibald:
> HQX is probably not used very often

Unfortunately we always load it so the site can work offline.

> Which makes me wonder whether either one of you did some manual post-processing

Wasn't me 😀.

@surma if you can recreate the previous size with the versions of everything we currently use on live, then we can start figuring out which part of the toolchain bloated the size.

surma:
Ah, so after realizing that the only thing that isn’t pinned to a specific release is the Docker base image, I can reproduce the smaller binary (219KB) on `live` by adjusting the `codecs/hqx/Dockerfile` as follows:

```diff
- FROM rust
+ FROM rust:1.37
```

(1.37 is roughly a year old, which is the last time the binary file was updated.)

So something in the more recent Rust compiler versions generates bigger binaries.

RReverser:
Oh, that's very interesting. Did you pinpoint it to specifically 1.37 or is it just an arbitrary version from that time? Either way, whether you can pinpoint it by "bisecting" or not, it's worth submitting as an issue upstream - such different in the output filesize is huge by any standards.

surma:
I found out that `1.35` was roughly a year old and started incrementing from there until it compiled (some features were unstable until `1.37`).

> it's worth submitting as an issue upstream
Would you mind taking that on? You are definitely more knowledgable and established (and also I still have a freakin’ inbox that’s overflowing 😅 )

RReverser:
> Would you mind taking that on? You are definitely more knowledgable and established (and also I still have a freakin’ inbox that’s overflowing 😅 )

So I'm actually off starting Wednesday, returning next Tuesday, and there's still quite a bit to do during today/tomorrow. In case your inbox and todo list will get cleared earlier, it could be helpful if you could do that, but, if not, I can do that upon return :)

surma:
Sg :)

RReverser:
@jakearchibald In the face of this being an upstream Rust issue, do you think we should still block this PR? I suspect it will take a while to get it fixed.

surma:
I know it’s a 100K, but on a lazy-loaded resource and Squoosh was never that small to begin with in total size. I’d vote for moving forward with this.

jakearchibald:
Could we continue to use the old wasm for HQX?

surma:
Yeah, we could exclude HQX from using the universal Rust Dockerimage for now until we have a fix upstream.

jakearchibald:
+1 to that

jakearchibald:
@googlebot I consent.

Also you're a prick.

jakearchibald:
I've added a fix to our resizer here.

surma:
Gonna reset HQX to the old build system pinned to Rust v1.37 and then get this landed.

surma:
@jakearchibald HQX is reset to the old build system.

However, I still see codec crash when resizing with linear sRGB + premultiply 🤔 

(edit: Jake and I figured out that we forgot to commit the new `resize` wasm module and I kept building `hqx` 🤦 )

PR #780, opened 2020-07-17
Memory fixes for C++ codecs

RReverser:
Fixes various memory leaks and other pointer-related issues, while simplifying the exposed API and generally reducing number of lines.

(See the long description in the blog post draft or details in the individual commit messages.)

PR #781, opened 2020-07-17
Consolidate C++ builds

RReverser:
Use a shared base image with fixed Emscripten version, autotools and optimisation flags for all C++ codecs.

Additionally, move build commands for codecs themselves to Makefile - they're already platform-specific, and Make allows for better caching and parallelisation that custom ad-hoc scripts.

This is essentially same as #777 but for C++.

(This should be merged after #780.)


codecs/imagequant/Makefile
```diff
@@ -0,0 +1,33 @@
...
+CODEC_OUT_RELATIVE := libimagequant.a
+CODEC_OUT := $(addprefix $(CODEC_DIR)/, $(CODEC_OUT_RELATIVE))
+OUT_JS := imagequant.js
+OUT_WASM := $(OUT_JS:.js=.wasm)
+
+.PHONY: all clean
+
+all: $(OUT_JS)
+
+%.js: %.cpp $(CODEC_OUT)
```
surma, line 11: I did not miss Makefiles lol

surma (approved):

codecs/imagequant/Makefile
RReverser: Think of `&` as usual `*` in wildcards and it gets easier :D

codecs/mozjpeg_enc/Makefile
```diff
@@ -0,0 +1,45 @@
...
+
+$(CODEC_DIR)/rdswitch.o: $(CODEC_DIR)/Makefile
+	$(MAKE) -C $(CODEC_DIR) rdswitch.o
+
+$(CODEC_DIR)/Makefile: $(CODEC_DIR)/configure
+	cd $(CODEC_DIR) && ./configure \
+		--disable-shared \
+		--without-turbojpeg \
+		--without-simd \
+		--without-arith-enc \
```
RReverser, line 37: @surma By the way, just to double-check here - I've disabled few more options/modes that we don't use, and, as far as I can tell from their description, we don't want to use.

This shaved off another ~9.5 KB off the Wasm binary, but I want to make sure this is okay as I don't understand image codecs that well :)

RReverser: @surma Ping.

surma: Oh yeah, that’s totally fine by me if it still works :D

RReverser: Well I *hope* it does :D

PR #784, opened 2020-07-27
Don't fire install analytics on hidden pages

petele:
Minor tweaks to PWA install analytics 

* **Don't fire installed analytics event if page is hidden.** This prevents installs from being double counted if users have multiple tabs open and installs Squoosh.
* **Hard codes the `page` and `title` when `pageview` analytics are fired.** Ensures filenames are never reported to analytics.


surma (approved):

src/index.ts
```diff
@@ -24,7 +24,7 @@ if (typeof PRERENDER === 'undefined') {
   ga('create', 'UA-128752250-1', 'auto');
   ga('set', 'transport', 'beacon');
   ga('set', 'dimension1', displayMode);
-  ga('send', 'pageview');
+  ga('send', 'pageview', '/index.html', { title: 'Squoosh' });
```
jakearchibald, line 27: Should this be `/` rather than `/index.html`?

petele: In analytics, we've been tracking index.html since the beginning of July. Either work, and I tried to keep it consistent to what we have right now.

In an ideal world, we'd trigger `/` for intro, and `/editor` for when the user is in the editor, but that's adding a bunch more we probably don't need until v2.

PR #787, opened 2020-07-30
Switch from napa to curl

RReverser:
Instead of using 127 deps brought by napa just to download the one dependency we actually care about, just use curl & tar directly from Makefile.


jakearchibald:
Would this impact Windows compatibility?

RReverser:
We already build inside Docker, and codecs already use Make, so no :)

jakearchibald:
Ah cool, I thought the napa part currently ran outside of docker.

RReverser:
Ah yeah it did, but is it useful for user without ability to build?

jakearchibald:
Nope, makes sense to move that all into dockerland

surma (approved):

PR #791, opened 2020-08-03
Update build system for AVIF

surma:
- Switch to Makefile build
- Remove `package.json` and `napa`
- Rebuilds are not flakey anymore (after what I said on chat)
- Updated to libavif 0.8 and libaom 2.0
- Ran `clang-format`


codecs/cpp.Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM emscripten/emsdk:1.39.19
+FROM emscripten/emsdk:1.40.0
```
surma, line 1: I did update to Emscripten 1.40 because libavif doesn’t build with 1.39.19, but I haven’t re-built all the other codecs.

codecs/avif/Makefile
```diff
@@ -0,0 +1,75 @@
...
+	$(MAKE)
+
+$(CODEC_DIR)/.git/index:
+	mkdir -p $(ROOT_DIR) && \
+	cd $(ROOT_DIR) && \
+	git clone -b $(CODEC_VERSION) --depth 1 https://github.com/AOMediaCodec/libavif $(CODEC_DIR_RELATIVE)
+
+clean:
+	$(RM) $(OUT_JS) $(OUT_WASM)
+	$(RM) -rf $(LIBAOM_OUT)
```
RReverser: You don't need `-f`, it's part of `$(RM)`.

codecs/avif/Makefile
```diff
@@ -0,0 +1,75 @@
...
+		DCMAKE_BUILD_TYPE=Release \
+		-DAVIF_CODEC_AOM=1 \
+		-DAVIF_LOCAL_AOM=1 \
+		../ && \
+	$(MAKE)
+
+$(CODEC_DIR)/.git/index:
+	mkdir -p $(ROOT_DIR) && \
+	cd $(ROOT_DIR) && \
+	git clone -b $(CODEC_VERSION) --depth 1 https://github.com/AOMediaCodec/libavif $(CODEC_DIR_RELATIVE)
```
RReverser: Do we need the repo as an actual Git repo? If not, let's follow precedent from other codecs and just use CODEC_URL set to an archive from Github releases + curl + tar instead.

It should be both simpler to download and faster than git clone (which has to do extra work to download Git-specific objects for history and stuff).

codecs/avif/Makefile
```diff
@@ -0,0 +1,75 @@
...
+		-DCONFIG_INSPECTION=0 \
+		-DCONFIG_MULTITHREAD=0 \
+		-DCONFIG_RUNTIME_CPU_DETECT=0 \
+		-DCONFIG_WEBM_IO=0 \
+		../ && \
+	$(MAKE) 
+
+$(LIBAOM_DIR)/.git/index: $(CODEC_DIR)/.git/index
+	cd $(CODEC_DIR)/ext && \
+	git clone -b $(LIBAOM_VERSION) --depth 1 https://aomedia.googlesource.com/aom aom
```
RReverser: Same here about Git repo vs tar archive as below.

codecs/avif/Makefile
```diff
@@ -0,0 +1,75 @@
+ROOT_DIR = ./node_modules
+CODEC_DIR_RELATIVE = libavif
+CODEC_DIR = $(addprefix $(ROOT_DIR)/, $(CODEC_DIR_RELATIVE))
+CODEC_OUT_RELATIVE = build/libavif.a
+CODEC_OUT := $(addprefix $(CODEC_DIR)/, $(CODEC_OUT_RELATIVE))
+
+CODEC_VERSION = "v0.8.0"
+
+LIBAOM_RELATIVE = ext/aom/
```
RReverser: Let's avoid extra vars and `addprefix` where they're not necessary to keep it simple.

In this case, I suppose we won't have more than one folder, and `addprefix` is mainly useful for things like targets (where you might have >1 and want to add the same prefix to each of them). Instead, in this case it could be just

```
LIBAOM_DIR := $(CODEC_DIR)/ext/aom
```

Same for some other cases around.

codecs/avif/enc/avif_enc.cpp
```diff
@@ -29,34 +29,38 @@ val encode(std::string buffer, int width, int height, AvifOptions options) {
...
+  avifRGBImageAllocatePixels(&srcRGB);
 
   for (int y = 0; y < height; y++) {
     for (int x = 0; x < width; x++) {
       int pixelOffset = y * width + x;
-      image->rgbPlanes[0][pixelOffset] = rgba[pixelOffset * 4 + 0];
-      image->rgbPlanes[1][pixelOffset] = rgba[pixelOffset * 4 + 1];
-      image->rgbPlanes[2][pixelOffset] = rgba[pixelOffset * 4 + 2];
-      image->alphaPlane[pixelOffset] = rgba[pixelOffset * 4 + 3];
+      uint8_t* pixel = &srcRGB.pixels[(4 * x) + (srcRGB.rowBytes * y)];
```
RReverser: Not related to the build system, but it appears that this loop just copies bytes in original order without any reordering or recomputation? If so, can it be replaced entirely with `memcpy` or similar? Or, maybe, there is a method to create `avifRGBImage` from existing pointer to avoid copying data altogether?

codecs/cpp.Dockerfile
RReverser: Could you update & rebuild codecs in separate PR perhaps?

codecs/avif/enc/avif_enc.cpp
```diff
@@ -72,7 +76,9 @@ val encode(std::string buffer, int width, int height, AvifOptions options) {
   return val(typed_memory_view(output.size, output.data));
 }
 
-void free_result() { avifRWDataFree(&output); }
+void free_result() {
```
RReverser: Oh, I see there is `free_result` left. Are you planning to updating the memory management model in a separate PR or should it be done as part of this one?

codecs/avif/Makefile
RReverser: Actually, this should probably be replaced with `$(MAKE) clean` or similar, which knows better what temporary objects to remove and can parallelise such removal with other ones.

codecs/avif/enc/avif_enc.cpp
surma: I wasn’t sure if they word-align rows or something, so technically I think `rowBytes` could be greater than `width * 4`, but not 100% sure. I was mostly following the example code in the AVIF repo.

codecs/avif/Makefile
surma: > keep it simple

That ship has sailed :P

Okay, I’m happy to reduce number of variables.

codecs/avif/enc/avif_enc.cpp
surma: Will fix in this one.

codecs/avif/Makefile
surma: sg

codecs/cpp.Dockerfile
surma: That’s the plan

codecs/avif/Makefile
RReverser: It's not even that much number of variables that bothers me, as complexity of them layered with `addprefix` combinations to figure out the final value. I think simple concat should be enough for anything except target list :)

codecs/avif/enc/avif_enc.cpp
RReverser: I just found my similar comment from the original PR, looks like you should be able to assign pixels directly: https://github.com/GoogleChromeLabs/squoosh/pull/722#discussion_r419785905

Here's another relevant comment from AVIF.h: https://github.com/AOMediaCodec/libavif/blob/01ee4caa5ab52b2f6f3e500fd3a0f70e0099b031/include/avif/avif.h#L365-L370

codecs/avif/Makefile
```diff
@@ -0,0 +1,76 @@
...
+	$(MAKE)
+
+$(CODEC_DIR)/README.md:
+	mkdir -p $(CODEC_DIR) && \
+	cd $(CODEC_DIR) && \
+	curl -L $(CODEC_URL)/archive/$(CODEC_VERSION).tar.gz | tar -xzf - --strip 1 
+
+clean:
+	$(RM) $(OUT_JS) $(OUT_WASM)
+	cd $(CODEC_DIR)/$(CODEC_BUILD_DIR) && $(MAKE) clean
```
RReverser:
```suggestion
	$(MAKE) -C $(CODEC_DIR)/$(CODEC_BUILD_DIR) clean
```

codecs/avif/Makefile
```diff
@@ -0,0 +1,76 @@
...
+	cd $(LIBAOM_DIR)/$(LIBAOM_BUILD_DIR) && $(MAKE) clean
```
RReverser:
```suggestion
	$(MAKE) -C $(LIBAOM_DIR)/$(LIBAOM_BUILD_DIR) clean
```

codecs/avif/Makefile
```diff
@@ -0,0 +1,76 @@
...
+		-DENABLE_TESTS=0 \
+		-DCONFIG_ACCOUNTING=1 \
+		-DCONFIG_INSPECTION=0 \
+		-DCONFIG_MULTITHREAD=0 \
+		-DCONFIG_RUNTIME_CPU_DETECT=0 \
+		-DCONFIG_WEBM_IO=0 \
+		../ && \
+	$(MAKE) 
+
+$(LIBAOM_DIR)/README.md: $(CODEC_DIR)/README.md
```
RReverser: Nitpick: instead of README, I'd point to CMakeLists.txt, which is a build config and always guaranteed to be there (otherwise we can't build with existing commands).

codecs/avif/Makefile
```diff
@@ -0,0 +1,76 @@
...
+	mkdir -p $(LIBAOM_DIR) && \
+	cd $(LIBAOM_DIR) && \
+	curl -L $(LIBAOM_URL)/+archive/$(LIBAOM_VERSION).tar.gz | tar -xzf -
+
+$(CODEC_OUT): $(LIBAOM_OUT)
+	mkdir -p $(CODEC_DIR)/$(CODEC_BUILD_DIR) && \
+	cd $(CODEC_DIR)/$(CODEC_BUILD_DIR) && \
+	emcmake cmake \
+		DCMAKE_BUILD_TYPE=Release \
+		-DAVIF_CODEC_AOM=1 \
```
RReverser, line 63: Does this need similar `-DENABLE_DOCS=0` and so on or are they disabled by default here?

codecs/avif/Makefile
```diff
@@ -0,0 +1,76 @@
...
+		-DCONFIG_ACCOUNTING=1 \
+		-DCONFIG_INSPECTION=0 \
+		-DCONFIG_MULTITHREAD=0 \
+		-DCONFIG_RUNTIME_CPU_DETECT=0 \
+		-DCONFIG_WEBM_IO=0 \
+		../ && \
+	$(MAKE) 
+
+$(LIBAOM_DIR)/README.md: $(CODEC_DIR)/README.md
+	mkdir -p $(LIBAOM_DIR) && \
```
RReverser: Nitpick: this doesn't have to be part of the `&&`, it works as a separate command.

Perhaps worth copy-pasting https://github.com/GoogleChromeLabs/squoosh/blob/87955ab9a03401211ef37964cdc60e95a9a96dd9/codecs/webp/Makefile#L50-L52 for consistency between Makefiles? It's also a bit simpler in that output folder is specified directly in `tar` command.

codecs/avif/Makefile
RReverser: Actually, just using folder name should work in this case too (since there are no sources or timestamps involved).

codecs/avif/Makefile
surma: Don’t _think_ so, I was following the docs in the README

codecs/avif/Makefile
surma: It didn’t work last time I tried 🤷 
edit: As in, it would try to re-download on rebuild.

codecs/avif/Makefile
RReverser: You just mentioned earlier you had to build a bunch of examples, and noticed that they don't seem to be disabled in either of the codecs configs.

surma: Ah no, the examples are in libaom and still get built :S

RReverser: Looks like there is indeed `ENABLE_EXAMPLES`, in the spirit of other similar options, that you can disable: https://aomedia.googlesource.com/aom/+/refs/heads/master/CMakeLists.txt#227

codecs/avif/enc/avif_enc.cpp
```diff
@@ -49,17 +49,7 @@ val encode(std::string buffer, int width, int height, AvifOptions options) {
...
-    for (int x = 0; x < width; x++) {
-      int pixelOffset = y * width + x;
-      uint8_t* pixel = &srcRGB.pixels[(4 * x) + (srcRGB.rowBytes * y)];
-      pixel[0] = rgba[pixelOffset * 4 + 0];
-      pixel[1] = rgba[pixelOffset * 4 + 1];
-      pixel[2] = rgba[pixelOffset * 4 + 2];
-      pixel[3] = rgba[pixelOffset * 4 + 3];
-    }
-  }
+  memcpy(srcRGB.pixels, rgba, width * height * 4);
```
RReverser: This looks better. What I've meant by my last suggestion is that, as the next step after `memcpy`, you can just set `pixels` field to the original pointer - then you don't need to allocate and copy any data at all, and instead AVIF will perform conversion directly from provided source into the YUV format.

One more thing I've noticed is that you call `avifRGBImageAllocatePixels` but not `avifRGBImageFreePixels`. As far as I can tell from a quick glance, this will currently cause a memory leak for the allocated pixels.

It can be fixed by calling the `*Free*` counterpart, but, better yet, if you set `pixels` field as described above, then you can remove `avifRGBImageAllocatePixels` line and this issue will go away too, making code both faster and safer :)

I'd probably recommend trying this out in Squoosh with sanitizer enabled once we are done, but this can be left as part of the original PR.

RReverser:
> * Remove `package.json`

Hmm do we want this? I've kept it in other places, assuming that we want to be able to do `npm run build` in arbitrary folder instead of fiddling with Docker options in the command-line.

codecs/avif/enc/avif_enc.cpp
RReverser: I suspect you'll need to free `output` too, like the official example does, using `avifRWDataFree`.

RReverser: You'll probably also want to perform free/destroy above the `encodeResult` condition, otherwise those structures will get leaked on encoding errors.

surma: > What I've meant by my last suggestion is that, as the next step after memcpy, you can just set pixels field to the original pointer

I tried using `rgba` directly, but it made the codec crash. My hunch is that `avifRGBImageAllocatePixels` does more under the hood. I’ll add the `*Free*` counterpart.

surma:
I mean, the `package.json` doesn’t do much more than calling `../build-cpp.sh`. But yeah, might be nice to use it as “documentation”. I’ll re-add it.

codecs/avif/enc/avif_enc.cpp
RReverser: > My hunch is that `avifRGBImageAllocatePixels` does more under the hood.

Looking at the source, the only thing it does is setting `rowBytes` (which you can set manually).

And option of supplying your own `pixels` is documented as allowed in README, too: https://github.com/AOMediaCodec/libavif/blob/f41bcea5c6ec1a8b1f28833e878ced4468144a79/README.md#L158

RReverser: However, if you want to, you can leave this for later, I'm happy to play with it myself if it's not something of interest to you :)

surma: OOoh, I never set `rowBytes`. I’ll try doing that.

surma: Yeah that was it 🤦

surma:
I’ll merge this tomorrow if there are no other remarks (and CI passes)

codecs/avif/enc/avif_enc.cpp
surma: I like that GitHub never bothered to show me your comments until I reloaded.

RReverser:
The open conversation with `ENABLE_EXAMPLES` remark still stands. It doesn't cost anything to disable it, and seems relevant to the "build system" PR.

surma:
Yeah, I saw that comment via email, because GH’s UI is horrible. Testing that flag now…

surma:
The flag worked! Nice find!

RReverser:
Glad to hear!

Overall, I think I'd remove few more intermediate variables, but if you find it cleaner that way, I don't have any strong opinion and this should be good to go :)

I've moved non-build-system related comments to the original PR.

PR #792, opened 2020-08-04
Upload Batch images or Queue

thiagolead:
Upload more then one image per time


jakearchibald:
@thiagolead it looks like you're trying to merge some of my old commits from 2018 into the main branch. Was that your intention?

jakearchibald:
I'm going to assume not

PR #793, opened 2020-08-04
Added File Extension Check

ikeyurp:
Added file extension verification before uploading files Users will not be allowed to upload any other file types except proposed in this PR. This will help to filter uploaded files.


surma:
I don’t think this PR is sufficient or even necessary.

1. We support more than those file extensions and we use meta data sniffing instead, because extensions can be wrong, content cannot.
2. How does this reduce server load?

jakearchibald:
@bhutkeyur Files aren't uploaded anywhere. All processing is done locally on the user's machine.

See https://github.com/GoogleChromeLabs/squoosh/blob/dev/src/components/compress/index.tsx#L415 where we handle new files.

PR #797, opened 2020-08-12
Parcel 2

developit:
HMR is mostly working, workers and wasm are working, css modules is working.

I haven't done much testing of the prod build, and there's currently no prerendering or critical CSS.


.babelrc
```diff
@@ -0,0 +1,32 @@
+{
+  "presets": [
```
mischnic, line 2: You probably want to add `@parcel/babel-preset-env` in here, so that the browserslist in package.json is applied

developit: Ah good point, I forgot that defining the babelrc overrode that.

src/codecs/decoders.ts
```diff
@@ -1,6 +1,6 @@
 import { builtinDecode, sniffMimeType, canDecodeImage } from '../lib/util';
 import Processor from './processor';
-import webpDataUrl from 'url-loader!./tiny.webp';
+import webpDataUrl from 'url:./tiny.webp';
```
jakearchibald, line 3: Does this give you a data url?

developit: whoops - nope. I'll switch it to `data-url`.

PR #800, opened 2020-08-21
Add MozJPEG Decoder

surma:
We’ll need this for the CLI. 

(Also removes `version()` from the wasm module)


codecs/mozjpeg/dec/mozjpeg_dec.cpp
```diff
@@ -0,0 +1,69 @@
...
+  jpeg_finish_decompress(&cinfo);
+
+  /* Step 7: release JPEG compression object */
+
+  auto data = Uint8ClampedArray.new_(typed_memory_view(output_size, output_buffer));
+  auto js_result = ImageData.new_(data, cinfo.output_width, cinfo.output_height);
+
+  /* This is an important step since it will release a good deal of memory. */
+  jpeg_destroy_decompress(&cinfo);
+  delete[] output_buffer;
```
RReverser: Let's use C++isms instead of manual allocation where possible to reduce surface of bugs. See `std::vector<uint8_t>` example in other codecs.

codecs/mozjpeg/dec/mozjpeg_dec.cpp
```diff
@@ -0,0 +1,69 @@
...
+  // Prepare output buffer
+  size_t output_size = cinfo.output_width * cinfo.output_height * 4;
+  auto output_buffer = new uint8_t[output_size];
+  uint8_t* row_ptr = &output_buffer[0];
+  auto stride = cinfo.output_width * 4;
+
+  /* Process data */
+  while (cinfo.output_scanline < cinfo.output_height) {
+    jpeg_read_scanlines(&cinfo, &row_ptr, 1);
+    // Turn that row from RGB into RGBA. I’m starting at the end so
```
RReverser: Looks like there is `JCS_EXT_RGBA` that would do this natively, let's use that instead.

codecs/mozjpeg/dec/mozjpeg_dec.cpp
```diff
@@ -0,0 +1,69 @@
...
+
+using namespace emscripten;
+
+thread_local const val Uint8ClampedArray = val::global("Uint8ClampedArray");
+thread_local const val ImageData = val::global("ImageData");
+
+val decode(std::string image_in) {
+  uint8_t* image_buffer = (uint8_t*)image_in.c_str();
+
+  struct jpeg_decompress_struct cinfo;
```
RReverser: Nit: you can omit `struct ` keyword in C++ (like I did in encoder), but up to you.

surma: That was copy/paste’d from the sample code :3

codecs/mozjpeg/dec/mozjpeg_dec.cpp
surma: Good shout.

codecs/mozjpeg/dec/mozjpeg_dec.cpp
RReverser: Yeah I understand :)

jakearchibald:
Fwiw, I have some interest in adding this to Squoosh, for arithmetic-encoded JPEGs.

codecs/mozjpeg/dec/mozjpeg_dec.cpp
```diff
@@ -15,52 +15,40 @@ thread_local const val ImageData = val::global("ImageData");
...
-      // RGB channels
-      for (auto i = 0; i < 3; i++) {
-        row_ptr[x * 4 + i] = row_ptr[x * 3 + i];
-      }
-      // Alpha channel
-      row_ptr[x * 4 + 3] = 255;
-    }
-    row_ptr += stride;
+    uint8_t* ptr = &output_buffer[stride * cinfo.output_scanline];
+    jpeg_read_scanlines(&cinfo, &ptr, 1);
```
RReverser: I see that the last parameter to `jpeg_read_scanlines` is documented as number of lines to read, top-to-bottom. Since our output also has top-to-bottom layout, apparently we should be able to just pass total number of lines to this function and call it once instead of in a loop. If that works, it would be both simpler and more efficient.

codecs/mozjpeg/dec/mozjpeg_dec.cpp
```diff
@@ -15,52 +15,40 @@ thread_local const val ImageData = val::global("ImageData");
...
   }
   jpeg_finish_decompress(&cinfo);
 
-  /* Step 7: release JPEG compression object */
+  // Step 7: release JPEG compression object
 
-  auto data = Uint8ClampedArray.new_(typed_memory_view(output_size, output_buffer));
+  auto data = Uint8ClampedArray.new_(typed_memory_view(output_size, &output_buffer[0]));
```
RReverser: Nit / seems a bit more self-explanatory:

```suggestion
  auto data = Uint8ClampedArray.new_(typed_memory_view(output_buffer.size(), output_buffer.data()));
```

codecs/mozjpeg/dec/mozjpeg_dec.cpp
surma: I actually meant to change that, but wanted to go step-by-step. I’ll do this now (although apparently it never does it all in one step anyway, so we’d need to keep the loop).

RReverser: Oh, weird. In which cases it won't do that? I thought that only applies to something like streaming and not memory sources.

codecs/mozjpeg/dec/mozjpeg_dec.cpp
```diff
@@ -36,7 +36,7 @@ val decode(std::string image_in) {
   // Process data
   while (cinfo.output_scanline < cinfo.output_height) {
     uint8_t* ptr = &output_buffer[stride * cinfo.output_scanline];
-    jpeg_read_scanlines(&cinfo, &ptr, 1);
+    jpeg_read_scanlines(&cinfo, &ptr, cinfo.output_height);
```
RReverser: If this can be reached multiple times, shouldn't it be containing "how many lines are left" as opposed to total number of lines all the time?

If it's getting too complicated, I'm okay with just reverting to `1` for now and addressing this separately.

RReverser (approved):

PR #804, opened 2020-08-25
AVIF fixes and options

jakearchibald:


codecs/avif/enc/avif_enc.cpp
```diff
@@ -32,18 +35,33 @@ val encode(std::string buffer, int width, int height, AvifOptions options) {
...
   avifImage* image = avifImageCreate(width, height, depth, format);
 
+  if (
+    options.maxQuantizer == AVIF_QUANTIZER_LOSSLESS &&
+    options.minQuantizer == AVIF_QUANTIZER_LOSSLESS &&
+    options.minQuantizerAlpha == AVIF_QUANTIZER_LOSSLESS &&
+    options.maxQuantizerAlpha == AVIF_QUANTIZER_LOSSLESS &&
+    format == AVIF_PIXEL_FORMAT_YUV444
+  ) {
+    image->matrixCoefficients = AVIF_MATRIX_COEFFICIENTS_IDENTITY;
```
jakearchibald, line 60: The reference encoder seems to think this is VERY IMPORTANT

codecs/avif/enc/avif_enc.cpp
```diff
@@ -32,18 +35,33 @@ val encode(std::string buffer, int width, int height, AvifOptions options) {
...
+  } else {
+    image->matrixCoefficients = AVIF_MATRIX_COEFFICIENTS_BT709;
```
jakearchibald, line 62: Chrome's decoder gets the colours wrong unless we do this.

src/codecs/avif/encoder-meta.ts
```diff
@@ -12,12 +14,14 @@ export const label = 'AVIF';
 export const mimeType = 'image/avif';
 export const extension = 'avif';
 export const defaultOptions: EncodeOptions = {
-  minQuantizer: 16,
-  maxQuantizer: 16,
+  minQuantizer: 33,
+  maxQuantizer: 63,
```
jakearchibald, line 18: I invented these defaults by looking for numbers that seem to match the quality of the default mozjpeg options.

src/codecs/avif/encoder-meta.ts
```diff
@@ -12,12 +14,14 @@ export const label = 'AVIF';
...
+  minQuantizerAlpha: 33,
+  maxQuantizerAlpha: 63,
```
jakearchibald, line 20: Lossy alpha by default anyone? I'm undecided if this is the right choice.

src/codecs/avif/encoder-meta.ts
```diff
@@ -12,12 +14,14 @@ export const label = 'AVIF';
...
   tileColsLog2: 0,
   tileRowsLog2: 0,
-  speed: 10,
-  subsample: 0,
+  speed: 8,
```
jakearchibald, line 23: This is the default in the reference encoder and the speed seems reasonable. 7 brings significant improvements but is significantly slower.

src/codecs/avif/options.tsx
```diff
@@ -1,72 +1,304 @@
...
+  losslessAlpha: boolean;
+  maxAlphaQuality: number;
+  minAlphaQuality: number;
+  showAdvanced: boolean;
+  grayscale: boolean;
+  subsample: number;
+  tileRows: number;
+  tileCols: number;
+  effort: number;
+}
```
jakearchibald, line 31: I went in a slightly different direction with these options compared to others.

Other options try to set the form state from the incoming encoder options, but that gets really complicated in cases where the form represents things slightly differently. For example, here we have 'lossless' as a toggle, but toggling that shouldn't change the quality slider, just the output values. Also, if toggling one option results in a form field being removed, code has to be defensive about that.

Instead, I've kept the form state and encoder options separate. Encoder options are converted to state whenever they've changed from the outside, in `getDerivedStateFromProps`. Form options are converted to encoder options in `_inputChange`.

@developit is this a reasonable use of `getDerivedStateFromProps`?

src/codecs/decoders.ts
```diff
@@ -1,14 +1,16 @@
 import { builtinDecode, sniffMimeType, canDecodeImage } from '../lib/util';
 import Processor from './processor';
 import webpDataUrl from 'url-loader!./tiny.webp';
+import avifDataUrl from 'url-loader!./tiny.avif';
```
jakearchibald: This is over 300 bytes 😢 

I just took a 1x1 black pixel and played with settings until it was as small as possible. There might be better ways.

src/codecs/avif/encoder-meta.ts
jakearchibald: Actually, that `maxQuantizer` looks pretty ugly on the illustration example.

jakearchibald:
Pretty happy for this, except that 'grayscale' produces a file that cannot be decoded, and '4:2:2' produces a file that our decoder can't decode (but Chrome can). @surma @RReverser any ideas?

However, I'd be happy launching with those features removed.

The encoder and decoder are pretty beefy (700k each), but I guess there's nothing we can do about that?

src/codecs/avif/encoder-meta.ts
surma: I’m okay with this.

src/codecs/decoders.ts
RReverser: I remember @kornelski mentioning on Twitter that Chrome currently requires too much metadata on AVIF images, whereas Firefox is more lax. Probably the case here as well - we could remove some more metadata, but it wouldn't render in Chrome?

surma (approved):
I wonder if grayscale fails because this

https://github.com/GoogleChromeLabs/squoosh/blob/4e5a8107704aa66873f3f01d8348285f69fde069/codecs/avif/dec/avif_dec.cpp#L32

I’d say we launch as-is as a “AVIF beta” (modulo review things) and get back to the grayscale issue throughout v2.

Re file size: At this point not much we can do, no. Although it could be interesting how much bigger we get if we put encoder and decoder into _one_ Wasm module, at least until support has shipped more widely.

src/codecs/decoders.ts
jakearchibald: Yeah there's https://bugs.chromium.org/p/chromium/issues/detail?id=1120973. But yeah, rendering in Chrome is pretty important here 😄

codecs/avif/enc/avif_enc.cpp
kornelski: Yeah, that stores image as RGB instead of YUV, so you don't get loss from colorspace conversion.

codecs/avif/enc/avif_enc.cpp
kornelski: BT709 is compatible with sRGB. Anything else is asking for trouble.

src/codecs/avif/encoder-meta.ts
kornelski: I think you're supposed to set them both to the same value. min/max makes sense only for video with variable rate encoding.

src/codecs/decoders.ts
kornelski: 300 bytes sounds about right for AVIF. The header is like Oscars speech: "shout out to MPEG-4, and my cousins from the HEIF spec. And big thanks to ISO BMFF, and a chunk of love for MIAF spec, and let's not forget video manipulation tools and their backwards compatibility layers".

Alpha channel is supported by adding "auxiliary image" attachment with its own chunk that contains an index of locations where it has bytes in the file, and duplicate metadata that re-states its an AV1 image, and a chunk that copies AV1 encoding parameters in case you want to know them without looking 200 bytes further in the file (this is mandatory and Chrome enforces that it's present), and a copy of image size just to be clear. And then you need to add auxiliary image reference chunk that says auxiliary image is related to the primary image. And then add a properties chunk with properties container and properties association chunks that give the auxiliary image an auxiliary property that contains, and I'm not kidding, *a URN*, equal to "urn:mpeg:mpegB:cicp:systems:auxiliary:alpha".

Other codes have a bit in header that says alpha=1.

src/codecs/processor-worker/index.ts
```diff
@@ -88,14 +88,14 @@ async function avifEncode(
   const { encode } = await import(
     /* webpackChunkName: "process-avif-enc" */
     '../avif/encoder');
-  return encode(data, options);
+  return timed('avifEncode', () => encode(data, options));
```
RReverser, line 91: Oh, good catch!

jakearchibald:
Spotted a bug: If I enable 'separate alpha', then adjust the sliders so they match, it disables 'separate alpha', which is exactly the thing I was trying to avoid 😄 

Yeah ok it's hitting the derive path when it shouldn't.

Fixed.

src/codecs/decoders.ts
RReverser: @jakearchibald Cross-linking a potential alternative here: https://twitter.com/RReverser/status/1298346612067098627

RReverser: Btw, have you tried an empty (0x0) image instead? I imagine it won't help too much if the sample is dominated by metadata, but maybe it will cut size down at least a bit?

src/codecs/avif/encoder-meta.ts
jakearchibald: I thought the same - I'd only ever seen it in target bitrate encoding, however it seems to do something here. https://5f4512e9cdb7030008747032--squoosh.netlify.app/ - try the "artwork" example image. The low min setting causes a lot of smoothing. It made me wonder if min/max was similar to SNS in WebP. I'll do some research though.

src/codecs/decoders.ts
jakearchibald: @kornelski does that mean it can't render an image with alpha progressively? Since required data is stuffed at the end?

jakearchibald: Ah, I see you're already pushing on this https://github.com/AOMediaCodec/libavif/issues/287

jakearchibald:
I've committed my manually built smaller wasms for AVIF. https://github.com/GoogleChromeLabs/squoosh/issues/805 is filed to fix it up properly.

I reverted all the mozjpeg stuff, since it significantly increased the encoder wasm size, but I kept the minor webp changes from the v2 codecs branch. Is that ok @surma?

I think this is good to go. I'd still like @developit to review, but it doesn't need to block launch.

PR #813, opened 2020-08-27
Fix #346 Add keyboard shortcuts for moving the split view separator

### LINKED ISSUE #346: Add better interaction for split bar
Splitting from issue #270

-------

Doubleclicking a separator should move it between three positions: left view, right view and middle point. There should be a keyboard shortcut for switching between first two. And maybe another one to return to middle (split) view. Rationale for one shortcut to not go through all three states, but only two is to have an unambiguous and quick way to see before/after fullscreen without being distracted by the split view.

samal-rasmussen:
Fix #346.

I used the number keys 1, 2, and 3.

I would have used the arrow keys. Seems natural that pressing left and right would move the separator to the left and right, and pressing the middle up and down buttons moves it to the center. But the arrow keys are already in use for navigating the interface, so I had to come up with something else.

I also made an overlay/modal with some info about the shortcuts so that they are discoverable. I would have used the url router and a separate page for this info, but I couldn't find any router :sweat_smile:

This is my first time writing any preact or react code at all, so I'm not sure at all how idiomatically I've done things. Feel free to make any suggestions.

### DIFF
--- a/src/client/lazy-app/Compress/Output/custom-els/TwoUp/index.ts
+++ b/src/client/lazy-app/Compress/Output/custom-els/TwoUp/index.ts
@@ -68,6 +68,8 @@ export default class TwoUp extends HTMLElement {
         );
       },
     });
+
+    window.addEventListener('keydown', event => this._onKeyDown(event));
   }
 
   connectedCallback() {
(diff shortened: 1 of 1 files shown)


surma:
@samal-rasmussen I’m sorry we never got around to reviewing this. We have been quite busy preparing for Squoosh v2, which is now live.

At a glance, your PR looks really good. Would you mind reworking this for the new `dev` branch? We changed the folder structure quite a bit, but most of the UI components stayed somewhat the same, so I am hoping it shouldn’t be too much work.  There’s some minor things to potentially address (for example should pressing `1` move the slider to the left or show the left image?), but we can get to that if and when you update the PR.

samal-rasmussen:
Updated.

I removed the modal with the documentation from the front page, as it wasn't behaving properly, and I don't think it's the right place for that info either way. Help should be more contextual, it should be available in the same place as the functionality it describes, and that is inside the image viewer. But I'm not sure where to place it there. I'm thinking maybe a question mark that you can press somewhere inside the image viewer? But this is a separate feature of it's own...

> There’s some minor things to potentially address (for example should pressing 1 move the slider to the left or show the left image?), but we can get to that if and when you update the PR.

I think there is minimal mental overhead for the user, when the slider moves with the key press. The connection between the key press and slider moving is visually obvious and so you don't have to think.

surma (approved):
This LGTM, but gonna wait for another LGTM from @jakearchibald

surma:
Let’s merge this 🎉  It is a small PR that doesn’t change any UI and doesn’t break any behavior. If others have concerns we can revert and revisit. Thanks for seeing this through, @samal-rasmussen!

PR #829, opened 2020-09-28
Multithread AVIF & OxiPNG

RReverser:
This adds support for multithreaded C++ and Rust codecs on Wasm, with AVIF and OxiPNG used as reference examples.

Some rough, non-scientific, numbers from my machine on preheated (Turbofanned) Wasm with default codec settings, on Device screen demo:
 - OxiPNG: 7560ms -> 2930ms (-61%)
 - AVIF: 5533ms -> 3932ms (-29%)

In general, OxiPNG is much more parallelisable because all it does is tries various independent combinations, while AVIF has some shared state & computations, but either way the improvements seem nice.

Multithreaded versions are built alongside single-threaded once, and then wasm-feature-detect is used for Wasm+threads feature detection and appropriate version is loaded dynamically.

I've also added COOP+COEP headers to make this work in browsers other than desktop Chrome (e.g. confirmed it works in Firefox).

Note: there seems to be a race condition bug in Chrome Canary that will sometimes trip at least OxiPNG, but potentially other codecs using threads too. This bug doesn't appear on stable Chrome so I think there's no point in blocking on it.


RReverser:
(I'll clean up commit history after reviews)

surma:
Nice! Gonna review later. 

The hunch was that AVIF can parallelize much more if you tile the image under advanced settings, so I did a a quick manual measurement (DevTools closed, waiting sufficiently long to make sure TurboFan kicked in, using a stop watch lol):

- Red fox in prod Squoosh, Min/Max Quantizer = 30: **42s**
- Red fox in multithreaded Squoosh, Min/Max Quantizer = 30, 8x8 tiling (both tiling options set to 3): **21s**

Literally 50% faster 🎉

RReverser:
@surma Very nice! I tried to use default settings for the results to be a bit more representative of a typical user (but also because I would be too lazy to change settings those hundreds of times I did the testing, lol).

Do you think it's worth changing the defaults as part of this PR to enable tiling? Are there any downsides (in terms of size or anything else) I'd we do that?

surma:
> Do you think it's worth changing the defaults as part of this PR to enable tiling

I think it would be interesting to enable tiling (maybe 2x2) by default _for people with thread support_. Although, yes, it’s worth investigating how big of a file size penalty tiling brings.

RReverser:
OTOH, Squoosh already produces different results (in terms of the output size) on different browsers, so maybe nobody will notice 😅

codecs/oxipng/Cargo.toml
```diff
@@ -12,7 +12,19 @@ crate-type = ["cdylib"]
 oxipng = { version = "3.0.0", default-features = false }
 wasm-bindgen = "0.2.64"
 log = { version = "0.4", features = ["release_max_level_off"] }
+rayon = { version = "1.3.0", optional = true }
+once_cell = { version = "1.3.1", optional = true }
+crossbeam-channel = { version = "0.4.2", optional = true }
+
+[patch.crates-io]
+oxipng = { git = "https://github.com/RReverser/oxipng", branch = "crossbeam" }
```
surma, line 21: Oh dear 😅  Why is this necessary? Are we going to maintain a fork long-term now?

codecs/oxipng/Cargo.toml
```diff
@@ -12,7 +12,19 @@ crate-type = ["cdylib"]
...
 
 [profile.release]
 lto = true
 opt-level = "s"
+
+[features]
+parallel = ["oxipng/parallel", "rayon", "crossbeam-channel", "once_cell"]
+
+[package.metadata.wasm-pack.profile.release]
+wasm-opt = ["-O", "--no-validation"]
```
surma, line 31: Why `--no-validation`? Isn’t `--enable-bulk-memory --enable-threads` doing the trick?

codecs/oxipng/build.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+set -e
+
+rm -rf pkg,{-parallel}
+wasm-pack build
+RUSTFLAGS='-C target-feature=+atomics,+bulk-memory' wasm-pack build -t web -d pkg-parallel -- -Z build-std=panic_abort,std --features=parallel
+sed -i "s|input = import.meta.url.replace(/\\\.js$/, '_bg.wasm');||" pkg{,-parallel}/squoosh_oxipng.js
```
surma, line 8: What is happening here?

package.json
```diff
@@ -73,5 +73,8 @@
     "webpack-cli": "3.3.4",
     "webpack-dev-server": "3.8.0",
     "worker-plugin": "3.1.0"
+  },
+  "dependencies": {
+    "wasm-feature-detect": "^1.2.6"
```
surma: move this to `devDependencies`

src/missing-types.d.ts
```diff
@@ -33,6 +33,20 @@ declare module 'url-loader!*' {
...
+  const value: string;
+  export default value;
+}
+
+declare module '*.worker.js' {
+  const value: string;
+  export default value;
+}
+
+declare module 'wasm-feature-detect' {
```
surma: This can be removed after updating `wasm-feature-detect`, right?

webpack.config.js
```diff
@@ -149,7 +149,7 @@ module.exports = async function(_, env) {
         {
           // Emscripten modules don't work with Webpack's Wasm loader.
           test: /\.wasm$/,
-          exclude: /_bg\.wasm$/,
+          exclude: /(?<!\bpkg-parallel[\\/]squoosh_oxipng)_bg\.wasm$/,
```
surma, line 152: I find this bit confusing. Are we re-enabling Webpack’s Wasm loader for those two modules? If so, why?

surma (changes requested):
The AVIF stuff is nice and clean, super happy with that.

The Rust stuff is a bit… messy? I guess that reflects the current state of the Rust toolchain when it comes to Wasm threads. I wonder if there is an opportunity to open upstream bugs with some of this.

Couple of Qs and comments on the diff :)

codecs/oxipng/Cargo.toml
RReverser: No, I'll upstream changes later, but for now wanted to have flexibility. It's not the first time we're pointing to my fork, did the same in the initial version of OxiPNG while waiting for upstream :)

codecs/oxipng/Cargo.toml
RReverser: Something kept failing for me, but perhaps I didn't try `--enable-bulk-memory`. Personally, I don't think validation is that important when Wasm is produced by a legitimate compiler, it's mostly useful for cases when you try to build it by hand.

package.json
RReverser: How so? This is a runtime dependency.

src/missing-types.d.ts
RReverser: Yes.

webpack.config.js
RReverser: I agree it is confusing. I could clean it up if you want, but then you said we're migrating from Webpack anyway and there's no point in investing too much time to do it in a clean way?

codecs/oxipng/build.sh
RReverser: I'm removing an autogenerated line "input = import.meta.url.replace(/\\\.js$/, '_bg.wasm');" to work around an issue in Webpack (it doesn't like / recognise `import.meta` syntax). There are other ways to do this, but this one requires the least effort, and we hopefully won't have the same problem with other toolchains anyway.

RReverser:
As one more update, I've fixed `EXPORT_ES6` + `pthread` upstream, and tried it on AVIF codec, but it currently hits the same missing Webpack support for `import.meta` and fails to compile; I could make it work with more hacks, but for now prefer to keep `importScripts` and wait for our switch to a different bundler.

package.json
surma: This is off the back of a longer debate, but we have all our dependencies in `devDependencies`.

src/missing-types.d.ts
surma: Please make a release of `wasm-feature-detect` and update this :)

webpack.config.js
surma: Fair.

codecs/oxipng/build.sh
surma: Okay, let’s ignore this you are correctly pointing out that we are moving to Rollup anyway.

codecs/oxipng/Cargo.toml
surma: Alright.

codecs/oxipng/Cargo.toml
surma: Can you add a comment with the upstream issue/PR so we know when we can replace the fork with the original again?

surma (approved):
LGTM modulo open TODOs

package.json
jakearchibald: > This is a runtime dependency.

Is buildtime also runtime?

RReverser: > Is buildtime also runtime?

I'd suggest not, at least I prefer to separate "this is a build tool / a plugin for the build tool" vs "this is a runtime library that could as well be requested from external storage".

But I don't really want to get into a longer debate atm, happy to update to follow the existing precedent.

codecs/oxipng/Cargo.toml
RReverser: I don't have one ready yet, didn't get to the "upstreaming" part of this feature.

src/missing-types.d.ts
RReverser: Done.

package.json
RReverser: Done.

codecs/oxipng/Cargo.toml
RReverser: Done - created PR [here](https://github.com/shssoichiro/oxipng/pull/327) and added as a comment.

PR #835, opened 2020-10-29
Update JXL to latest and do proper color management.

veluca93:
The JXL decoder now does colorspace conversion to sRGB correctly,
independently of the color space that the JPEG XL file was in.


codecs/jxl/dec/jxl_dec.cpp
```diff
@@ -1,31 +1,77 @@
...
   auto next_in = (const uint8_t*)data.c_str();
   auto avail_in = data.size();
-  EXPECT_EQ(JPEGXL_DEC_BASIC_INFO, JpegxlDecoderProcessInput(dec, &next_in, &avail_in));
+  EXPECT_EQ(JXL_DEC_BASIC_INFO, JxlDecoderProcessInput(dec, &next_in, &avail_in));
   size_t buffer_size;
-  const JpegxlPixelFormat format = {4, JPEGXL_LITTLE_ENDIAN, JPEGXL_TYPE_UINT8};
-  EXPECT_EQ(JPEGXL_DEC_SUCCESS, JpegxlDecoderImageOutBufferSize(dec, &format, &buffer_size));
-  JpegxlBasicInfo info;
-  EXPECT_EQ(JPEGXL_DEC_SUCCESS, JpegxlDecoderGetBasicInfo(dec, &info));
+  const JxlPixelFormat format = {4, JXL_LITTLE_ENDIAN, JXL_TYPE_UINT8};
```
lvandeve: Using output type uint8 will cause clamping if the image was not originally sRGB, is this intended? If not, float is better. uint16 will also be clamped

veluca93: Squoosh uses uint8 anyway (possibly because that's what Canvas expects?), so I don't think I can do this properly with floats.
Floats were also causing me an out-of-memory error (or a segfault), which is at least a bit problematic.

lvandeve: you're using skcms first though, right?

the correct method is:

decode to float --> convert with skcms to the correct color space --> convert to uint8 or uint16 and clamp

the clamping should not be done before the conversion with skcms (it throws away colors).

looks like skcms can take a different pixel input format and output format, so it should be able to take float input, and output clamped integers

will that work?

surma: The web platform — as of now — only supports `uint8` for [`ImageData`](https://developer.mozilla.org/en-US/docs/Web/API/ImageData). We have to figure something out here eventually if we want to support HDR, but for now, yes, we are clamping (or scaling) to fit everything into `uint8`.

codecs/jxl/dec/jxl_dec.cpp
```diff
@@ -1,31 +1,77 @@
...
+  // space defined by the decoded ICC profile. Otherwise, it is in Linear sRGB. TODO: the decoded
+  // color profile should also be Linear sRGB if !uses_original_profile.
+  if (info.uses_original_profile) {
+    EXPECT_TRUE(skcms_Parse(icc_profile.data(), icc_profile.size(), &jxl_profile));
+  } else {
+    auto s = jxl::ColorEncoding::LinearSRGB(/*gray=*/false);
+    EXPECT_TRUE(s.CreateICC());
+    EXPECT_TRUE(skcms_Parse(s.ICC().data(), s.ICC().size(), &jxl_profile));
+  }
+  EXPECT_TRUE(skcms_Transform(
```
RReverser, line 69: Just to double-check - is it guaranteed to be safe to use same source & dest for pixel data?

RReverser:
Can you please include the new JS & Wasm generated by `npm run build`? (we commit those into the repo rather than building each time, and right now looks like the old ones got deleted in the PR but new ones are not there yet)

codecs/jxl/dec/jxl_dec.cpp
veluca93: Yes, the skcms doc explicitly states it's OK as long as it's the same pixel format.

veluca93:
> Can you please include the new JS & Wasm generated by `npm run build`? (we commit those into the repo rather than building each time, and right now looks like the old ones got deleted in the PR but new ones are not there yet)

I think they are included, but not shown in the diff (they did get quite a bit smaller though)

RReverser:
> they did get quite a bit smaller though

🤦🏻‍♂️ right, I mistook the red messages for deleted files on the first glance. Thanks!

RReverser:
As for the size - yeah, they're much smaller because previously I've pushed builds with debug info included to this branch to help with debugging the assertion issue.

codecs/jxl/dec/jxl_dec.cpp
```diff
@@ -1,31 +1,77 @@
 #include <emscripten/bind.h>
 #include <emscripten/val.h>
 
-#include <jpegxl/decode.h>
+#include <jxl/decode.h>
+#include "lib/jxl/color_encoding_internal.h"
```
RReverser, line 5: I'm assuming those will become public APIs over time? Seems a bit dangerous to rely on internals, but I suppose as long as we use a fixed commit SHA and don't attempt to upgrade, it's fine.

veluca93: It's not a public API, but it will be removed once the API functions are actually fixed to behave as intended :)

codecs/jxl/enc/jxl_enc.cpp
```diff
@@ -31,10 +32,11 @@ val encode(std::string image, int width, int height, JXLOptions options) {
 
   uint8_t* inBuffer = (uint8_t*)image.c_str();
 
-  auto result =
-      main->SetFromSRGB(width, height, /*is_gray*/ false, /*has_alpha*/ true,
-                        /*alpha_is_premultiplied*/ true, /*pixels*/ (uint8_t*)image.c_str(),
-                        /* end */ (uint8_t*)image.c_str() + width * height * 4);
+  auto result = jxl::ConvertImage(
```
RReverser, line 32: I see this new API accepts `bits_per_alpha` and `bits_per_sample` via params, does this mean we don't need the `io.metadata.Set*` lines above anymore?

Also, a more general question (not in scope of this PR) - should we generally switch to and use C API instead? (I'm assuming that one is public given that it's the only one exposed from `include/`)

codecs/jxl/dec/jxl_dec.cpp
veluca93: I'm using floats as intermediate values now - my previous attempt used the buffer size instead of the number of pixels, explaining the segmentation fault...

veluca93:
PTAL - I've added an -O3 flag to the Makefile, as it seems to have a quite dramatic effect on decoding performance (~30%). If that's not the best place to do it, let me know and I will move it.

RReverser:
> PTAL - I've added an -O3 flag to the Makefile, as it seems to have a quite dramatic effect on decoding performance (~30%).

Interesting. I think we'll need to compare sizes with the "proper" branch, but sounds reasonable - in this case the performance difference is definitely non-negligible.

However, you've put `-O3` in a place where it should affect only the top module and not the actual JPEG-XL library. Given that the top module is just a thin wrapper that calls into JPEG-XL, I'm surprised it has any effect, especially such a dramatic one.

How consistent was the perf difference?

codecs/jxl/enc/jxl_enc.cpp
veluca93: bits_per_sample is not needed, but AlphaBits is.

This is an internal API, but AFAIU there's no encoder C API yet - it's better to switch once it is.

veluca93:
> > PTAL - I've added an -O3 flag to the Makefile, as it seems to have a quite dramatic effect on decoding performance (~30%).
> 
> Interesting. I think we'll need to compare sizes with the "proper" branch, but sounds reasonable - in this case the performance difference is definitely non-negligible.
> 
> However, you've put `-O3` in a place where it should affect only the top module and not the actual JPEG-XL library. Given that the top module is just a thin wrapper that calls into JPEG-XL, I'm surprised it has any effect, especially such a dramatic one.
> 
> How consistent was the perf difference?

The perf difference is reasonably consistent. JPEG XL is already compiled with pretty aggressive optimizations, to the point where the skcms conversion actually takes about 50% of the decoder time - hence why the bigger-than-expected difference here.

RReverser:
Hmm but that conversion is handled by `skcms_Parse` and `skcms_Transform` which are also part of a separate object file and shouldn't be affected by the flag change here?

Anyway, I'm happy to try and play with optimisation levels further on my own time :)

codecs/jxl/enc/jxl_enc.cpp
RReverser: > but AlphaBits is

Interesting. What's the difference between `bits_per_alpha` param and `io.metadata.SetAlphaBits`?

> This is an internal API, but AFAIU there's no encoder C API yet - it's better to switch once it is.

Fair enough, thanks for the confirmation.

veluca93:
> Hmm but that conversion is handled by `skcms_Parse` and `skcms_Transform` which are also part of a separate object file and shouldn't be affected by the flag change here?
> 
> Anyway, I'm happy to try and play with optimisation levels further on my own time :)

I must have had some old build files lying around :) I'll leave this up to you (as well as choosing a memory limit)

codecs/jxl/enc/jxl_enc.cpp
veluca93: I don't think there's a difference per se, but I think we access the field before the API below sets it.

RReverser:
> I'll leave this up to you (as well as choosing a memory limit)

I wish we still got to the bottom of those OOM issues. 4GB is the most we can allocate for Wasm, and even that seems quite excessive.

veluca93:
The code now works with 1GB. We resolved a couple of memory usage
related issues recently, it might be enough to not require so much memory
anymore.

codecs/jxl/dec/jxl_dec.cpp
```diff
@@ -1,31 +1,82 @@
...
+    EXPECT_TRUE(s.CreateICC());
+    EXPECT_TRUE(skcms_Parse(s.ICC().data(), s.ICC().size(), &jxl_profile));
+  }
+  EXPECT_TRUE(skcms_Transform(
+      float_pixels.get(), skcms_PixelFormat_RGBA_ffff,
+      info.alpha_premultiplied ? skcms_AlphaFormat_PremulAsEncoded : skcms_AlphaFormat_Unpremul,
+      &jxl_profile, pixels.get(), skcms_PixelFormat_RGBA_8888, skcms_AlphaFormat_Unpremul,
+      skcms_sRGB_profile(), info.xsize * info.ysize));
+
+  float_pixels.reset();
```
RReverser:
```suggestion
```
We don't need to manually free this.

codecs/jxl/dec/jxl_dec.cpp
RReverser: I see you changed this in the last version - was this not ok after all?

RReverser: Ah, we **don't** use the same format (anymore?), so makes sense to change it.

veluca93: Yes, indeed with intermediate floats we can't use the same buffer.

veluca93:
This manual free was meant to reduce memory usage before the creation of
the ImageData array - if you say this is not a concern, then no reason to,
I suppose.

RReverser:
Ah I see. ImageData and its Uint8Array copy are created by the JS side, and so not subject to Wasm memory limits, so I think it's fine to skip.

PR #838, opened 2020-10-30
Switch to -O3 for C++ codecs

RReverser:
This makes certains codecs from 1.5x to 2.5x faster (depending on codec * file * options combination). The size difference is not as large, so it seems totally worth it.


jakearchibald (approved):
Trade-off sounds fine to me

PR #839, opened 2020-10-30
Upgrade LibAVIF & use new flags for building only AVIF encoder/decoder

RReverser:
See the discussion in https://github.com/AOMediaCodec/libavif/issues/254 where this was implemented.

This allows us to avoid using `ERROR_ON_UNDEFINED_SYMBOLS` and build truly separate encoder/decoder libs.


RReverser:
Note on the further changes above: also changed the build scripts to optimise the directory structure.

Before we were producing:
```
libavif-enc
  ...AVIF sources...
  build
    ...AVIF built with required params...
  ext
    aom
      ...AOM sources...
      libavif.build
        ...AOM built with required params...
libavif-enc-mt
  ...AVIF sources...
  build
    ...AVIF built with different params...
  ext
    aom
      ...AOM sources...
      libavif.build
        ...AOM built with different params...
libavif-dec
  ...AVIF sources...
  build
    ...AVIF built with different params...
  ext
    aom
      ...AOM sources...
      libavif.build
        ...AOM built with different params...
```

Now I changed the way AVIF looks for AOM and directories could be simplified to:
```
libavif
  ...AVIF sources...
libaom
  ...AOM sources...
build
  enc
    libavif
      ...AVIF built with required params...
    libaom
      ...AOM built with required params...
  enc-mt
    libavif
      ...AVIF built with different params...
    libaom
      ...AOM built with different params...
  dec
    libavif
      ...AVIF built with different params...
    libaom
      ...AOM built with different params...
```

That is, we no longer have to unpack multiple copies of same sources, but instead unpack them once and only produce different builds.

jakearchibald (approved):
LGTM

surma (approved):

PR #847, opened 2020-11-19
Fix Lossless+Progressive JXL and add debug options.

veluca93:
Setting CODEC_DEBUG in the JXL Makefile will produce a debug build (with
error messages enabled).


codecs/jxl/enc/jxl_enc.cpp
```diff
@@ -57,10 +57,13 @@ val encode(std::string image, int width, int height, JXLOptions options) {
...
   }
 
-  if (options.progressive) {
+  if (options.progressive && !cparams.modular_mode) {
     cparams.qprogressive_mode = true;
     cparams.progressive_dc = 1;
     cparams.responsive = 1;
+  } else {
+    cparams.responsive = 1;
+    cparams.progressive_mode = true;
```
jakearchibald: Isn't this setting progressive mode even if `options.progressive` is false?

veluca93: ... Yes, thanks for catching that...

jakearchibald:
@RReverser can you take a look at the makefile stuff here?

jakearchibald:
This should be merged into `dev`, but I'm happy to take care of that.

veluca93:
Rebased in #848

PR #848, opened 2020-11-19
Fix Lossless+Progressive JXL.

veluca93:
Rebased from #847


RReverser:
I think I'd prefer not to include the debug changes in Makefile. If you can run `make` directly, then you can pass necessary flags from the command line like `make CXXFLAGS="-g -DJXL_DEBUG_ON_ALL_ERROR=1 -s ASSERTIONS=1"` or any other set useful for a particular build.

My earlier suggestion to add things to Makefile was meant to be as a temporary quick measure to avoid having to pass those through to the Docker container.

veluca93:
> I think I'd prefer not to include the debug changes in Makefile. If you can run `make` directly, then you can pass necessary flags from the command line like `make CXXFLAGS="-g -DJXL_DEBUG_ON_ALL_ERROR=1 -s ASSERTIONS=1"` or any other set useful for a particular build.
> 
> My earlier suggestion to add things to Makefile was meant to be as a temporary quick measure to avoid having to pass those through to the Docker container.

I think there is some value in having an easy way to come up with a debug build (and I'm not sure CXXFLAGS would be passed to jxl).
If you have a strong opinion I can revert this though.

RReverser:
> and I'm not sure CXXFLAGS would be passed to jxl

Yeah it would be, CMake reads them from the environment (maybe you need `CXXFLAGS="..." make` instead of what I wrote though).

veluca93:
I reverted the DEBUG mode changes.

codecs/jxl/Makefile
```diff
@@ -11,7 +11,7 @@ OUT_WASM = $(OUT_JS:.js=.wasm)
 
 all: $(OUT_JS)
 
-%.js: %.cpp $(LIBAOM_OUT) $(CODEC_OUT)
+%.js: %.cpp $(CODEC_OUT) Makefile
```
RReverser: Since $(CODEC_OUT) already depends on `Makefile`, I think you don't need it as dependency here too.

veluca93: $(CODEC_OUT) might not have its timestamp updated even if you do run the $(CODEC_OUT) rule, so this dependency is necessary.

RReverser: Isn't that a good thing? If $(CODEC_OUT) didn't have to get rebuilt (e.g. Makefile didn't change CXXFLAGS), then there is no need to rebuild .js either.

RReverser: Otherwise what you're defining here is essentially equivalent to `make -B` (force-rebuild everything) and doesn't need to be encoded in the Makefile.

veluca93: I do want to rebuild everything if I change the Makefile itself :)

RReverser: Right, so what I mean is, in that case you can do `make -B` or `make clean && make`.

I'm just not very comfortable with _any_ Makefile change triggering full rebuild, as I tend to do lots of them while adding new build targets (e.g. SIMD) or just refactoring rules, and in those cases it's nicer to make sure that existing targets remain as-is.

veluca93: A Makefile change not triggering a full rebuild would be a potentially very surprising thing though - for example, it could cause an update of the tag to have no effect whatsoever unless the user explicitly deletes everything.
Also, a "full rebuild" is not quite a full rebuild either, as the codec build system uses make itself and will just rebuild what needs to be updated.
However, if you have a strong opinion, I will just revert all the changes in the Makefile.

RReverser: > for example, it could cause an update of the tag to have no effect whatsoever unless the user explicitly deletes everything

That's a good point - we usually need `make clean` for that - but current changes don't fix that problem either IIUC? As in, they will reuse existing directly with JPEG-XL.

Let's revert all changes in Makefile for now, that will also simplify conflict resolution when merging SIMD-enabled makefile, but I'll definitely think of some pattern that should apply to all codecs as it's a common issue.

veluca93: Done.

RReverser (approved):

PR #851, opened 2020-11-19
SIMD+threads for WebP2; SIMD for WebP

RReverser:
This is first PR adding SIMD build to WebP and SIMD and SIMD+threads builds to WebP2 with feature detection and dynamic switching, rebased to work with Rollup builds.

For now pointing to my own forks in both codecs until corresponding patches allowing Emscripten SIMD are merged upstream.

Also, for now not including JPEG-XL changes as waiting for an ongoing update in #848 to be landed first to reduce conflicts with each other.


jakearchibald:
I've filed https://github.com/GoogleChromeLabs/squoosh/issues/852 for me to fix the header thing for live. I'm happy with the rest of the PR but most of it is outside my area of expertise 😄 

@surma can you review the rest?

codecs/webp/Makefile
```diff
@@ -1,19 +1,25 @@
-CODEC_URL := https://storage.googleapis.com/downloads.webmproject.org/releases/webp/libwebp-1.1.0.tar.gz
+CODEC_URL = https://github.com/RReverser/libwebp/archive/emscripten-simd.tar.gz
```
surma: Can you add a comment explaining when we can switch back to the upstream repo? Ideally with a link to the issue/PR or something that needs to land.

codecs/wp2/Makefile
```diff
@@ -1,20 +1,30 @@
-CODEC_URL = https://chromium.googlesource.com/codecs/libwebp2/+archive/c90b5b476004c9a98731ae1c175cebab5de50fbf.tar.gz
+CODEC_URL = https://github.com/RReverser/libwebp2/archive/emscripten-simd.tar.gz
```
surma: Same here as much as possible.

codecs/wp2/enc/wp2_enc.cpp
```diff
@@ -18,6 +19,7 @@ val encode(std::string image_in, int image_width, int image_height, WP2::Encoder
   }
 
   WP2::MemoryWriter memory_writer;
+  config.thread_level = emscripten_num_logical_cores() - 1;
```
surma, line 47: Why -1?

src/copy/_headers
```diff
@@ -8,3 +8,7 @@
 /*
   Cross-Origin-Embedder-Policy: require-corp
   Cross-Origin-Opener-Policy: same-origin
+# Origin trial for WebAssembly SIMD.
+# ATTENTION: This one is configured for dev--squoosh.netlify.app preview.
+# For production squoosh.app, replace with AgoKiDqjr0GVPtrwV/vuVlrrSvbDa5Yb99s+q66ly816DrrAQ8Cdas33NgDtmhxM4BtDP9PEdyuxHPyTQHD5ZAcAAABUeyJvcmlnaW4iOiJodHRwczovL3NxdW9vc2guYXBwOjQ0MyIsImZlYXR1cmUiOiJXZWJBc3NlbWJseVNpbWQiLCJleHBpcnkiOjE2MDg2NzI5OTR9.
```
surma, line 13: Might be time to finally switch this file to an .ejs file and use the right key depending on which branch is being built. But we can move this to a separate issue.

src/features/encoders/wp2/worker/wp2Encode.ts
```diff
@@ -10,24 +10,56 @@
...
+import wasmUrlWithMT from 'url:codecs/wp2/enc/wp2_enc_mt.wasm';
+import workerUrl from 'omt:codecs/wp2/enc/wp2_enc_mt.worker.js';
+
+import wasmUrlWithMTAndSIMD from 'url:codecs/wp2/enc/wp2_enc_mt_simd.wasm';
+import workerUrlWithSIMD from 'omt:codecs/wp2/enc/wp2_enc_mt_simd.worker.js';
+
 let emscriptenModule: Promise<WP2Module>;
 
+async function init() {
+  if (await threads()) {
```
surma, line 30:
```suggestion
  // At the time of writing, all browsers that support Wasm SIMD also support Wasm threads.
  // We might have to revisit this logic in the future.
  if (await threads()) {
```

surma (changes requested):
Some nits re documentation. Logic LGTM :)

codecs/wp2/enc/wp2_enc.cpp
RReverser: Because in WebP2 thread_level is number of additional threads (where `0` is threads being disabled). Yes, it's unusual and I had to check :)

src/copy/_headers
RReverser: Yeah Jake submitted one separately.

src/features/encoders/wp2/worker/wp2Encode.ts
RReverser: Will we add this to every codec using threads+SIMD or better to document this elsewhere in a centralised way if we want to?

codecs/wp2/enc/wp2_enc.cpp
surma: Add comment pls :)

src/features/encoders/wp2/worker/wp2Encode.ts
surma: For now, I’d just keep adding this comment until we have a better solution. Open to suggestions.

RReverser: I personally don't think it needs explanation, it's just "a combo we currently support" :)

RReverser: Initially I thought it would require explanations if we used just `if (await simd()) { ...assume that threads are also available... }` like initially discussed, but we're not relying on this kind of magic here, so the conditions on their own seem fairly straightforward to follow.

codecs/wp2/enc/wp2_enc.cpp
RReverser: Done.

surma (approved):

PR #857, opened 2020-11-20
Fix Rust / OxiPNG multithreaded builds

RReverser:


codecs/rust.Dockerfile
```diff
@@ -7,7 +7,7 @@ RUN wget -qO- https://github.com/rustwasm/wasm-pack/releases/download/v0.9.1/was
 FROM $RUST_IMG AS rust
 ARG RUST_IMG
 RUN rustup target add wasm32-unknown-unknown
-RUN if [[ $RUST_IMG = rustlang/rust:* ]] ; then rustup component add rust-src ; fi
+RUN case $RUST_IMG in rustlang/rust@*) rustup component add rust-src; esac
```
surma, line 10: wtf?

codecs/oxipng/build.sh
```diff
@@ -3,6 +3,8 @@
 set -e
 
 rm -rf pkg,{-parallel}
-wasm-pack build --target web
+wasm-pack build -t web
 RUSTFLAGS='-C target-feature=+atomics,+bulk-memory' wasm-pack build -t web -d pkg-parallel -- -Z build-std=panic_abort,std --features=parallel
+# Workaround https://github.com/rustwasm/wasm-bindgen/issues/2133:
+sed -i "s|maybe_memory:|maybe_memory?:|" pkg-parallel/squoosh_oxipng.d.ts
```
surma, line 9: Classic.

surma (approved):

codecs/rust.Dockerfile
RReverser: If you really want to know, turns out that syntax is bash-ism and not supported by regular `sh` which Docker uses. This is the recommended alternative.

PR #859, opened 2020-11-20
Autoformat staged C++ and Rust

RReverser:
Adding autoformat for codec binding sources, and also fixed formatting of existing ones in the same commit.

It could be possible to set up installation of clang-format and rustfmt too, but it's a bit more involved, and, I think, given that those commands will only be executed if the author edits C++ or Rust, I think it's fair to assume they already have the necessary toolchains installed. If not, we can revisit this later.


surma (approved):
Nice!

PR #861, opened 2020-11-25
New homepage

jakearchibald:


jakearchibald:
@argyleink https://deploy-preview-861--squoosh.netlify.app/ - let me know if the blob is moving too much. Can easily play with the numbers.

Also, the initial position is always the same (so it works with our prerender). If you're not happy with the initial position, go to the page, and press `b` when things are in a position you like, then give me the console output.

jakearchibald:
I haven't done a mobile layout yet - don't worry about that

src/shared/prerendered-app/colors.css
```diff
@@ -0,0 +1,19 @@
...
+  --pink: #ff3385;
+  --hot-pink: #ff0066;
+  --white: #fff;
+  --dim-blue: #0a7bcc;
+  --deep-blue: #09f;
+  --light-blue: #76c8ff;
+  --light-gray: #eaeaea;
+  --dark-text: #343a3e;
+
+  /* Old stuff: */
```
jakearchibald, line 11: I intend to remove/refactor these as I apply the redesign to the rest of the app

src/shared/prerendered-app/util.css
```diff
@@ -0,0 +1,24 @@
+:global {
```
jakearchibald, line 1: Making these global simplifies the build, and reduces duplication in the CSS

src/client/initial-app/custom-els/missing-types.d.ts
```diff
@@ -1,5 +1,5 @@
-/// <reference path="../../../shared/initial-app/custom-els/snack-bar/missing-types.d.ts" />
-/// <reference path="../../../shared/initial-app/custom-els/loading-spinner/missing-types.d.ts" />
+/// <reference path="../../../shared/custom-els/snack-bar/missing-types.d.ts" />
```
developit, line 1: Any chance this is worth doing via ambient declarations? We'd just need to add a package.json to each directory with `"types": "./missing-types.d.ts"` I think.

src/shared/prerendered-app/Intro/blob-anim/index.ts
```diff
@@ -0,0 +1,401 @@
...
+  document.addEventListener('visibilitychange', visibilityListener);
+
+  function destruct() {
+    removeEventListener('focus', focusListener);
+    removeEventListener('blur', blurListener);
+    document.removeEventListener('visibilitychange', visibilityListener);
+  }
+
+  function drawFrame(delta: number) {
+    const canvasBounds = canvas.getBoundingClientRect();
```
developit, line 348: A thought: there's already a ResizeObserver on `canvas`, and it could store the bounding rect rather than it being computed on every frame. Could also have it observe the `loadImgEl` to avoid both `getBoundingClientRect` calls. It also seems like the computed styles here are only used in a way that is static, so they could be hoisted out of `drawFrame()`.

src/shared/prerendered-app/util.css
developit: Are all `.css` files presumed to be ICSS modules by default? I've been seeing most folks move to only treating `.module.css` files as "css modules", which feels nice.

argyleink:
There's so many awesome things in your implementation! you've done a rad job with the morphing, pacing, etc. It's very very nice. 

**big blob notes:**
- none, it's rad.

**blob particle notes/idea/requests:**
- cut number of smaller blobs by 50%
- increase the dynamic range of blobs to allow more smaller ones
- gravitational pull needs some easing or strength? help it looks as though it pulls items faster when they're closer
- bigger blobs could move slower than smaller blobs?

**drop zone notes/idea/requests:**
- can you make the border dashed pink?
- (we can ship it with a matching border color i feel, but the next few tasks would be coooooo)
- can we forward the dragEnter event to make the main big blobs jiggle or show feedback of some kind?
- can clicking to upload an image jiggle or show feedback on the big blobs too?

**try one of these:**
- needs focus/hover styles (want a comp?)

**install pwa:**
- update the manifest to a new `theme_color`: either the new pink or blue (you pick!)

anything i missed that you hoped i'd inspect?

src/shared/prerendered-app/Intro/blob-anim/index.ts
jakearchibald: I figured calling `getBoundingClientRect` twice was cheap if there wasn't a layout in between, but I agree that I should just use the resize observer to read the bounds.

jakearchibald:
@developit 

> Are all `.css` files presumed to be ICSS modules by default? I've been seeing most folks move to only treating `.module.css` files as "css modules", which feels nice.

Can you file an issue for that? It feels low priority for the deadline, but I don't want to forget about it.

src/client/initial-app/custom-els/missing-types.d.ts
jakearchibald: Is that better?

src/shared/prerendered-app/Intro/blob-anim/index.ts
jakearchibald: I've removed this call to `getBoundingClientRect`, but I've left the `loadImgEl`. I'm pretty sure that `getBoundingClientRect` is cheap if the layout is valid, and devtools isn't showing any forced calculations. In fact, `getBoundingClientRect` might even be cheaper than an extra observer? I don't have any evidence though.

jakearchibald:
@argyleink 

> **blob particle notes/idea/requests:**
> 
> * cut number of smaller blobs by 50%
> * increase the dynamic range of blobs to allow more smaller ones
> * gravitational pull needs some easing or strength? help it looks as though it pulls items faster when they're closer
> * bigger blobs could move slower than smaller blobs?

Done! Happy to do further tweaks (or play with the numbers yourself)

> **drop zone notes/idea/requests:**
> 
> * can you make the border dashed pink?
> * (we can ship it with a matching border color i feel, but the next few tasks would be coooooo)

Done!

> * can we forward the dragEnter event to make the main big blobs jiggle or show feedback of some kind?
> * can clicking to upload an image jiggle or show feedback on the big blobs too?

Could you file an issue for these? I don't think I'll be able to do it before CDS though.

> **try one of these:**
> 
> * needs focus/hover styles (want a comp?)

Don't need a comp, but just let me know what needs to change & to what.
 
> **install pwa:**
> 
> * update the manifest to a new `theme_color`: either the new pink or blue (you pick!)

Done! Pink it is.

argyleink (approved):
self assigned to issues:
- https://github.com/GoogleChromeLabs/squoosh/issues/870
- https://github.com/GoogleChromeLabs/squoosh/issues/871

also opened a PR against this PR with eye-balled blob nudges 👀 
- https://github.com/GoogleChromeLabs/squoosh/pull/872

PR #866, opened 2020-12-02
Site isolation headers & redirect fix

### LINKED ISSUE #852: Use correct origin trial on live build
We need to use a different header on live than dev. Make this happen.

jakearchibald:
Fixes #852


src/static-build/utils.tsx
```diff
@@ -56,3 +56,16 @@ export function escapeStyleScriptContent(str: string): string {
     .replace(/<style/g, '<\\style')
     .replace(/<\/style/g, '<\\/style');
 }
+
+/**
+ * Removes empty lines from start and end, and removes the initial indent from
```
RReverser: I probably would have used `dedent` from npm to avoid dealing with getting regexes right, but maybe it's not worth it for something still fairly small :)

jakearchibald: Done! I wasn't aware of dedent.

RReverser (approved):
LGTM. I wish we could keep config options separate from code, but I don't see a good way to do that and it probably doesn't matter as much anyway.

jakearchibald:
I was going to create a build step that creates these from ejs templates, but it seemed to make more sense to use the templating system we had, rather than add another one.

PR #875, opened 2020-12-05
Add CLI

surma:
This PR adds the CLI to the main repository.

One heads up: The CLI currently doesn’t support encoding or decoding JPEG. I needed to add the MozJPEG _decoder_, as we can rely on browsers for decoding JPEG for the web. That needs a bit of re-jigging of the folder structure so I want to do that in a separate follow-up PR.


cli/src/codecs.js
```diff
@@ -0,0 +1,363 @@
+import { promises as fsp } from 'fs';
+import { instantiateEmscriptenWasm, pathify } from './emscripten-utils.js';
+
+// MozJPEG
+// import mozEnc from '../../codecs/mozjpeg/enc/mozjpeg_enc.js';
```
surma, line 5: This will get fixed in a follow-up PR

cli/src/codecs.js
```diff
@@ -0,0 +1,363 @@
...
+      };
+    },
+    defaultOptions: {
+      method: 'lanczos3',
+      fitMethod: 'stretch',
+      premultiply: true,
+      linearRGB: true,
+    },
+  },
+  // // TODO: Need to handle SVGs and HQX
```
surma, line 145: This should get turned into 2 issues.

cli/README.md
```diff
@@ -0,0 +1,58 @@
+# Squoosh CLI
+
+Squoosh CLI is an _experimental_ way to run all the codecs you know from the [Squoosh] web app on your command line using WebAssembly. The Squoosh CLI uses a worker pool to parallelize processing images. This way you can apply the same codec to many images at once.
```
RReverser, line 3: I don't know how Markdown references work, does `[Squoosh]` work despite reference having `[squoosh]` casing?

cli/README.md
```diff
@@ -0,0 +1,58 @@
...
+
+Squoosh CLI is currently not the fastest image compression tool in town and doesn’t aim to be on the thrown. It is, however, fast enough to compress many images sufficiently quick at once.
```
RReverser: What does "on the thrown" mean? Is it a typo?

cli/lib/asset-plugin.js
```diff
@@ -0,0 +1,75 @@
...
+ */
+import { promises as fs } from "fs";
+import { basename } from "path";
+
+const defaultOpts = {
+  prefix: "asset-url"
+};
+
+export default function assetPlugin(opts) {
+  opts = Object.assign({}, defaultOpts, opts);
```
RReverser:
```suggestion
  opts = { ...defaultOpts, ...opts };
```

cli/lib/json-plugin.js
```diff
@@ -0,0 +1,38 @@
...
+      const realId = id.slice(prefix.length);
+      const source = await fsp.readFile(realId, "utf8");
+
+      let code = "";
+      for (const [key, value] of Object.entries(JSON.parse(source))) {
+        if (reservedKeys.includes(key)) {
+          continue;
+        }
+        code += `
+				export const ${key} = ${JSON.stringify(value)};
```
RReverser: Is this code intentionally indented with tabs instead of spaces like the non-template parts?

cli/rollup.config.js
```diff
@@ -0,0 +1,45 @@
...
+/** @type {import('rollup').RollupOptions} */
+export default {
+  input: 'src/index.js',
+  output: {
+    dir: 'build',
+    format: 'cjs',
+    assetFileNames: '[name]-[hash][extname]',
+    // This is needed so the resulting `index.js` can be
+    // executed by `npx`.
+    banner: '#!/usr/bin/env node',
```
RReverser, line 18: Hmm, I thought `bin` field in `package.json` is enough? Otherwise, how would `npx` execute it on Windows where shebang is ignored?

cli/src/auto-optimizer.js
```diff
@@ -0,0 +1,70 @@
...
+  );
+
+  let bitmapOut;
+  let binaryOut;
+  // Increasing quality means _decrease_ in Butteraugli distance.
+  // `binarySearch` assumes that increasing `parameter` will
+  // increase the metric value. So multipliy Butteraugli values by -1.
+  const { parameter } = await binarySearch(
+    -1 * butteraugliDistanceGoal,
+    async quality => {
```
RReverser, line 56: Is it possible this is never executed and `*Out` vars remain `undefined`? Should we handle that case or is it fine?

cli/src/codecs.js
```diff
@@ -0,0 +1,363 @@
...
+    defaultOptions: {
+      numColors: 255,
+      dither: 1.0,
+    },
+  },
+  rotate: {
+    name: 'Rotate',
+    description: 'Rotate image',
+    instantiate: async () => {
+      return async (buffer, width, height, { numRotations }) => {
```
RReverser, line 167: IIRC, we have similar code in GUI. Can't we extract it into `codecs/rotate/rotate.js` and reuse between both?

cli/src/codecs.js
```diff
@@ -0,0 +1,363 @@
...
+  //     separate_chroma_quality: false,
+  //     chroma_quality: 75
+  //   },
+  //   autoOptimize: {
+  //     option: "quality",
+  //     min: 0,
+  //     max: 100
+  //   }
+  // },
+  webp: {
```
RReverser, line 228: All of this (extensions, default options) also feels like something we had largely in `encode_meta` modules - didn't we? Can we reuse/extend them?

cli/src/emscripten-utils.js
```diff
@@ -0,0 +1,13 @@
+export function pathify(path) {
+  if (path.startsWith("file://")) {
+    path = path.slice("file://".length);
```
RReverser: There are various cross-platform edge-cases with `file:` URIs being different. Better to use `require('url').fileURLToPath` here.

But, why do we ever get URIs in CLI in the first place?

cli/src/image_data.js
```diff
@@ -0,0 +1,12 @@
+export default class ImageData {
+  constructor(data, width, height) {
+    // Need to manually copy the memory as wasm-bindgen does not by default
+    // and by the time we get control in JS land, the memory has already
+    // been corrupted.
+    // FIXME: This is bad because it’s overhead that we should only need
```
RReverser: This is only necessary for `rotate.rs` because we do manual management there; any other Rust codecs will be fine, just like C++. Seems easy enough to move this copying only to `rotate`'s JS wrapper.

cli/src/index.js
```diff
@@ -0,0 +1,334 @@
...
+        return workerPool.dispatchJob({
+          file,
+          operation: "preprocess",
+          preprocessorName,
+          options: preprocessorOptions
+        });
+      })
+    );
+
+    for (const { file, bitmap, size } of decodedFiles) {
```
RReverser: What's happening here? Can you add a comment in the body if this loop is necessary?

cli/src/worker_pool.js
```diff
@@ -0,0 +1,103 @@
+import { Worker, parentPort } from "worker_threads";
+import { TransformStream } from "web-streams-polyfill";
+
+function uuid() {
+  return Array.from({ length: 16 }, () =>
+    Math.floor(Math.random() * 256).toString(16)
+  ).join("");
+}
```
RReverser:
```suggestion
import { randomFillSync } from "crypto";

function uuid() {
  return randomFillSync(Buffer.alloc(16)).toString('hex');
}
```

RReverser: Feel free to ignore, just seems a little more readable.

cli/src/worker_pool.js
```diff
@@ -0,0 +1,103 @@
...
+
+function jobPromise(worker, msg) {
+  return new Promise(resolve => {
+    const id = uuid();
+    worker.postMessage({ msg, id });
+    worker.on("message", function f({ result, id: rid }) {
```
RReverser: Should worker errors be handled as well? Or are they handled elsewhere?

cli/README.md
surma: Yeah that works :)

cli/README.md
surma:
```suggestion
Squoosh CLI is currently not the fastest image compression tool in town and doesn’t aim to be on the throne. It is, however, fast enough to compress many images sufficiently quick at once.
```

cli/lib/json-plugin.js
surma: I blame Jason :P

codecs/avif/Makefile
```diff
@@ -42,6 +59,7 @@ $(OUT_ENC_MT_JS): $(OUT_ENC_CPP) $(CODEC_DIR)/CMakeLists.txt $(LIBAOM_DIR)/CMake
 			-DCONFIG_AV1_DECODER=0 \
 			-DCONFIG_AV1_HIGHBITDEPTH=0 \
 		" \
+		ENVIRONMENT="worker" \
```
RReverser: Hmm, can you try to make this a variable dependency? I think that would still work and allow to avoid duplication between targets here and there.

Something like:

```make
$(OUT_ENC_JS) $(OUT_ENC_MT_JS): ENVIRONMENT=worker
$(OUT_NODE_ENC_JS): ENVIRONMENT=node

$(OUT_ENC_JS) $(OUT_ENC_MT_JS) $(OUT_NODE_ENC_JS): $(OUT_ENC_CPP) $(CODEC_DIR)/CMakeLists.txt $(LIBAOM_DIR)/CMakeLists.txt
    ...
    ENVIRONMENT=$(ENVIRONMENT) \
    ...
```

codecs/imagequant/Makefile
```diff
@@ -38,5 +55,5 @@ $(CODEC_DIR):
 	curl -sL $(CODEC_URL) | tar xz --strip 1 -C $@
 
 clean:
-	$(RM) $(OUT_JS) $(OUT_WASM)
+	$(RM) $(OUT_JS) *.wasm
```
RReverser: Same as above, once ENVIRONMENT is made a variable dependency, I think you can just put them into the same var like `OUT_JS := imagequant.js imagequant_node.js` and then this part can be reverted.

codecs/jxl/enc/jxl_enc.cpp
```diff
@@ -26,7 +26,7 @@ val encode(std::string image, int width, int height, JXLOptions options) {
   jxl::CodecInOut io;
   jxl::PaddedBytes bytes;
   jxl::ImageBundle* main = &io.Main();
-  jxl::ThreadPoolInternal *pool_ptr = nullptr;
+  jxl::ThreadPoolInternal* pool_ptr = nullptr;
```
RReverser, line 29: Huh, surprised my commit handler didn't catch this when I pushed stuff. Maybe I didn't do `npm install` lately though.

cli/rollup.config.js
surma: I didn’t work without this. Maybe you can give `npx @squoosh/cli` a quick test on your Windows machine?

cli/src/auto-optimizer.js
surma: It should at least be executed once.

cli/src/codecs.js
surma: Yup, we can and should. Just couldn’t do it because the other branch was in-flight. Will open an issue.

cli/src/codecs.js
surma: Yeah, this is one of these duplications that is going to be much easier to remove once we land this. Also the main project is TypeScript while the CLI currently is not. That should probably get changed as well at some point.

codecs/wp2/Makefile
```diff
@@ -4,8 +4,9 @@ CODEC_BUILD_ROOT := $(CODEC_DIR)/build
 CODEC_BASELINE_BUILD_DIR := $(CODEC_BUILD_ROOT)/baseline
 CODEC_MT_BUILD_DIR := $(CODEC_BUILD_ROOT)/mt
 CODEC_MT_SIMD_BUILD_DIR := $(CODEC_BUILD_ROOT)/mt-simd
+ENVIRONMENT = worker
```
RReverser, line 7: Nice!

cli/src/emscripten-utils.js
surma: > But, why do we ever get URIs in CLI in the first place?

That’s rollup’s doing :D

cli/src/image_data.js
surma: Not sure how without it being overwritten by every rebuild of rotate. But I’ll open an issue to fix this.

cli/rollup.config.js
RReverser: Seems to work.

cli/src/worker_pool.js
surma: The error handling of the entire CLI is not good. But I could at least add a log message I guess.

codecs/avif/Makefile
RReverser: (I see you already did this in other makefiles, should be easy to do the same here)

cli/src/image_data.js
RReverser: You have JS binding above, where you grow memory and so on - can't you just put `new Uint8ClampedArray(...)` in there?

RReverser: JS is not autogenerated for rotate, we have custom one.

cli/src/index.js
surma: lol wtf. That loop is not needed :D

codecs/png/package.json
```diff
@@ -0,0 +1,6 @@
+{
+  "name": "oxipng",
+  "scripts": {
+    "build": "../build-rust.sh rm -rf pkg && wasm-pack build --debug --target web -- --verbose --locked && rm pkg/.gitignore"
```
RReverser: Um is this meant to still be built in a debug mode?

surma: Ah good catch! Definitely not.

cli/src/image_data.js
RReverser: Resize JS code already does `new Uint8ClampedArray`, so can just remove here.

cli/src/codecs.js
```diff
@@ -0,0 +1,363 @@
...
+          (size * 2 - memory.buffer.byteLength + 8) / (64 * 1024),
+        );
+        if (additionalPagesNeeded > 0) {
+          memory.grow(additionalPagesNeeded);
+        }
+        const view = new Uint8ClampedArray(memory.buffer);
+        view.set(buffer, 8);
+        instance.exports.rotate(width, height, degrees);
+        return new ImageData(
+          new Uint8ClampedArray(view.slice(size + 8, size * 2 + 8)),
```
RReverser:
```suggestion
          view.slice(size + 8, size * 2 + 8),
```

`.slice` result is already a copy, no need to wrap again.

cli/src/image_data.js
```diff
@@ -0,0 +1,12 @@
...
+    // to pay for Rust, not for C++.
+    this.data = new Uint8ClampedArray(data);
```
RReverser:
```suggestion
    this.data = data;
```

cli/src/codecs.js
```diff
@@ -0,0 +1,363 @@
...
+    },
+    enc: async () => {
+      await pngEncDecPromise;
+      await oxipngPromise;
+      return {
+        encode: (buffer, width, height, opts) => {
+          const simplePng = new Uint8Array(
+            pngEncDec.encode(new Uint8Array(buffer), width, height),
+          );
+          return new Uint8Array(oxipng.optimise(simplePng, opts.level));
```
RReverser:
```suggestion
          return oxipng.optimise(simplePng, opts.level);
```

Also unnecessary wrapping, wasm-bindgen returns JS-owned typed arrays.

codecs/png/src/lib.rs
```diff
@@ -0,0 +1,47 @@
...
+        for i in (0..num_pixels).rev() {
+            buf[i * 4 + 0] = buf[i * 3 + 0];
+            buf[i * 4 + 1] = buf[i * 3 + 1];
+            buf[i * 4 + 2] = buf[i * 3 + 2];
+            buf[i * 4 + 3] = 255;
+        }
+    }
+
+    ImageData::new_with_u8_clamped_array_and_sh(
+        wasm_bindgen::Clamped(&mut buf),
```
RReverser: Now this one I think is a problem, because it uses a reference to Rust memory and then stores it in the ImageData.

codecs/jxl/Makefile
```diff
@@ -4,24 +4,27 @@ CODEC_DIR = node_modules/jxl
...
 
-OUT_JS = enc/jxl_enc.js enc/jxl_enc_mt.js enc/jxl_enc_mt_simd.js dec/jxl_dec.js
+OUT_JS = enc/jxl_enc.js enc/jxl_enc_mt.js enc/jxl_enc_mt_simd.js dec/jxl_dec.js enc/jxl_node_enc.js dec/jxl_node_dec.js
 OUT_WASM = $(OUT_JS:.js=.wasm)
 OUT_WORKER = $(OUT_JS:.js=.worker.js)
 
 .PHONY: all clean
 
-all: $(OUT_JS)
+all: $(OUT_JS)
```
RReverser:
```suggestion
all: $(OUT_JS)
```

cli/src/codecs.js
```diff
@@ -0,0 +1,363 @@
...
+      await pngEncDecPromise;
+      return { decode: pngEncDec.decode };
+    },
+    enc: async () => {
+      await pngEncDecPromise;
+      await oxipngPromise;
+      return {
+        encode: (buffer, width, height, opts) => {
+          const simplePng = new Uint8Array(
+            pngEncDec.encode(new Uint8Array(buffer), width, height),
```
RReverser:
```suggestion
            pngEncDec.encode(buffer, width, height),
```

Also, wasm-bindgen returns a `Vec<u8>` so this stuff is already owned by JS - no need to reclone.

RReverser: Er, wait, wrong suggestion.

cli/src/codecs.js
```diff
@@ -0,0 +1,363 @@
...
+          );
```
RReverser:
```suggestion
          const simplePng = pngEncDec.encode(new Uint8Array(buffer), width, height);
```

Also, wasm-bindgen returns a Vec<u8> so this stuff is already owned by JS - no need to reclone.

cli/src/image_data.js
RReverser: I think comments can be removed now as well.

codecs/png/src/lib.rs
RReverser: Their API binding is pretty bad here TBH. It basically doesn't allow a JS-owned Uint8ClampedArray. You'll have to add a custom binding (should be easy enough though, let me know if you need help).

RReverser: Fixed in [`5758e5d` (#875)](https://github.com/GoogleChromeLabs/squoosh/pull/875/commits/5758e5db9ae6c0cc30e7341d49fcdd731f09f1d0).

developit:
@googlebot I consent.

RReverser (approved):

cli/README.md
```diff
@@ -0,0 +1,58 @@
+# Squoosh CLI
+
+Squoosh CLI is an _experimental_ way to run all the codecs you know from the [Squoosh] web app on your command line using WebAssembly. The Squoosh CLI uses a worker pool to parallelize processing images. This way you can apply the same codec to many images at once.
+
+Squoosh CLI is currently not the fastest image compression tool in town and doesn’t aim to be on the throne. It is, however, fast enough to compress many images sufficiently quick at once.
```
developit: It's nitpicky but "be on the throne" is an american colloquialism for going to the washroom.

```suggestion
Squoosh CLI is currently not the fastest image compression tool in town and doesn’t aim to be. It is, however, fast enough to compress many images sufficiently quick at once.
```

cli/package.json
```diff
@@ -0,0 +1,32 @@
...
+  "public": true,
+  "bin": {
+    "squoosh-cli": "build/index.js",
+    "@squoosh/cli": "build/index.js"
+  },
+  "scripts": {
+    "build": "rollup -c"
+  },
+  "keywords": [],
+  "author": "Surma <surma@surma.dev>",
```
developit:
```suggestion
  "author": "Google Chrome Developers <chromium-dev@google.com>",
```

cli/tmp.txt
```diff
@@ -0,0 +1,12 @@
+
```
developit, line 1: accidentally included?

developit (approved):

cli/tmp.txt
RReverser: Looks so!

cli/src/image_data.js
```diff
@@ -0,0 +1,12 @@
+export default class ImageData {
+  constructor(data, width, height) {
+    // Need to manually copy the memory as wasm-bindgen does not by default
+    // and by the time we get control in JS land, the memory has already
+    // been corrupted.
+    // FIXME: This is bad because it’s overhead that we should only need
+    // to pay for Rust, not for C++.
```
surma:
```suggestion
```

PR #876, opened 2020-12-07
Allow non-0 quality in webp

### LINKED ISSUE #873: WebP quality broken in `dev` branch
Since WebP was update, the quality has gone really bad. I haven't dug into it, but maybe they changed how their quality setting works and we didn't update UI/bindings to cater for it?

jakearchibald:
Fixes #873.

I think we're doing the wrong thing by accepting a `WebPConfig` as an arg, unless we can somehow pass it through `WebPConfigInit` before emscripten applies the values. But anyway, this fixes it for now.


surma (approved):
JAKE SAVES THE DAY

RReverser:
Thanks for fixing!

> I think we're doing the wrong thing by accepting a `WebPConfig` as an arg, unless we can somehow pass it through `WebPConfigInit` before emscripten applies the values

Oh, `WebPConfig` doesn't have a default C++ constructor? Then yeah, we should wrap it... Good catch!

PR #878, opened 2020-12-07
Make the JXL Makefile work with both tags and SHAs.

veluca93:
This allows to update the codec without relying on upstream tags.

I didn't make the checkout rule depend on the Makefile, which means that version updates will only come into effect after removing `node_modules`, but it should be possible in principle.

### DIFF
--- a/codecs/jxl/Makefile
+++ b/codecs/jxl/Makefile
@@ -1,5 +1,5 @@
 CODEC_URL = https://gitlab.com/wg1/jpeg-xl.git
-CODEC_VERSION = v0.1.1
+CODEC_VERSION = 739e6cd1305fdec5acfa47ad414189b3d3ecb6a4
 CODEC_DIR = node_modules/jxl
 CODEC_BUILD_ROOT := $(CODEC_DIR)/build
 CODEC_MT_BUILD_DIR := $(CODEC_BUILD_ROOT)/mt
@@ -78,9 +78,12 @@ $(CODEC_MT_SIMD_BUILD_DIR)/Makefile: CXXFLAGS+=-msimd128
 	-B $(@D) \
 	$(<D)
 
-$(CODEC_DIR)/CMakeLists.txt:
-	mkdir -p $(@D)
-	git clone $(CODEC_URL) --recursive -j`nproc` --depth 1 --branch $(CODEC_VERSION) $(@D)
+$(CODEC_DIR)/CMakeLists.txt: 
+	$(RM) -r $(@D)
+	git init $(@D)
+	git -C $(@D) fetch $(CODEC_URL) $(CODEC_VERSION) --depth 1
+	git -C $(@D) checkout FETCH_HEAD
+	git -C $(@D) submodule update --init --depth 1 --recursive --jobs `nproc`
 
 clean:
 	$(RM) $(OUT_JS) $(OUT_WASM) $(OUT_WORKER)


codecs/jxl/Makefile
```diff
@@ -78,9 +78,14 @@ $(CODEC_MT_SIMD_BUILD_DIR)/Makefile: CXXFLAGS+=-msimd128
 	-B $(@D) \
 	$(<D)
 
-$(CODEC_DIR)/CMakeLists.txt:
+$(CODEC_DIR)/CMakeLists.txt: 
+	rm -rf $(@D)
 	mkdir -p $(@D)
-	git clone $(CODEC_URL) --recursive -j`nproc` --depth 1 --branch $(CODEC_VERSION) $(@D)
+	cd $(@D) && git init .
```
deymo: `git init $(@D)`

and then you can do `git -C $(@D) remote add ......` so you don't need to do `cd`

codecs/jxl/Makefile
```diff
@@ -78,9 +78,14 @@ $(CODEC_MT_SIMD_BUILD_DIR)/Makefile: CXXFLAGS+=-msimd128
...
+	cd $(@D) && git remote add origin $(CODEC_URL)
+	cd $(@D) && git fetch origin $(CODEC_VERSION) --depth 1
+	cd $(@D) && git checkout $(CODEC_VERSION)
```
deymo: I'd checkout `FETCH_HEAD` here since the "$(CODEC_VERSION)" may resolve to something different here. For example, if CODEC_VERSION is "master" this will resolve first to the local "master" branch if one exists rather than origin/master.

surma:
I actually think we used to have this and we changed it for some reason. @RReverser do you remember?

Either way, this LGTM. Gonna wait for Ingvar to make the final call, tho.

veluca93:
> I actually think we used to have this and we changed it for some reason. @RReverser do you remember?
> 
> Either way, this LGTM. Gonna wait for Ingvar to make the final call, tho.

I think the reason was that the old way didn't do shallow checkouts and/or parallel cloning of submodules (don't remember which, possibly both), but this implementation ought to do both.

RReverser:
> I actually think we used to have this and we changed it for some reason. @RReverser do you remember?

Yeah, I added it on my JPEG-XL SIMD branch but then reverted before merging the PR because a new version was tagged, and it didn't seem worth the Makefile complication anymore.

Do we think it's still / again worth it? If there's a usecase (e.g. tags are not catching up fast enough), I don't mind.

> I think the reason was that the old way didn't do shallow checkouts and/or parallel cloning of submodules (don't remember which, possibly both), but this implementation ought to do both.

The one we're talking about, did - it was pretty similar to your PR (I think the only difference is `-C $(@D)` vs `cd ... && git ... && git ...`).

codecs/jxl/Makefile
```diff
@@ -78,9 +78,13 @@ $(CODEC_MT_SIMD_BUILD_DIR)/Makefile: CXXFLAGS+=-msimd128
 	-B $(@D) \
 	$(<D)
 
-$(CODEC_DIR)/CMakeLists.txt:
-	mkdir -p $(@D)
-	git clone $(CODEC_URL) --recursive -j`nproc` --depth 1 --branch $(CODEC_VERSION) $(@D)
+$(CODEC_DIR)/CMakeLists.txt: 
+	rm -rf $(@D)
+	git init $(@D)
+	git -C $(@D) remote add origin $(CODEC_URL)
```
RReverser: This line is unnecessary, you can just `fetch $(CODEC_URL) $(CODEC_VERSION)` below.

veluca93: Nice!

veluca93:
> > I actually think we used to have this and we changed it for some reason. @RReverser do you remember?
> 
> Yeah, I added it on my JPEG-XL SIMD branch but then reverted before merging the PR because a new version was tagged, and it didn't seem worth the Makefile complication anymore.
> 
> Do we think it's still / again worth it? If there's a usecase (e.g. tags are not catching up fast enough), I don't mind.
> 
> > I think the reason was that the old way didn't do shallow checkouts and/or parallel cloning of submodules (don't remember which, possibly both), but this implementation ought to do both.
> 
> The one we're talking about, did - it was pretty similar to your PR (I think the only difference is `-C $(@D)` vs `cd ... && git ... && git ...`).

Yup, it's easier to get bugfixes by checking out a sha instead of waiting for a release on our side (which is actually slightly more involved than just tagging a commit)

codecs/jxl/Makefile
```diff
@@ -78,9 +78,12 @@ $(CODEC_MT_SIMD_BUILD_DIR)/Makefile: CXXFLAGS+=-msimd128
 	-B $(@D) \
 	$(<D)
 
-$(CODEC_DIR)/CMakeLists.txt:
-	mkdir -p $(@D)
-	git clone $(CODEC_URL) --recursive -j`nproc` --depth 1 --branch $(CODEC_VERSION) $(@D)
+$(CODEC_DIR)/CMakeLists.txt: 
+	rm -rf $(@D)
```
RReverser:
```suggestion
	$(RM) $(@D)
```

codecs/jxl/Makefile
```diff
@@ -78,9 +78,12 @@ $(CODEC_MT_SIMD_BUILD_DIR)/Makefile: CXXFLAGS+=-msimd128
...
+	git init $(@D)
+	git -C $(@D) fetch $(CODEC_URL) $(CODEC_VERSION) --depth 1
+	git -C $(@D) checkout FETCH_HEAD
+	git -C $(@D) submodule update --init --recursive --jobs `nproc`
```
RReverser: Found one more optimisation that was in my fork but not here:

```suggestion
	git -C $(@D) submodule update --init --depth 1 --recursive --jobs `nproc`
```

codecs/jxl/Makefile
veluca93: This still needs a -r, right? Done.

RReverser (approved):

PR #880, opened 2020-12-08
(Almost the) rest of the redesign

jakearchibald:
Aside from the viewport toolbar, and pop-out menus, which @developit is working on.


argyleink (approved):
stoked!

big-skateboard

PR #896, opened 2020-12-09
leverage fix from https://github.com/v8/v8.dev/pull/504

argyleink:
This makes it so that AVIF images still work when requested directly, see https://github.com/v8/v8.dev/pull/504

cc @mathiasbynens


mathiasbynens (approved):

surma:
I don’t think we serve any AVIF files, do we?

RReverser:
> I don’t think we serve any AVIF files, do we?

Yeah I don't think we do.

surma:
I actually also would Netlify expect to add this mime type by default. But imma close this for now :D

PR #899, opened 2020-12-10
Typo fix in README

tomayac:
#Hacktoberfest


surma:
#891 was quicker :P Thanks tho!

PR #906, opened 2020-12-10
Clarify CLI file size unit output

mathiasbynens:
This patch clarifies that the file size output uses 1024-based calculations (KiB, MiB) instead of 1000-based ones (kB, MB). It also adds a space between the number and unit per the Unicode CLDR conventions (which can also be observed via the `Intl.NumberFormat` API).

Related (but out of scope for this patch): [DevTools uses 1000-based calculations and units in its UI.](https://goo.gle/devtools-si) Not sure if this is something you’d want to align on.

### DIFF
--- a/cli/src/index.js
+++ b/cli/src/index.js
@@ -19,11 +19,11 @@ function clamp(v, min, max) {
   return v;
 }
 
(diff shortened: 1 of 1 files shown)


ziemek99:

PR #911, opened 2020-12-13
Fix oxipng build script permissions

eugenezastrogin:
Building oxipng fails due to missing x permission on build.sh script with something along the lines of:

`docker: Error response from daemon: OCI runtime create failed: container_linux.go:370: starting container process caused: exec: "./build.sh": permission denied: unknown.`

This fixes it.


eugenezastrogin:
@googlebot I signed it!

RReverser (approved):
Thanks!

Issue #871, opened 2020-12-03
[new homepage] samples images need focus/hover styles


argyleink:

harsh26shah03:
@argyleink I have made these changes please have a look

PR #914, opened 2020-12-16
Restore focus outline on buttons

KittyGiraudel:
Hello there! ✨

Most buttons on the home page use the class `unbutton` to reset the default button styles such as background, border and font styles. Unfortunately, they also remove the focus outline, and do not declare any focus styles. This is an accessible issue, because it completely breaks keyboard navigation. This PR intends to restore the default browser outline for focusable buttons.

### DIFF
--- a/src/shared/prerendered-app/Intro/style.css
+++ b/src/shared/prerendered-app/Intro/style.css
@@ -222,12 +222,13 @@
 .demo-icon-container {
   border-radius: var(--demo-size);
   position: relative;
-  overflow: hidden;
+  transition: box-shadow 250ms;
 }
 .demo-icon {
   width: var(--demo-size);
   height: var(--demo-size);
   display: block;
+  border-radius: 50%;
 }
 .demo-loader {
   composes: abs-fill from global;
@@ -245,5 +246,15 @@
 .drop-text {
   @media (max-width: 599px) {
     display: none;
+  }
+}
+
+.sample-button {
+  &:focus {
+    outline: 0;
+  }
+
+  &:focus-visible .demo-icon-container {
+    box-shadow: 0 0 0 4px var(--deep-blue), 0 0 0 8px var(--pink);
   }
 }
--- a/src/shared/prerendered-app/Intro/style.css
+++ b/src/shared/prerendered-app/Intro/style.css
@@ -251,7 +251,7 @@
 
 .sample-button {
   &:focus {
-    outline: 0;
+    outline: transparent solid 2px;
   }
 
   &:focus-visible .demo-icon-container {
--- a/src/shared/prerendered-app/util.css
+++ b/src/shared/prerendered-app/util.css
@@ -16,9 +16,5 @@
     font: inherit;
     padding: 0;
     margin: 0;
-
-    &:focus {
-      outline: none;
-    }
   }
 }
--- a/src/shared/prerendered-app/Intro/index.tsx
+++ b/src/shared/prerendered-app/Intro/index.tsx
@@ -312,7 +312,7 @@ export default class Intro extends Component<Props, State> {
               {demos.map((demo, i) => (
                 <li>
                   <button
-                    class="unbutton"
+                    class={'unbutton ' + style.sampleButton}
                     onClick={(event) => this.onDemoClick(i, event)}
                   >
                     <div>


KittyGiraudel:
@googlebot I signed it!

src/shared/prerendered-app/util.css
```diff
@@ -16,9 +16,5 @@
     font: inherit;
     padding: 0;
     margin: 0;
```
argyleink, line 18:
```suggestion
    margin: 0;
    outline-offset: 10px;
```

argyleink (changes requested):
related to #871 

there's more focus work to be done 😉  you up for it?

src/shared/prerendered-app/util.css
KittyGiraudel, line 18: Screenshot 2020-12-16 at 17 14 21
Decided to go this route for focus styles. :)

src/shared/prerendered-app/Intro/style.css
```diff
@@ -247,3 +248,13 @@
...
 }
+
+.sample-button {
+  &:focus {
+    outline: 0;
+  }
+
+  &:focus .demo-icon-container {
+    box-shadow: 0 0 0 4px var(--deep-blue), 0 0 0 8px var(--pink);
+  }
```
jakearchibald: Feels like we could use `focus-visible` for these?

src/shared/prerendered-app/util.css
```diff
@@ -16,9 +16,5 @@
...
-
-    &:focus {
-      outline: none;
-    }
```
jakearchibald, line 22: I think we'd need to review the impact of this across the rest of the app before landing this.

src/shared/prerendered-app/Intro/style.css
KittyGiraudel: I’m not sure how would that help. What do you have in mind? :)

src/shared/prerendered-app/util.css
KittyGiraudel: I understand where you’re coming from although I respectfully disagree. Right now, buttons are being stripped out of their focus outline entirely, which is a critical accessibility issue. I would argue the default browser outline is better than no outline at all. These focus states can then be incrementally updated to fit the design more.

KittyGiraudel: That being said, I’m happy removing that part of the diff if you would feel more comfortable.

src/shared/prerendered-app/Intro/style.css
jakearchibald: It would avoid showing the focus outline during mouse interaction.

src/shared/prerendered-app/util.css
jakearchibald: I didn't say it should be removed, I just want it to be reviewed.

Specifically, I want to avoid buttons that look "stuck" for mouse users.

src/shared/prerendered-app/Intro/style.css
KittyGiraudel: Fair enough. :) Applied!

src/shared/prerendered-app/util.css
KittyGiraudel: @argyleink Do you want me to add that `outline-offset: 10px` line to this PR?

src/shared/prerendered-app/util.css
KittyGiraudel: I see, that makes sense.

src/shared/prerendered-app/util.css
argyleink: not needed anymore with your enhanced box-shadow solution, will remove 👍🏻

src/shared/prerendered-app/util.css
aquaductape: > Specifically, I want to avoid buttons that look "stuck" for mouse users.

Use focus-visible polyfill, the focus outline only triggers when activated by keyboard use https://github.com/WICG/focus-visible

src/shared/prerendered-app/Intro/style.css
```diff
@@ -247,3 +248,13 @@
     display: none;
   }
 }
+
+.sample-button {
+  &:focus {
+    outline: 0;
```
darekkay: Was this tested with the Windows High Contrast Mode? Using a box shadow instead of an outline usually doesn't work in WHCM, so one needs to provide a [workaround](https://stackoverflow.com/questions/52589391/css-box-shadow-vs-outline/52616313#52616313):

```suggestion
    outline: transparent solid 2px;
```

KittyGiraudel: Great idea! Thank you for the suggestion. :)

KittyGiraudel:
The commit suggestion feature from GitHub caused my last commit to be co-authored with @darekkay since they rightfully suggested a WCHM improvement. I don’t know whether they intend to sign the SLA. If they don’t, I’ll rebase my commit to remove them as co-author.

darekkay:
I've signed the CLA, so it should be fine now.

@googlebot I fixed it.

PR #923, opened 2020-12-26
[CLI] allow passing entire directories as input

DetachHead:
useful on windows where `./*` doesn't expend the contents of a directory, and unlike `./*` this method allows for spaces in filenames


DetachHead:
@googlebot I signed it!

surma (approved):
Ah nice one. Thank you.

PR #926, opened 2020-12-27
[CLI] fix hanging when there are more files than cpu cores

### LINKED ISSUE #925: [CLI] hangs after processing one image for each thread when passing more files than cpu cores
**Describe the bug**
if you pass in more files than your computer has CPU cores, it will process one file in each thread, then hang.
image

**To Reproduce**

1. navigate to a folder with more images than your CPU's cores (in my case 8, but reproduced on two different PCs with 4 and 6 cores)
2. run the squoosh CLI on all images in that folder

**Expected behavior**
images successfully squooshed

**Version:**
 - OS w/ version: Ubuntu 20.04.1 (also reproduced on windows 10)
 - Browser w/ version: N/A
 - Node version: 15.4.0 & 15.5.0
 - npm version: 7.3.0

DetachHead:
closes #925 

as far as i could tell, it looks like the issue was caused by `closing` being set to `true` before all of the jobs were completed, causing `_readLoop` to terminate workers instead of picking up the next job.

### DIFF
--- a/cli/src/worker_pool.js
+++ b/cli/src/worker_pool.js
@@ -81,9 +81,9 @@ export default class WorkerPool {
   }
 
   async join() {
-    this.closing = true;
     this.jobQueue.writable.close();
     await this.done;
+    this.closing = true;
   }
 
   dispatchJob(msg) {


surma:
Thanks for taking the time to whip this up!

I think you are right about the diagnosis, but this is sadly not quite the correct fix. I think i have something working on a branch and will merge that soon. Closing this for now.

PR #929, opened 2021-01-07
Simplify image support detection for SW

RReverser:
Instead of using extra files + a plugin to inline images, report image support from the main thread before caching assets.


jakearchibald:
Won't this break if we add a new browser format like `supergif`?

1. Cached version of site loads.
2. Service worker update found.
3. Service worker looks up `get('supports-supergif')` which is `undefined`, so the decoder is cached.
4. Next time the page loads, the page sets `'supports-supergif'` to `true`, since the browser supports the format.

But too late, the decoder is cached.

RReverser:
Don't we ask user to reload the page when an update is found?

I did see caching invoked from https://github.com/GoogleChromeLabs/squoosh/blob/9062a755415eac5724ee0d08e2652562ef1ff966/src/sw/index.ts#L18-L32 but wasn't sure when it's used as opposed to the one from `mainAppLoaded` function - I guess SW updates is precisely the user-case?

jakearchibald:
> Don't we ask user to reload the page when an update is found?

That happens when an update is ready to go, which is after everything's been cached.

RReverser:
> That happens when an update is ready to go, which is after everything's been cached.

Ah, interesting. So, should I call format support detection from some another place? Or the only way would be if SW itself asks the main thread which formats are supported via postMessage?

jakearchibald:
> Or the only way would be if SW itself asks the main thread which formats are supported via postMessage?

There's no guarantee there's a document open for the service worker to ask.

RReverser:
TIL. So it's impossible to implement this approach in a fully correct way?

jakearchibald:
I don't think it's possible. The service worker needs to be able to conduct updates on its own.

RReverser:
Okay. Too bad, I thought this was a nice way to reduce deps.

PR #930, opened 2021-01-07
Update WP2

surma:
Update WP2 and make use of new flag to reduce Wasm binary size (shaves off between 50-60kB).


RReverser (approved):
Nice. What does WP2_REDUCED exclude?

skal65535:
WP2_REDUCED removes all the visual-debug code and calls.
This visual-debug is only useful to the 'vwp2' inspection tool...

PR #933, opened 2021-01-14
Add manifest fields for PWA install bottom sheet

beaufortfrancois:
This PR adds [standardized](https://developer.mozilla.org/en-US/docs/Web/Manifest#members) manifest fields for PWA install bottom sheet such as `description`, `lang`, `categories`, and `screenshots` so that the PWA install bottom sheet looks great. See screenshot below.

Note that it currently requires `chrome://flags/#mobile-pwa-install-use-bottom-sheet` flag in Chrome Canary on Android.

image


src/static-build/index.tsx
```diff
@@ -60,6 +63,27 @@ const toOutput: Output = {
         purpose: 'maskable',
       },
     ],
+    description:
+      'Compress and compare images with different codecs, right in your browser.',
+    lang: 'en',
+    categories: ['photo', 'productivity', 'utilities'],
```
jakearchibald, line 73: @beaufortfrancois are these from a particular list of categories?

beaufortfrancois: Indeed. it comes from https://github.com/w3c/manifest/wiki/Categories#standardized-categories

jakearchibald:
@googlebot I consent.

jakearchibald:
@beaufortfrancois I've switch the images to webp so they're around a 10th of the size. Is webp ok to use here?

beaufortfrancois:
WebP is still not supported on Android sadly. See https://source.chromium.org/chromium/chromium/src/+/master:components/webapps/browser/installable/installable_manager.cc;l=113;bpv=1;bpt=1?q=fetchScreenshot&ss=chromium and https://bugs.chromium.org/p/chromium/issues/detail?id=466958
It is supported on Desktop though but this bottom sheet is only for Android, hence why I've used png there.
We can try to reduce image dimension if you want.

jakearchibald:
@surma could you do a review of this? I've added quite a bit to it

src/static-build/index.tsx
kenchris: https://www.w3.org/TR/manifest-app-info/#categories-member

lib/url-plugin.js
```diff
@@ -12,9 +12,11 @@
  */
 import { promises as fs } from 'fs';
 import { basename } from 'path';
+import { imageSize } from 'image-size';
 
 const defaultOpts = {
   prefix: 'url',
+  imagePrefix: 'img-url',
```
surma, line 19: Cheeky suggestion: Wouldn’t it keep this plugin simpler if instead of a second prefix it would take a list of file extensions that we consider images?

surma (approved):

lib/url-plugin.js
jakearchibald: Hmm, that would make the typing a lot harder. RIght now the module definition is based on prefixes. What you're proposing would depend on both a prefix and a (set of) suffixes, and I don't think TS supports that.

I'll keep it explicit for now, but if the typing can be done, happy for it to change.

PR #935, opened 2021-01-14
Rebuild codecs

RReverser:
It appears that C++ codecs haven't been rebuilt after some config changes (AVIF size difference is particularly suspicious).


jakearchibald:
That is a big jump for AVIF. Any idea what caused that?

surma:
Hm no idea why AVIF would increase that much 🤔

RReverser:
Yeah me neither; I only did `rm -rf node_modules && npm run build` for each codec. I don't see anything that would explain it in the history either.

I initially thought https://github.com/GoogleChromeLabs/squoosh/pull/838 got lost during rebase, but that one was still a lot smaller.

RReverser:
Looks like https://github.com/GoogleChromeLabs/squoosh/pull/839 had the same size increase to 2MB. So there are two parts:

1) Why did size regress so much in that version that was supposed to be pretty much no-op other than AVIF upgrade?
2) Why did size bounce back later - did someone build codecs with stale version of AVIF?

RReverser:
I think most of the size increase is just an upgrade of LibAVIF + usage of -O3. -O3 was added in the first place for the benefit of AVIF, and, as far as I remember, made a significant performance increase, but it does so by inlining everything everywhere, so size increases are also inevitable.

RReverser:
I'll play around with it a bit more.

RReverser:
Tried with -O2, there is pretty much no difference in terms of size (in fact it went a bit up). I'll try -Os and see what benchmarks show. But I'm starting to suspect that the AVIF upgrade just had a lot more code and it's something we'll have to live with.

RReverser:
Okay so yeah our stale Wasm version was either pre-upgrade or built with -Os. Rebuilding from -O3 to -Os gives:

Wasm: 2093358 bytes -> 1432766 bytes (-32% or 1.46x smaller)
Speed: 0.31 ops/sec -> 0.18 ops/sec (-42% or 1.72x slower)

Like in the initial -O3 PR, I'd still say it's worth it given how slow AVIF is otherwise / in general.

jakearchibald (approved):
I say ship it. The performance increase is worth it.

PR #938, opened 2021-01-18
Disable filesystem in C++ codecs

RReverser:
Explicitly disable filesystem for Emscripten codecs. There are no changes in most cases, as compiler managed to detect that code doesn't use filesystem anyway, but in AVIF case, this shaves 15 KB off generated JS files (which is 30-40% of the corresponding JS file depending on the config).


surma (approved):
Nice catch!

RReverser:
Note: this also speeds up initialization, because Emscripten doesn't have to precreate filesystem tree structure.

PR #939, opened 2021-01-18
Disable Worker termination on dev

RReverser:
This helps inspect the source code without it disappearing right from under your nose in 10 seconds.


jakearchibald:
Could this be a different flag? I want to avoid as many live-only behaviours as possible.

RReverser:
I guess I can add a separate flag, yeah. I thought this was a minor enough difference.

jakearchibald:
I'm worried that, if the termination code gets broken, we might end up killing the worker half way through its task, but we wouldn't see the problem until live.

PR #940, opened 2021-01-19
Make MozJPEG builds reproducible

RReverser:
I was wondering why MozJPEG produces different Wasm binaries even when nothing is changed.

After looking at the binary diffs, I think I have figured & fixed the reason.


codecs/mozjpeg/Makefile
```diff
@@ -46,7 +46,10 @@ $(CODEC_DIR)/Makefile: $(CODEC_DIR)/configure
 		--without-turbojpeg \
 		--without-simd \
 		--without-arith-enc \
-		--without-arith-dec
+		--without-arith-dec \
+		--with-build-date=squoosh
```
surma, line 50: lol

surma (approved):

codecs/mozjpeg/Makefile
RReverser: tHiS iS fInE

PR #941, opened 2021-01-20
Use latest wasm-opt instead of the one in wasm-pack

RReverser:
wasm-pack auto-downloads and runs a fixed version of wasm-opt, but wasm-pack itself hasn't been maintained for a while, and so wasm-opt that comes with it is also severely outdated, leading to all sorts of hidden issues (compiler errors, broken atomics code, etc.) as well as missed optimisations.

This change disables wasm-opt feature of wasm-pack in all Rust codecs and runs the latest wasm-opt manually instead.


codecs/oxipng/build.sh
```diff
@@ -4,8 +4,12 @@ set -e
 
 rm -rf pkg,{-parallel}
 export CFLAGS="${CFLAGS} -DUNALIGNED_ACCESS_IS_FAST=1"
-wasm-pack build -t web
-RUSTFLAGS='-C target-feature=+atomics,+bulk-memory' wasm-pack build -t web -d pkg-parallel -- -Z build-std=panic_abort,std --features=parallel
+wasm-pack build -t web -- --locked
```
surma, line 7: What does `--locked` do?

surma (approved):

codecs/oxipng/build.sh
RReverser: We already use it in default rust.Dockerfile command, just noticed we missed it here. It tells Cargo to use Cargo.lock and not check for any versions compatible with Cargo.toml range, improving reliability of builds.

surma: ah of course

PR #943, opened 2021-01-21
Update HQX to latest

RReverser:
I've played a bit and added a non-invasive change to the HQX - https://github.com/CryZe/wasmboy-rs/pull/1 - to work around the code size regression (https://github.com/rust-lang/rust/issues/74947) introduced in the latest Rust.

In fact, due to the eliminated bound checks, it's now even ~70KB (~30%) smaller than before :)

As another side benefit of the change, the build time also went down significantly and now takes only 1 minute altogether - including spawning Docker, fetching Cargo, building Wasm and optimising it with wasm-opt - instead of 15-20 minutes it took before.

P.S. h/t @CryZe for a very quick review & publish.


surma (approved):
Lovely! LGTM

(Hi @cryze 👋 )

PR #945, opened 2021-01-22
Don't render "OR Paste" when Clipboard API is unsupported

rupesh1:
Hello,

When testing #944, I noticed that although Firefox doesn't support the pasting we still render some copy.

This PR prevents rendering "OR Paste" when Clipboard API is not fully supported.

Demo: https://user-images.githubusercontent.com/3313870/105429245-f1db1680-5c48-11eb-8a61-853560a9ac50.mp4

### DIFF
--- a/src/shared/prerendered-app/Intro/index.tsx
+++ b/src/shared/prerendered-app/Intro/index.tsx
@@ -281,13 +281,14 @@ export default class Intro extends Component<Props, State> {
                 </svg>
               </button>
               <div>
-                <span class={style.dropText}>Drop </span>OR{' '}
-                {supportsClipboardAPI ? (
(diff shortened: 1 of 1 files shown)


RReverser:
LGTM, but I'll defer to @jakearchibald to review.

RReverser:
@jakearchibald Ping - should we merge this?

jakearchibald:
Paste still works without the clipboard API. It works via https://github.com/GoogleChromeLabs/file-drop/blob/master/lib/filedrop.ts#L117. However, it currently seems a bit fussy around which element needs to have focus for paste to work.

We should probably fix that if possible.

Although, maybe the prompt is causing confusion, and we shouldn't mention it if we don't need the button? WDYT @argyleink?

IOLOII (commented):
Merge into fork

PR #947, opened 2021-01-22
Fix browser version of imagequant

### LINKED ISSUE #946: Reduce palette not working
**Describe the bug**
When attempting to use the **reduce palette** feature (on the online version at least), you get an error: "Processing error: ReferenceError: __dirname is not defined"

**To Reproduce**
1. Go to https://squoosh.app/
2. Use one of the existing images.
3. Try to use reduce palette.
4. See error.

**Expected behavior**
To be able to use the reduce palette feature, to reduce the colors.

**Version:**
 - OS w/ version: Windows 10
 - Browser w/ version: Chrome Version 87.0.4280.141 / Firefox 84.0.2

RReverser:
Fix Makefile compiling same file twice and overriding the correct browser version (ENVIRONMENT=$(ENVIRONMENT)) with Node.js one (ENVIRONMENT=node).

Looks like this was introduced by accident in 33d60658cdfc63812cd8e56cf563b7b06a9c6df1.

Fixes #946.


jakearchibald (approved):
LGTM

jakearchibald:
Sending this version to live

surma:
We still need the Node version for the CLI, which is now not being built at all. I think changing the target from `$(OUT_JS)` to `%.js` should make it work as expected. @RReverser can you confirm?

RReverser:
It is still being built, it's part of OUT_JS.

surma:
Oh, than we should remove the stray `$(OUT_NODE_JS)` at some point :D

RReverser:
> Oh, than we should remove the stray `$(OUT_NODE_JS)` at some point :D

Way ahead of you :D

PR #957, opened 2021-02-10
Bump jxl version and remove workaround for enc memory usage.

veluca93:


surma:
Code lgtm! Not sure why there are conflicts? @veluca93 can you make sure your branch is rebased against `dev`?

veluca93:
> Code lgtm! Not sure why there are conflicts? @veluca93 can you make sure your branch is rebased against `dev`?

I was apparently using an older version of dev - it should be OK now.

surma (approved):
LGTM. Gonna wait for @RReverser to take a look too

codecs/jxl/enc/jxl_enc.cpp
```diff
@@ -102,7 +91,7 @@ val encode(std::string image, int width, int height, JXLOptions options) {
   auto result = jxl::ConvertImage(
       jxl::Span<const uint8_t>(reinterpret_cast<const uint8_t*>(image.data()), image.size()), width,
       height, jxl::ColorEncoding::SRGB(/*is_gray=*/false), /*has_alpha=*/true,
-      /*alpha_is_premultiplied=*/false, /*bits_per_sample=*/8, JXL_LITTLE_ENDIAN,
+      /*alpha_is_premultiplied=*/false, /*bits_per_sample=*/8, /*endiannes=*/JXL_LITTLE_ENDIAN,
```
RReverser, line 94: The new comment seems a bit unnecessary - unlike raw values, enum name is fairly self-descriptive.

RReverser (approved):

codecs/jxl/enc/jxl_enc.cpp
veluca93: I think clang-tidy will produce warnings if say an argument gets added in the middle - as this is a C enum value, it would otherwise be silently converted to mean something entirely different...

RReverser: Ideally, I'd change the API of `ConvertImage` to accept a `struct` with named options rather than a long lits of params anyway :) Having lists this long where one needs to remember or manually annotate each argument is usually a code smell.

But I suppose this is not the final API, so it's fine.

PR #965, opened 2021-03-04
Better AVIF options

jakearchibald:
With the "Adaptive quantization mode", only "off" and "variance" seem to make a difference.

"Enable delta quantization" seems to make a really small difference.

I'm going to check these options make sense with the AVIF folks.


codecs/avif/Makefile
```diff
@@ -1,7 +1,9 @@
-CODEC_URL = https://github.com/AOMediaCodec/libavif/archive/31d7c6d1e32cf467ac24fb8c7a76c4902a4c00db.tar.gz
+# Taken from https://source.corp.google.com/piper///depot/google3/third_party/libavif/METADATA
+CODEC_URL = https://github.com/AOMediaCodec/libavif/archive/v0.9.0.tar.gz
 CODEC_PACKAGE = node_modules/libavif.tar.gz
 
-LIBAOM_URL = https://aomedia.googlesource.com/aom/+archive/v2.0.0.tar.gz
+# Taken from https://source.corp.google.com/piper///depot/google3/third_party/libaom/METADATA
+LIBAOM_URL = https://aomedia.googlesource.com/aom/+archive/4a8e276e3e0ef3c76083f3975d5caa85bc9593ce.tar.gz
```
RReverser, line 6: Why? If you just want to update to the latest, https://aomedia.googlesource.com/aom shows 2.0.2. I don't think we care what internal third_party stores (or should link to it as a reference on Github).

jakearchibald: From Wan-Teh Chang:

> Also, we (Jingning in particular) recently improved the libaom encoder significantly to speed up still image encoding and reduce heap memory consumption. I suggest you use a recent stable snapshot of the libaom master branch. A good way is to use the snapshot checked into google3/third_party/libaom, which is roughly two weeks behind the tip of the master branch. You can find the git revision in https://source.corp.google.com/piper///depot/google3/third_party/libaom/METADATA

wantehchang: Ingvar: Although libaom 2.0.2 is the latest official release, it is based on a release branch from May 2020. If Squoosh can use a recent stable snapshot of libaom from the master branch, I recommend it.

codecs/avif/enc/avif_enc.cpp
```diff
@@ -9,10 +9,8 @@ struct AvifOptions {
   // [0 - 63]
   // 0 = lossless
   // 63 = worst quality
-  int minQuantizer;
-  int maxQuantizer;
+  int cqLevel;
   int minQuantizerAlpha;
-  int maxQuantizerAlpha;
```
wantehchang, line 15: If you want to control the quality of the "color" and "alpha" sub-images separately, then I suggest you remove avifEncoderSetCodecSpecificOption() changes from this pull request, because currently avifEncoderSetCodecSpecificOption() sets the option on both the "color" and "alpha" sub-images.

I am planning to fix this limitation of libavif by allowing the "c:" and "a:" prefix so that:
- "cq-level" applies to both "color" and "alpha" sub-images,
- "c:cq-level" applies to the "color" sub-image only,
- "a:cq-level" applies to the "alpha" sub-image only.

codecs/avif/enc/avif_enc.cpp
```diff
@@ -9,10 +9,8 @@ struct AvifOptions {
   // [0 - 63]
   // 0 = lossless
   // 63 = worst quality
-  int minQuantizer;
-  int maxQuantizer;
+  int cqLevel;
```
wantehchang, line 12: I need to find out if cq-level=0 implies lossless compression. I will get back to you.

codecs/avif/enc/avif_enc.cpp
```diff
@@ -49,12 +53,12 @@ val encode(std::string buffer, int width, int height, AvifOptions options) {
...
   avifImage* image = avifImageCreate(width, height, depth, format);
 
-  if (options.maxQuantizer == AVIF_QUANTIZER_LOSSLESS &&
-      options.minQuantizer == AVIF_QUANTIZER_LOSSLESS &&
-      options.minQuantizerAlpha == AVIF_QUANTIZER_LOSSLESS &&
-      options.maxQuantizerAlpha == AVIF_QUANTIZER_LOSSLESS && format == AVIF_PIXEL_FORMAT_YUV444) {
+  if (lossless) {
     image->matrixCoefficients = AVIF_MATRIX_COEFFICIENTS_IDENTITY;
   } else {
     image->matrixCoefficients = AVIF_MATRIX_COEFFICIENTS_BT709;
```
wantehchang: I suggest using AVIF_MATRIX_COEFFICIENTS_BT601 instead because that is the current default matrix coefficients in AVIF.

Detail:

BT.601 matrix coefficients with full range and BT.709 matrix coefficients with limited range are both widely supported.

However, BT.709 matrix coefficients with full range are not supported by older versions of libyuv. Since I prefer using full range in still images, you will get better compatibility if you use BT.601 matrix coefficients with full range.

codecs/avif/enc/avif_enc.cpp
```diff
@@ -69,14 +73,34 @@ val encode(std::string buffer, int width, int height, AvifOptions options) {
   avifImageRGBToYUV(image, &srcRGB);
 
   avifEncoder* encoder = avifEncoderCreate();
+
```
wantehchang, line 76: Change line 67 from:

`  uint8_t* rgba = (uint8_t*)buffer.c_str();`

to:

`  const uint8_t* rgba = (const uint8_t*)buffer.data();`

or better yet:

`  const uint8_t* rgba = reinterpret_cast<const uint8_t*>(buffer.data());`

codecs/avif/Makefile
RReverser: Yeah, quoted explanation in Jake's comment above makes sense to me.

codecs/avif/enc/avif_enc.cpp
jakearchibald: Nice! Could you split color and luma too?

codecs/avif/enc/avif_enc.cpp
jakearchibald: Why's it better? (I'm a C/C++ novice)

wantehchang: I just realized that we do need a non-const pointer here because of a limitation of libavif's C API. So I changed my suggestion to:

`uint8_t* rgba = (uint8_t*)buffer.data();`

or

`uint8_t* rgba = reinterpret_cast<uint8_t*>(const_cast<char*>(buffer.data()));`

Detail:

The `(type)` cast is called a C-style cast in C++. C++ supports C-style casts for compatibility with C, but C++ casts, such as `static_cast<type>()`, `const_cast<type>()`, and `reinterpret_cast<type>()`, are preferred. Each kind of C++ cast is intended for a particular purpose, so it is clear why we are casting.

For example, buffer.c_str() returns `const char*`:

https://en.cppreference.com/w/cpp/string/basic_string/c_str

The `(uint8_t*)` cast not only changes the pointer type but also removes the `const`. C++ casts force you to use two casts to perform these two operations: `const_cast<char*>()` to remove the `const`, followed by `reinterpret_cast<uint8_t*>` to change the pointer type.

In this case, because of the verbosity of C++ casts, you may consider it a plus to use the C-style cast. But in general C++ casts are preferred.

As for `.c_str()` vs. `.data()`, the `.data()` method is not required to return a buffer with a terminating `NUL` character, so it could be more efficient than the `.c_str()` method, and is semantically more correct in this case (because `buffer` contains binary data, not a text string). In practice `.data()` and `.c_str()` often have the same implementation.

codecs/avif/enc/avif_enc.cpp
wantehchang: Here the "color" sub-image refers to the Y, U, V channels, i.e., both luma (Y) and chroma (U, V). This term is used by libavif.

Detail:

The AV1 video codec has either one (Y) and three (Y, U, V) channels. Therefore, an AVIF image with an alpha channel internally consists of two sub-images:
1. A three-channel sub-image, which libavif calls "color". This sub-image provides the Y, U, V channels. **Note**: This sub-image may have only one channel (the Y channel), in which case it is a monochrome image.
2. A one-channel sub-image, which libavif calls "alpha". (The alpha channel is stored in the Y channel of this sub-image.)

jakearchibald: Sure, but is it possible to provide different quality settings to the Y and the UV? MozJPEG allows this and it'd useful for high density images.

RReverser (approved):

PR #969, opened 2021-03-10
Update JPEG XL and support container format in mime sniffing.

veluca93:


jakearchibald:
The empty wasm files seem like a bad sign

veluca93:
> The empty wasm files seem like a bad sign

Indeed. I have no idea why every time the *first* push I do doesn't include the wasm file, but whatever... :D

jakearchibald (approved):
LGTM!

src/client/lazy-app/util/index.ts
```diff
@@ -151,6 +151,7 @@ const magicNumberToMimeType = new Map<RegExp, string>([
   [/^\xF4\xFF\x6F/, 'image/webp2'],
   [/^\x00\x00\x00 ftypavif\x00\x00\x00\x00/, 'image/avif'],
   [/^\xff\x0a/, 'image/jpegxl'],
+  [/^\x00\x00\x00\x0cJXL \x0d\x0a\x87\x0a/, 'image/jpegxl'],
```
sboukortt, line 154: I am a bit confused, why is this not `image/jxl` as in the [provisional registration](https://www.iana.org/assignments/provisional-standard-media-types/provisional-standard-media-types.xhtml)?

jakearchibald: Fix: https://github.com/GoogleChromeLabs/squoosh/pull/970

PR #979, opened 2021-03-22
Don't listen for two-up keyboard shortcuts if in input

jakearchibald:
This was causing weirdness when using text inputs.


developit (approved):

src/client/lazy-app/Compress/Output/custom-els/TwoUp/index.ts
```diff
@@ -98,19 +98,24 @@ export default class TwoUp extends HTMLElement {
 
   // KeyDown event handler
   private _onKeyDown(event: KeyboardEvent) {
+    const target = event.target;
+    if (target instanceof HTMLElement && target.closest('input')) return;
```
RReverser, line 102: I'd probably just check target's `tagName`, it's a bit simpler and `input` doesn't have children anyway.

PR #991, opened 2021-04-21
[CLI] Handle subdirectories without failing

atjn:
When you ask the CLI to minify an entire directory - something like this:
```
npx @squoosh/cli --webp "auto" /path/to/folder/
```
...and the directory contains subdirectories, squoosh will try to read the directories as files, and will fail spectacularly.

In the future, you might want to let the user choose to read the subdirectories recursively or something, but for the time being, this PR fixes the problem by making squoosh ignore any subdirectories.


atjn:
Wow okay, this rebase did not go as planned. I will open a new PR, this one is pretty much destroyed.

PR #995, opened 2021-04-28
Fix alpha regression in resizer

### LINKED ISSUE #980: Resizer transforms opaque images into partially transparent
Apparently one of the reasons for #958 that explains why it happened on Resized images but not original ones is that our Resizer transforms fully opaque image of red panda into one with non-255 alpha values under some params (e.g. resizing to width=1500 in that issue), and transparent images take a different codepath in AVIF.

While AVIF multithreading will be fixed separately, I wonder if this is intentional behaviour of the resizer or result of some precision loss?

Seems unexpected, even thought I guess it might provide some extra smoothing.

(This needs verification that this is indeed what's happening and the issue is not in our AVIF wrapping code.)

surma:
Closes #980


codecs/resize/src/lib.rs
```diff
@@ -133,8 +133,9 @@ pub fn resize(
                 unprocessed_output_image[4 * i + 3],
             ));
         }
-        output_image[4 * i + 3] =
-            (unprocessed_output_image[4 * i + 3] * 255.0).clamp(0.0, 255.0) as u8;
+        output_image[4 * i + 3] = (unprocessed_output_image[4 * i + 3] * 255.0)
+            .round()
+            .clamp(0.0, 255.0) as u8;
```
RReverser, line 138: Maybe I'd invert the order here to `(... as u8).clamp(0, 255)` so that it doesn't operate on floats anymore, but that's merely a microoptimisation and not important :)

RReverser (approved):
Thanks!

codecs/resize/src/lib.rs
RReverser: Lol no this would be even wrong.

jakearchibald:
I think this is worth shipping sooner rather than later. On it.

PR #1003, opened 2021-05-04
Make MozJPEG a WASI module

surma:
Here’s an experiment: I’m using [WASI-SDK] to build MozJPEG instead of Emscripten.

I am not insisting on merging this, but I do think it’s an interesting experiment. Also don’t merge this yet. If we want to move forward with this, I need to update the CLI as well.

## What?

[WASI] is the WebAssembly Systems Interface, a standardized(-ish) interface for WebAssembly modules. An effort started out of the need to run Wasm modules outside the browser _without_ any specific host environment. Instead, a standardized interface that can be implemented by many runtimes and languages.

## Why?

Because it’s a standardized interface, we’d only need _one_ “glue code” to run _all_ Wasm modules that use WASI, for both Web and Node, and all permutations of SIMD, Threads, etc. For C++ there’s the WASI-SDK now, Rust also already supports WASI (both are experimental because the interface itself is still changing).

## Drawbacks

- We lose embind, which gave use the ability to pass in structs and strings. In this instance I am just generating a setter function of each option, and that seems to work okay. But definitely less ergonomics.
- Similarly, we can’t (easily) create JS Typed Arrays from C++ to store the results in, so we need to go back to allocating, copying and then freeing up memory from JS instead (see `mozjpegEncode.ts`).
- No nice auto-generated TypeScript types
- WASI also doesn’t have a standard for threads yet, so that won’t work out-of-the-box. We can probably whip up something ourselves. Maybe.
- WASI-SDK doesn’t provide nice tooling like `emmake` or `emconfigure` to just wrangle all the settings and env variables for you. Figuring out how to compile something is definitely harder (but once you got it, it’s as easy as before).
 
## Results

In this instance, the compressed Wasm binary got roughly 1.5kB bigger, while we can delete about 7KB of module-specific compressed glue code. The bit of glue code that I had to write (see `wasi-utils.ts`) will be reusable for any module that we successfully compile to a WASI module. This could add up to significant savings.

[WASI-SDK]: https://github.com/WebAssembly/wasi-sdk
[WASI]: https://wasi.dev/


surma:
We can actually shave another 5KiB of the gzip’d Wasm binary if we manage to generate this [stupidly large table](https://github.com/mozilla/mozjpeg/blob/master/jpeg_nbits_table.h) at runtime rather than bundle it into the binary. 

@RReverser I tried to monkey-patch it but couldn’t get that to work lol.

src/features/worker-utils/wasi-utils.ts
```diff
@@ -0,0 +1,32 @@
...
+      };
+    },
+  });
+}
+
+export function makeWasiEnv() {
+  return {
+    environ_sizes_get: () => 0,
+    environ_get: () => 0,
+  };
```
sbc100, line 31: Nice.. Is this really all the wasi calls that this code depends on?   I guess you are just doing pure compute.

I'm curious, when you build with emscripten do you see more dependencies than this?   I would hope not.

surma: @sbc100 Well, there are some more WASI imports even in this binary, but they are unused:

<details><pre>
Import[8]:
 - func[0] sig=0 <wasi_snapshot_preview1.proc_exit> <- wasi_snapshot_preview1.proc_exit
 - func[1] sig=4 <wasi_snapshot_preview1.environ_sizes_get> <- wasi_snapshot_preview1.environ_sizes_get
 - func[2] sig=4 <wasi_snapshot_preview1.environ_get> <- wasi_snapshot_preview1.environ_get
 - func[3] sig=19 <wasi_snapshot_preview1.fd_seek> <- wasi_snapshot_preview1.fd_seek
 - func[4] sig=9 <wasi_snapshot_preview1.fd_write> <- wasi_snapshot_preview1.fd_write
 - func[5] sig=1 <wasi_snapshot_preview1.fd_close> <- wasi_snapshot_preview1.fd_close
 - func[6] sig=4 <wasi_snapshot_preview1.fd_prestat_get> <- wasi_snapshot_preview1.fd_prestat_get
 - func[7] sig=2 <wasi_snapshot_preview1.fd_prestat_dir_name> <- wasi_snapshot_preview1.fd_prestat_dir_name
</pre></details>

With Emscripten using `-s STANDALONE_WASM=1`, I see similar WASI imports, some Emscripten specific stuff + lots of embind (couldn’t be bothered to compile a version without embind):

<details><pre>
Import[24]:
 - func[0] sig=20 <env._embind_register_value_object_field> <- env._embind_register_value_object_field
 - func[1] sig=4 <env._embind_register_memory_view> <- env._embind_register_memory_view
 - func[2] sig=6 <env._embind_register_integer> <- env._embind_register_integer
 - func[3] sig=4 <env._embind_register_std_wstring> <- env._embind_register_std_wstring
 - func[4] sig=9 <env._embind_register_function> <- env._embind_register_function
 - func[5] sig=0 <env._emval_decref> <- env._emval_decref
 - func[6] sig=11 <wasi_snapshot_preview1.fd_write> <- wasi_snapshot_preview1.fd_write
 - func[7] sig=4 <env._embind_register_float> <- env._embind_register_float
 - func[8] sig=3 <env._embind_register_std_string> <- env._embind_register_std_string
 - func[9] sig=3 <env._embind_register_void> <- env._embind_register_void
 - func[10] sig=0 <env._emval_incref> <- env._emval_incref
 - func[11] sig=0 <env._embind_finalize_value_object> <- env._embind_finalize_value_object
 - func[12] sig=9 <env._embind_register_value_object> <- env._embind_register_value_object
 - func[13] sig=11 <env._emval_new> <- env._emval_new
 - func[14] sig=0 <env.emscripten_notify_memory_growth> <- env.emscripten_notify_memory_growth
 - func[15] sig=0 <wasi_snapshot_preview1.proc_exit> <- wasi_snapshot_preview1.proc_exit
 - func[16] sig=5 <wasi_snapshot_preview1.environ_get> <- wasi_snapshot_preview1.environ_get
 - func[17] sig=5 <wasi_snapshot_preview1.environ_sizes_get> <- wasi_snapshot_preview1.environ_sizes_get
 - func[18] sig=1 <wasi_snapshot_preview1.fd_close> <- wasi_snapshot_preview1.fd_close
 - func[19] sig=27 <wasi_snapshot_preview1.fd_seek> <- wasi_snapshot_preview1.fd_seek
 - func[20] sig=3 <env._embind_register_emval> <- env._embind_register_emval
 - func[21] sig=6 <env._embind_register_bool> <- env._embind_register_bool
 - func[22] sig=1 <env._emval_get_global> <- env._emval_get_global
 - func[23] sig=2 <env.__cxa_thread_atexit> <- env.__cxa_thread_atexit
</pre></details>

sbc100: > @sbc100 Well, there are some more WASI imports even in this binary, but they are unused:

Doesn't this imply there is a bunch of dead code in there?   I imagine that are building here is really what WASI calls a reactor.   Its still pretty experimental but you could try building with `-mexec-model=reactor`.   It works kind of like the `--no-entry` emscripten flag (which you should be using with emscripten if you are not already).    This should eliminate a bunch of those imports I hope (and remove a bunch of dead code).

> 
> With Emscripten using `-s STANDALONE_WASM=1`, I see similar WASI imports, some Emscripten specific stuff + lots of embind (couldn’t be bothered to compile a version without embind):

RReverser:
It's certainly an interesting experiment. I think my only scepticism is whether

> This could add up to significant savings.

is really true if you consider sizes of both Wasm and JS. Normally, JS glue code is still a tiny portion of the size of Wasm itself - even in MozJPEG case in this PR, the Wasm is ~77KB while JS was 7KB (IIUC your description correctly). In case of something like AVIF Wasm is a lot larger, so portion of JS is even smaller.

Those uncompressed numbers are probably not very useful on their own, but it could be interesting to estimate and compare savings for compressed sizes before trying this route for all the other codecs.

surma:
> Those uncompressed numbers are probably not very useful on their own, but it could be interesting to estimate and compare savings for compressed sizes before trying this route for all the other codecs.

Sorry, I think I deleted some words in my post by accident. Those _are_ the gzip’d numbers. Without compression, the Wasm got 80K larger while the old glue code was ~19K minified.

surma:
To add to that: The Wasm binaries (WASI vs Emscripten) are pretty much identical in size _if I don’t add the setter methods for the options object_. So the 80K probably come from just those setters and compress really well. 

_edit:_ I might have to re-evaluate this. As I should have expected, the setter functions are super tiny, so not sure where the 80K are really coming from. But either way gzip’d we still have a net positive.

_edit #2_: It was `-flto`. See below.

src/features/worker-utils/wasi-utils.ts
surma: Tried `-mexec-model=reactor`, the imports remain. I guess there are just some dynamic code branches that we never take but that can’t be fully eliminated.

surma:
Minor update:

- Discovered that `wasm-opt` has an `-O4`
- ~~Also invoked `wasm-strip` after build~~

With this, the resulting gzip’d wasm binary is 1.1kB **smaller** than the Emscripten build.

codecs/mozjpeg/Makefile
```diff
@@ -27,10 +27,12 @@ $(OUT_WASM): $(CODEC_OUT)
 		-I $(CODEC_DIR) \
 		${CXXFLAGS} \
 		${LDFLAGS} \
+		-mexec-model=reactor \
 		-o $@ \
 		enc/mozjpeg_enc.cpp \
 		$(CODEC_OUT)
-	$(BINARYEN_PREFIX)/bin/wasm-opt -O3 -o $(OUT_WASM) $(OUT_WASM)
+	$(BINARYEN_PREFIX)/bin/wasm-opt -O4 -o $(OUT_WASM) $(OUT_WASM)
+	$(WABT_PREFIX)/bin/wasm-strip $(OUT_WASM)
```
sbc100: I think wasm-opt can do everything that wasm-strip can do so you shouldn't need both tools

sbc100:
> Minor update:
> 
> * Discovered that `wasm-opt` has an `-O4`
> * Also invoked `wasm-strip` after build
> 
> With this, the resulting gzip’d wasm binary is 1.1kB **smaller** than the Emscripten build.

What emscripten opt level are you comparing too?   (Can you try with `-Oz` or `-Os`..  there are certainly think emscripten can do such as minify import/export names which wasi-sdk can't do.. it in theory it should always be smaller)

sbc100:
(I guess the benefit of import/export name compression is somewhat hidden by the gzip compression)

codecs/mozjpeg/Makefile
surma: @sbc100 Oh, I missed those flags. `--strip-debug --strip-dwarf --strip-producers --strip-target-features` seems to achieve the same thing. Cool!

surma:
> What emscripten opt level are you comparing to

[`-O3`](https://github.com/GoogleChromeLabs/squoosh/blob/efc89efba54a99ff7dc48b2bbe81ae8f1b2915b7/codecs/cpp.Dockerfile#L3)

>  it in theory it should always be smaller

I am comparing to the embind build, not standalone. So there are bunch more imports. (But yes, longer function names will mostly caught by gzip I reckon).

surma:
I forgot about `-flto`!! With that, the WASI build is smaller with and without gzip.

```
-rw-r--r--    1 surma  primarygroup   236147 May  4 18:39 mozjpeg_em.wasm
-rw-r--r--    1 surma  primarygroup    55379 May  4 18:39 mozjpeg_em.wasm.gz
-rwxr-xr-x    1 surma  primarygroup   235379 May  4 19:09 mozjpeg_wasi.wasm
-rwxr-xr-x    1 surma  primarygroup    53753 May  4 19:09 mozjpeg_wasi.wasm.gz
```

PR #1004, opened 2021-05-05
Custom progressive steps for MozJPEG

jakearchibald:
These progressive steps appear to play better with the DCT smoothing currently in Firefox and Chrome. However, I think there's a decoder bug right now, so it's possible we might not need this.

Or, maybe it should be upstreamed to MozJPEG itself?


jakearchibald:
I'm going to let this rest until the libjpeg-turbo decoder bugs are gone, but I still think there'll be a benefit to this.

Btw, https://static-misc-3.glitch.me/slow-img-load/ lets you slow-load images (you can select multiple)

codecs/mozjpeg/enc/mozjpeg_enc.cpp
```diff
@@ -160,7 +161,84 @@ val encode(std::string image_in, int image_width, int image_height, MozJpegOptio
...
+      scanptr->Ah = 1;
+      scanptr->Al = 0;
+      scanptr++;
+      scanptr->comps_in_scan = 1;
+      scanptr->component_index[0] = 1;
+      scanptr->Ss = 1;
+      scanptr->Se = 63;
+      scanptr->Ah = 1;
+      scanptr->Al = 0;
+      scanptr++;
```
jakearchibald, line 240: I'll extract these into a little helper function once we've got the output looking right

PR #1007, opened 2021-05-06
wasm-bindgen-rayon and new OMT plugin

RReverser:
Few months ago, I've extracted and published universal Rust + Rayon adapter as a separate library: https://github.com/GoogleChromeLabs/wasm-bindgen-rayon. It means we no longer need to support our own custom glue code in Squoosh, but can use it directly as a dependency, which is what this PR does.

However, in the process of working on wasm-bindgen-rayon I also made a series of changes to make it work with rollup-plugin-off-main-thread. Those changes made it closer to real AMD, as well as reduced size of the generated code, made it simpler etc., but, unfortunately, turned out to be incompatible with some of the tricks we use in Squoosh - in particular, defining multiple modules inline in a single HTML page.

Previously this worked because OMT was replacing Rollup-generated anonymous `define(` calls with named entry chunks, but that caused other problems when using multiple AMD modules on the same page, so I've removed that on the OMT side and made all the `define`s anonymous again. To preserve current behaviour on Squoosh side, I copied the updated loader code, and exposed an internal `nextDefineUri` as a global variable. This way, build system can emit `nextDefineUri="..."` before each `define(` call, which, in turn, will allow multiple defines on the same page again. This is a hack, but it works and preserves existing behaviour without having to extract everything into separate scripts as Rollup normally expects.


lib/client-bundle-plugin.js
```diff
@@ -137,10 +137,24 @@ export default function (inputOptions, outputOptions, resolveFileUrl) {
...
+          clientEntry,
+          ...dependencies.map((name) =>
+            clientOutput.find((item) => item.fileName === name),
+          ),
+        ];
+
+        const inlineDefines = [
+          ...allModules.map(
+            (item) =>
+              `nextDefineUri=location.origin+${resolveFileUrl(item)};${
```
RReverser: @jakearchibald Quick question: here I'm just calling `resolveFileUrl(item)` since our `resolveFileUrl` only needs `fileName` property anyway, but I see that other calls in the same file use

```js
        return resolveFileUrl({
          fileName: clientEntry.fileName,
          moduleId,
          format,
        });
```

or

```js
            return resolveFileUrl({
              fileName: entry.fileName,
              moduleId,
              format: outputOptions.format,
            });
```

instead of just passing object directly.

Should I do the same?

src/features/encoders/oxiPNG/worker/oxipngEncode.ts
```diff
@@ -77,7 +40,7 @@ export default async function encode(
   options: EncodeOptions,
 ): Promise<ArrayBuffer> {
   if (!wasmReady) {
-    wasmReady = (await threads()) ? initMT() : initST();
+    wasmReady = threads().then((hasThreads: boolean) => hasThreads ? initMT() : initST());
```
RReverser, line 43: Note: this is unrelated fix, just noticed that we weren't assiging Promise immediately, which could result in bugs.

rollup.config.js
```diff
@@ -37,13 +37,6 @@ function resolveFileUrl({ fileName }) {
   return JSON.stringify(fileName.replace(/^static\//, '/'));
 }
 
-// With AMD output, Rollup always uses document.baseURI, which breaks in workers.
-// This fixes it:
-function resolveImportMeta(property, { chunkId }) {
```
RReverser, line 42: This is no longer necessary because OMT plugin takes care of `import.meta.url` directly now.

lib/css-plugin.js
```diff
@@ -48,7 +48,7 @@ const appendCssSource = `
   }
 `;
 
-export default function (resolveFileUrl) {
+export default function () {
```
RReverser, line 51: This param was unused so I just removed it.

codecs/oxipng/build.sh
```diff
@@ -6,6 +6,4 @@ rm -rf pkg,{-parallel}
 export CFLAGS="${CFLAGS} -DUNALIGNED_ACCESS_IS_FAST=1"
 wasm-pack build -t web
 RUSTFLAGS='-C target-feature=+atomics,+bulk-memory' wasm-pack build -t web -d pkg-parallel -- -Z build-std=panic_abort,std --features=parallel
-# Workaround https://github.com/rustwasm/wasm-bindgen/issues/2133:
-sed -i "s|maybe_memory:|maybe_memory?:|" pkg-parallel/squoosh_oxipng.d.ts
```
RReverser, line 10: Also unrelated fix: I've fixed typings upstream in wasm-bindgen, so this hack is no longer necessary.

surma:
I’m not sure what broke, but staging version doesn’t work :D

RReverser:
Hm, it worked locally, let me recheck.

(If only we had tests... :P)

RReverser:
Heh, it's something that I first assumed wouldn't work, but then tried with `npm run dev` and it did, so I thought it's fine, but apparently it fails only in optimized builds. Fixing...

RReverser:
Ah no, that's some minifier issue where it doesn't see one of the variable usages. Sigh.

RReverser:
"Fixed".

surma (approved):

jakearchibald (changes requested):
There's something not quite right here. With the current version of squoosh.app we ensure that the HTML contains module definitions for everything needed for the first interaction, so we can get to first interaction without external dependencies.

However, this version introduces https://deploy-preview-1007--squoosh.netlify.app/c/logo-with-text.svg-8d769ff2.js as an external module needed for first-interaction.

The HTML contains the definition for this module, but it doesn't appear to be used.

lib/client-bundle-plugin.js
```diff
@@ -137,10 +137,24 @@ export default function (inputOptions, outputOptions, resolveFileUrl) {
...
+          clientEntry,
+          ...dependencies.map((name) =>
+            clientOutput.find((item) => item.fileName === name),
+          ),
+        ];
+
+        const inlineDefines = [
+          ...allModules.map(
+            (item) =>
+              `self.nextDefineUri=location.origin+${resolveFileUrl(item)};${
```
jakearchibald, line 150: I know we aren't governed by https://url.spec.whatwg.org/#goals, but maybe worth considering:

> Standardize on the term URL. URI and IRI are just confusing. In practice a single algorithm is used for both so keeping them distinct is not helping anyone.

jakearchibald:
Additionally, when I `npm run dev`, and load the result in a browser, I get `(index):59 Uncaught TypeError: depsNames.map is not a function`

Edit: I hadn't ran `npm i`, oops

RReverser:
> However, this version introduces [deploy-preview-1007--squoosh.netlify.app/c/logo-with-text.svg-8d769ff2.js](https://deploy-preview-1007--squoosh.netlify.app/c/logo-with-text.svg-8d769ff2.js) as an external module needed for first-interaction.

Hmm, good catch. Maybe it's not correctly defined then, let me recheck.

lib/client-bundle-plugin.js
RReverser: > Hmm, good catch. Maybe it's not correctly defined then, let me recheck.

Hah, right, it's actually something I correctly handled first, but then reverted because app seemed to work fine anyway. Fixed.

RReverser:
> Hmm, good catch. Maybe it's not correctly defined then, let me recheck.

Hah, right, it's actually something I correctly handled first, but then reverted because app seemed to work fine anyway. Fixed.

jakearchibald:
I could also put the dependencies first in the HTML, but I'm a little worried that dependencies-of-dependencies might go wrong.

lib/omt.ejs
```diff
@@ -13,80 +13,61 @@
...
-        return exports;
-      });
+    let exports = {};
+    const require = depUri => singleRequire(depUri, uri);
+    const specialDeps = {
+      module: { uri },
+      exports,
+      require
+    };
+    registry[uri] = Promise.resolve().then(() => Promise.all(depsNames.map(
```
jakearchibald, line 69: Comment, so we don't delete this in future 😄

rollup.config.js
jakearchibald: so good!

lib/omt.ejs
RReverser: Done.

lib/client-bundle-plugin.js
RReverser: I guess this makes sense; however the right place to change it would be upstream in https://github.com/surma/rollup-plugin-off-main-thread/blob/master/loader.ejs first, as this name comes from there (and I want to keep Squoosh loader not _too_ different from upstream one to make future updates easier).

jakearchibald (approved):
Love it. I know this was a pain in the arse. I wish there was an easier, standard way to inline modules.

lib/client-bundle-plugin.js
jakearchibald: I think it's fine. If we change format there'll be a lot of refactoring anyway.

RReverser:
> I wish there was an easier, standard way to inline modules.

In theory, Rollup should really take care of this - as in, one should be able to say "if module size is < given threshold, inline it like you would if it wasn't code-split".

surma:
> I wish there was an easier, standard way to inline modules.

OH DO I HAVE SOMETHING FOR YOU

RReverser:
> OH DO I HAVE SOMETHING FOR YOU

WELL DO YOU?!

surma:
Needed to finish the post :D

https://github.com/littledan/proposal-module-fragments/issues/5#issuecomment-840562415

PR #1008, opened 2021-05-07
Updating AVIF, and 'auto' SSIM option

jakearchibald:
I'm getting a build error right now, but it appears to have worked anyway:

```
/src/node_modules/libavif/src/codec_aom.c:41:9: error: macro is not used [-Werror,-Wunused-macros]
#define HAVE_AOM_CODEC_SET_OPTION 1
```


surma:
Delegating this to @RReverser, esp wrt build error

RReverser:
I think this is just an upstream error - please report it there. I've seen something similar before, as the config we're using is somewhat different from one they're normally testing with.

jakearchibald:
Yeah, the error is in libavif, and it's surfacing because we're splitting the encoder and decoder. It's the decoder that's failing to build, which isn't a big deal.

They're going to fix it, but it shouldn't stop us landing this.

surma (approved):

PR #1014, opened 2021-05-12
Updating oxi, adding interlace option

jakearchibald:
This is what I used for the HTTP 203 episode 😄


codecs/oxipng/src/lib.rs
```diff
@@ -5,14 +5,15 @@ use wasm_bindgen::prelude::*;
...
-pub fn optimise(data: &[u8], level: u8) -> Vec<u8> {
+pub fn optimise(data: &[u8], level: u8, interlace: u8) -> Vec<u8> {
     let mut options = oxipng::Options::from_preset(level);
     options.alphas.insert(AlphaOptim::Black);
     options.alphas.insert(AlphaOptim::White);
     options.alphas.insert(AlphaOptim::Up);
     options.alphas.insert(AlphaOptim::Down);
     options.alphas.insert(AlphaOptim::Left);
     options.alphas.insert(AlphaOptim::Right);
+    options.interlace = Some(interlace);
```
surma: I’d rather make the `interlace` parameter a `bool` and do this here than in JS:

```suggestion
    options.interlace = Some(if interlace { 1 } else { 0 });
```

surma (approved):
One suggestion, rest lgtm

codecs/oxipng/src/lib.rs
RReverser: Huh, why isn't it a bool upstream? Are there other modes `>=2` we might care about?

RReverser: Also... are `Some(false)` and `None` different then? It seems a bit confusing if so.

jakearchibald: None means "match the input", which is always the same as false for Squoosh. Other numbers just fail.

RReverser: > Other numbers just fail.

As in, they fail specifically in Squoosh or they are not meant to be supported? If the latter, then "why isn't it a bool upstream" still stands - should we raise an issue there to change it?

jakearchibald: There are three states for interlace in oxipng:

- `Some(0)` - Output is not interlaced
- `Some(1)` - Output is interlaced
- `None` - If the input is a PNG and it's interlaced, output is interlaced, otherwise the output is not interlaced

I don't know enough about rust to know if this is a good pattern or not. `Some(2)` fails at runtime, but I can't figure out why or where, but it definitely isn't supported.

RReverser: Ok. Since it's not supported, let's update here as per Surma's suggestion, and raise an issue upstream to change API to use either a bool or a more descriptive enum instead of raw u8.

src/features/encoders/oxiPNG/worker/oxipngEncode.ts
```diff
@@ -40,9 +40,12 @@ export default async function encode(
   options: EncodeOptions,
 ): Promise<ArrayBuffer> {
   if (!wasmReady) {
-    wasmReady = threads().then((hasThreads: boolean) => hasThreads ? initMT() : initST());
+    wasmReady = threads().then((hasThreads: boolean) =>
+      hasThreads ? initMT() : initST(),
+    );
```
jakearchibald, line 45: Hah, I guess prettier didn't pick this up last time

jakearchibald:
Filed https://github.com/shssoichiro/oxipng/issues/403

src/features/encoders/oxiPNG/worker/oxipngEncode.ts
RReverser: Weirddd.

RReverser (approved):

PR #1017, opened 2021-05-14
Basis Universal

### LINKED ISSUE #642: Codec Support Query: Basis Universal
Just noticed ["Basis Universal"](https://github.com/binomialLLC/basis_universal) pop up on [my twitter feed](https://mobile.twitter.com/Lady_Ada_King/status/1144546463365681152)
> ...a very compressed image format that are smaller to download will allow 6-8 times more textures in our WebGL scenes, enabling us to build richer worlds at higher fidelity

Was wondering if (in general) image formats that are targeted mostly at WebGL are in or out-of-scope for Squoosh, and (specifically) if this would be something that Squoosh could support?

Yes/No on "is this format out-of-scope for Squoosh" ?

**Does other service/app have this feature?**
* Three.js

**Additional context**
* [Using Basis Textures in Three.js](https://medium.com/samsung-internet-dev/using-basis-textures-in-three-js-6eb7e104447d) by @adarosecannon
* #608

surma:
Closes #642 

Basis is a texture format that aims to easily _transcode_ to memory-preserving, GPU-native data formats.

I have only exposed the “quality” option so far, because I don’t know which other ones would be useful. But even that option doesn’t seem to do much. 

## TODO

- [x] `Makefile` (or emcc?) deletes all object files in the end. I have no idea why.
- [x] `Makefiles` stops after download when run for the first time. I’m guessing I got my dependency chain wrong somewhere
- [x] Expose more options, but get input from some people who know this stuff first.
- [x] Double-check the mime type is appropriate (`image/basisu`)
- [x] Double-check the mime type sniffer is appropriate


jakearchibald:
Ohh exciting!

The quality slider doesn't seem to do anything right now, it always seems lossless.

I threw this image in it and it looks like a width is wrong somewhere:

Screenshot 2021-02-02 at 13 35 05

Output:

Screenshot 2021-02-02 at 13 35 05

surma:
Yeah, I’m trying to get help on the options.

The width thing is interesting. It worked for my test images just fine. Wonder if some for of stride heuristic is going wrong there

richgel999:
Which options do you need help with? Also, what formats are you adding support for (ETC1S and/or UASTC)?

surma:
@richgel999 Thanks for offering to help. Really appreciate it.

I did some more work last night and now support both ETC1S and UASTC, which is currently hidden under “Advanced options”. The UI is just thrown together for now, I’ll fix that once I sorted everything on the codec end.

What currently need help with:
- I currently detect `.basis` when a binary file starts with `sB`. Is that sufficient?
- Do you have an official(-ish) mime type for Basis? I went with `image/basisu` for now.
- As @jakearchibald reported above, with _some_ images the side seems to go wrong. I haven’t investigated yet whether that’s happening in the encoder or in the decoder.
- I added an option to embed mipmaps (`params.m_mip_gen = true`), but that throws an assertion `Assertion failed: (v >= 0) && (v < h), at: node_modules/basis/encoder/basisu_resampler.cpp,33,resampler_range_check)`. Not sure what I am missing here.

richgel999:
> I currently detect .basis when a binary file starts with sB. Is that sufficient?

You could also check the first 4 bytes for 's', 'B', 0x10, 0x00. 

> Do you have an official(-ish) mime type for Basis? I went with image/basisu for now.

No, we haven't worked on that yet.

> As @jakearchibald reported above, with some images the side seems to go wrong. I haven’t investigated yet whether that’s happening in the encoder or in the decoder.

I strongly suspect this is happening outside of our code somewhere. If you can encode the textures using our command line encoder, or the WASM encoder, and it transcodes OK, it's most likely either an issue with how the encoder is being called or how the transcoded data is interpreted. Just looking at the image above, it looks like a bytes per image row (pitch) issue, either on the source side or on the output (transcode) side. 

I would try encoding a variety of small checkboard textures of various really small sizes (like 1x1, 2x2, 3x3, 4x4, 5x5, etc.) and see if any particular size triggers the bug. Some image classes/libraries like to align their image rows to various boundaries, like 16 bytes. The "image" class in the encoder (https://github.com/BinomialLLC/basis_universal/blob/master/encoder/basisu_enc.h) does have minimal support for custom pitches that are larger than the image width, but I didn't expose that very well. If this is the issue, I can either add more support for custom pitches on my end, or you can copy the image data into a temporary non-aligned image.

> I added an option to embed mipmaps (params.m_mip_gen = true), but that throws an assertion Assertion failed: (v >= 0) && (v < h), at: node_modules/basis/encoder/basisu_resampler.cpp,33,resampler_range_check). Not sure what I am missing here.

Strange - I've never seen this exception. I suspect the resampler is somehow being called with some bad parameters. If you set "m_debug" to true, all of the input parameter values (in the basis_compressor_params struct) will be printed to stdout. If you could copy & paste them here, I could help debug it. 

There's only one place where the resampler is called for mipmaps, in basis_compressor::generate_mipmaps(). It calls image_resample, which then calls the low level resampler. If you could show me the parameters to these methods, maybe I could spot something.

As a sanity check, I saved the image above to gtest.png and just compressed it using either "basisu -mipmap gtest.png" and "basisu -mipmap -uastc gtest.png", using a Windows debug build, and it encoded ok (no assertions). I also unpacked the .basis files to .PNG/.KTX's ("basisu gtest.basis"), and the output looks good. 

One thing you can do to help debug your "side" problem is to encode using the command line tool, and then transcode that generated file using your code. The command line tool is just a thin wrapper overtop of the library, so it can be useful to tracking down issues like this.

richgel999:
Also, unrelated to your current issues: You can now encode to .KTX2 instead of .basis by setting the "m_create_ktx2_file" flag to true during encoding, then calling basis_compressor::get_output_ktx2_file() after compression. You'll have to use the "ktx2_transcoder" class instead to transcode, but the API is almost the same as transcoding .basis files. 

There are a few advantages to using KTX2: it's a standardized file format, you can embed arbitrary key value data in the file, and it supports built-in Zstd supercompression of UASTC data. The nitty gritty compressed data inside of .basis files is defined by the Khronos KTX2 standard, but the file format headers are not. With KTX2, the entire thing is standardized. Performance wise they should be basically the same because internally each format goes into the same low-level API's.

If you're working on a tool that will be exposed to end users, I would recommend supporting .KTX2 instead of .basis, unless you are positive you don't want KTX2 for some reason.

richgel999:
Also, looking at your .cpp code which calls the compressor, I see at least one issue:
https://github.com/GoogleChromeLabs/squoosh/blob/2e3361af797a19992c71d1ac1cea0a5242f9ef83/codecs/basis/enc/basis_enc.cpp

Currently, the compressor should be passed in a pointer to an instance of a basist::etc1_global_selector_codebook. (We haven't tested not passing in a valid codebook.) Here's some example code, from our JavaScript encode/transcode wrappers:

https://github.com/BinomialLLC/basis_universal/blob/master/webgl/transcoder/basis_wrappers.cpp#L1051

This is actually something we're completely removing really soon (because it's a feature we removed from the file format to ease standardization).

Another thing to be aware of: when compiling to WASM, be sure to disable multithreading in the encoder, using code like in basis_wrappers.cpp. We had so many threading issues that we disabled it completely in WASM builds until the platform supports it better.

There are runnable demos of the WASM .basis/.ktx2 encoder in the repo here:
https://github.com/BinomialLLC/basis_universal/tree/master/webgl

(Sorry calling the compressor directly is so fiddly. Most of our users either use the command line tool or call our WASM wrappers from JavaScript code. The KTX2 folks do call it directly though, so we know it works.)

surma:
Thanks for all the input @richgel999! I switched to KTX2. Standardized format sounds good :D

> > I currently detect .basis when a binary file starts with sB. Is that sufficient?

> You could also check the first 4 bytes for 's', 'B', 0x10, 0x00.

I was seeing files starting with 's', 'B', 0x13, 0x00. Although this is kinda irrelevant now because I switched to KTX2 anyway.

```
xxd -g 1 ~/Downloads/squoosh.basis 
00000000: 73 42 13 00 4d 00 9b 82 33 3e 00 00 9f 27 02 00  sB..M...3>...'..
...
```

**Re. @jakearchibald’s error**

I encoded the image with Squoosh and decoded it using the `basisu` CLI, and it looked okay, so it must have been in the decoder. I am now explicitly setting the stride (`output_row_pitch_in_blocks_or_pixels`) and that seems to fix the problem.

> Strange - I've never seen this exception. I suspect the resampler is somehow being called with some bad parameters. If you set "m_debug" to true, all of the input parameter values (in the basis_compressor_params struct) will be printed to stdout. If you could copy & paste them here, I could help debug it.

So `m_debug = true` didn’t do much. I recompiled with `-DBASISU_FORCE_DEVEL_MESSAGES=1` and now this is the output in the console:

<details>
<pre>
basisu_transcoder::basisu_transcoder_init: Initializing (this is not an error)
basis_compressor::basis_compressor
basis_compressor::init
basis_compressor::init:
Has global selector codebook: 1
Source images: 1, source filenames: 0, source alpha filenames: 0
m_uastc: 0 1
m_y_flip: 0 0
m_debug: 1 1
m_validate: 0 0
m_debug_images: 0 0
m_global_sel_pal: 0 0
m_auto_global_sel_pal: 0 0
m_compression_level: 2 1
m_no_hybrid_sel_cb: 0 0
m_perceptual: 1 0
m_no_endpoint_rdo: 0 0
m_no_selector_rdo: 0 0
m_read_source_images: 0 1
m_write_output_basis_files: 0 0
m_compute_stats: 0 0
m_check_for_alpha: 1 0
m_force_alpha: 0 0
swizzle: 0,1,2,3
m_renormalize: 0 0
m_multithreading: 1 0
m_disable_hierarchical_endpoint_codebooks: 0 0
m_hybrid_sel_cb_quality_thresh: 2.000000 0
m_global_pal_bits: 8 0
m_global_mod_bits: 8 0
m_endpoint_rdo_thresh: 1.500000 0
m_selector_rdo_thresh: 1.250000 0
m_mip_gen: 1 1
m_mip_renormalize: 0 0
m_mip_wrapping: 1 0
m_mip_fast: 1 0
m_mip_srgb: 0 0
m_mip_premultiplied: 1.000000 0
m_mip_scale: 1.000000 1
m_mip_smallest_dimension: 1 0
m_mip_filter: kaiser
m_max_endpoint_clusters: 0
m_max_selector_clusters: 0
m_quality_level: 128
m_tex_type: 0
m_userdata0: 0x0, m_userdata1: 0x0
m_us_per_frame: 0 (0.000000 fps)
m_pack_uastc_flags: 0x2
m_rdo_uastc: 0 0
m_rdo_uastc_quality_scalar: 1.000000 0
m_rdo_uastc_dict_size: 4096 0
m_rdo_uastc_max_allowed_rms_increase_ratio: 10.000000 0
m_rdo_uastc_skip_block_rms_thresh: 8.000000 0
m_rdo_uastc_max_smooth_block_error_scale: 10.000000 0
m_rdo_uastc_smooth_block_max_std_dev: 18.000000 0
m_rdo_uastc_favor_simpler_modes_in_rdo_mode: 1 0
m_rdo_uastc_multithreading: 1 0
m_resample_factor: 0.000000 0
Has global codebooks: 0
m_create_ktx2_file: 1 1
KTX2 UASTC supercompression: 0
KTX2 Zstd supercompression level: 6
KTX2 sRGB transfer func: 0
Total KTX2 key values: 0
basis_compressor::process
basis_compressor::read_source_images
Source image index 0 filename  600x600 has alpha: 1
Any source image has alpha: 1
basis_compressor::generate_mipmaps
Assertion failed: (v >= 0) && (v < h), at: node_modules/basis/encoder/basisu_resampler.cpp,33,resampler_range_check
</pre>
</details>

Is that giving you anything useful?

richgel999:
Yes that's very useful.
Just to be sure, you're compiling to WebAssembly right? If so you must disable multithreading in the compressor params (set "m_multithreading" to false), and create a job pool like this:

image

I'm not 100% positive yet, but I suspect what's happening is you are falling into an untested code path (one that our JS wrappers don't currently use). I will add checks for this on our end, to make sure the caller doesn't try to enable threading in WASM. 

For now, if you check out the basis_encoder::encode() method in webgl/transcoder/basis_wrappers.cpp (which also has encoder wrappers too), you'll see exactly how our JS wrappers call the compressor in WASM:
https://github.com/BinomialLLC/basis_universal/blob/master/webgl/transcoder/basis_wrappers.cpp

richgel999:
Also, looking at how you're calling the compressor here, it doesn't look like you're passing in a pointer to a valid job pool or a global selector codebook (like the code does in our JavaScript wrappers in the image I posted previously):

https://github.com/GoogleChromeLabs/squoosh/blob/558ee0e5bad7690c50432148bddfe8bdd9ccb87a/codecs/basis/enc/basis_enc.cpp

This internal API needs to be called just like the way our JavaScript wrapper does it (in the image above), otherwise you're going to fall off the rails. I'll shore up this API, which was formerly internal, to be much more paranoid.

Be sure to also "set m_write_output_basis_files" to false.

richgel999:
> I was seeing files starting with 's', 'B', 0x13, 0x00. Although this is kinda irrelevant now because I switched to KTX2 anyway.

Whoops! Yes we revved the .basis file format. The initial release was 0x10. Long term we're focusing on KTX2, but we're keeping .basis in case we ever work with other non-Khronos standards bodies.

richgel999:
I'm trying to hack our JS wrappers to match your parameters, in particular multithreading and the job pool, just to be sure you're falling into an untested code path:

image

richgel999:
I'm not positive, but another potential issue is you need to compile all of our .cpp files with no strict aliasing (the same as the Linux kernel):
https://github.com/GoogleChromeLabs/squoosh/blob/114d6869ea26422147f0344c0f1b76b269ebdcf8/codecs/basis/Makefile

Use "-fno-strict-aliasing -O3".

(I just compiled the encoder without this option to WASM and it still *seems* to work, but we don't test at all with strict aliasing disabled.)

richgel999:
Unfortunately, I've not been able to reproduce your problem yet. I tried a 600x600 texture, enabling threading in WASM, not passing in a job pool, and compiling without strict aliasing. Hmm. 

I'm attempting to build and run squoosh locally, under Windows Subsystem for Linux v1 using Ubuntu v18. No idea if this will work, but I'll try.

richgel999:
As an experiment, you could try setting the mipmap filter kernel to "box", and see how happens. Also if you could give me a callstack and function parameters I might be able to figure out what's happening. (That resampler code is very, very ancient so this is a head-scratching problem.)

richgel999:
Ok - I have squoosh (but not your PR yet) compiling and running here using WSL v1. I should be able to reproduce this problem, which I'll work on tomorrow (5/20). I'll help you get this working as best I can.

surma:
Again, thanks for all that insight, @richgel999. I really appreciate it.

I forgot to push a commit yesterday, actually. I already had `m_multithreading = false` and fixed the `m_write_output_basis_files` typo and added the proper codebook code. That commit is now pushed here :)

I also now added the job pool (I assumed `m_multithreading = false` was sufficient) and added the `-fno-strict-aliasing` compiler flag.

I don’t have an easy way of getting you function parameters, but I did push another commit that enables debug symbols, so you can at least have a stack trace:

```
Uncaught (in promise) RuntimeError: abort(Assertion failed: (v >= 0) && (v < h), at: node_modules/basis/encoder/basisu_resampler.cpp,33,resampler_range_check). Build with -s ASSERTIONS=1 for more info.
    at abort (http://localhost:5000/c/basis_enc-a6a32c87.js:631:13)
    at ___assert_fail (http://localhost:5000/c/basis_enc-a6a32c87.js:826:9)
    at basisu::resampler_range_check(int, int) (http://localhost:5000/c/basis_enc-2d986517.wasm:wasm-function[3154]:0x8de35)
    at basisu::Resampler::put_line(float const*) (http://localhost:5000/c/basis_enc-2d986517.wasm:wasm-function[806]:0x281c2)
    at basisu::image_resample(basisu::image const&, basisu::image&, bool, char const*, float, bool, unsigned int, unsigned int) (http://localhost:5000/c/basis_enc-2d986517.wasm:wasm-function[283]:0x152e2)
    at basisu::basis_compressor::generate_mipmaps(basisu::image const&, basisu::vector<basisu::image>&, bool) (http://localhost:5000/c/basis_enc-2d986517.wasm:wasm-function[289]:0x15a0a)
    at basisu::basis_compressor::read_source_images() (http://localhost:5000/c/basis_enc-2d986517.wasm:wasm-function[251]:0xfd84)
    at basisu::basis_compressor::process() (http://localhost:5000/c/basis_enc-2d986517.wasm:wasm-function[250]:0xf51d)
    at encode(std::__2::basic_string<char, std::__2::char_traits<char>, std::__2::allocator<char> >, int, int, BasisOptions) (http://localhost:5000/c/basis_enc-2d986517.wasm:wasm-function[3772]:0x138c3b)
    at emscripten::internal::Invoker<emscripten::val, std::__2::basic_string<char, std::__2::char_traits<char>, std::__2::allocator<char> >, int, int, BasisOptions>::invoke(emscripten::val (*)(std::__2::basic_string<char, std::__2::char_traits<char>, std::__2::allocator<char> >, int, int, BasisOptions), emscripten::internal::BindingType<std::__2::basic_string<char, std::__2::char_traits<char>, std::__2::allocator<char> >, void>::'unnamed'*, int, int, BasisOptions*) (http://localhost:5000/c/basis_enc-2d986517.wasm:wasm-function[3771]:0x13897d)
```

richgel999:
Okay - I'm grabbing your PR and try building. I'm not familiar with your project yet - are there any particular repo steps? 

I bet it's something really simple. Thanks for your patience.

richgel999:
I've reproduced it locally! 
image

surma:
Quick intro:

To run Squoosh:
- Make sure you have Node 14+
- `npm install` 
- `npm run dev` 
- Once `waiting for changes...` appears,  a local version of Squoosh is served on `https://localhost:5000`. Wasm binaries are committed so this should right after cloning. 

To re-build the basis codec:

- Go to `codecs/basis`
- Edit `Makefile` and/or the C++ files in `enc/` and `dec/` as desired.
- Run `npm run build` inside the `codecs/basis` folder to build the new wasm binaries. This relies on Docker!
- Restart the Squoosh dev server, as it currently doesn’t pick up on wasm binary changes automatically.

If anything isn’t working, let me know.

richgel999:
BTW - another option you'll definitely want to add is whether or not the texture is sRGB, or not. Most photos/images/textures will be sRGB, but some (like normal maps) aren't. This option would then set the "m_mip_srgb" and m_perceptual flags. There are others, but maybe (just talking out loud) I could add them.

richgel999:
When I run "npm run build" in the codecs/basis directory, I get this error:
image

Any ideas? I've tried Googling with no success. You mention Docker - but I'm not familiar with it. 

I've looked at how you're compiling and calling the encoder, and it seems OK. Once I can get this building locally I should be able to quickly track it down.

richgel999:
Looks like I'll need Docker, which doesn't work under WSL v1 (without jumping through a bunch of iffy looking hoops). I'm going to switch to running Ubuntu v19 in a VM. Bear with me!

surma:
Ah sorry about that. We use Docker to make sure that our wasm builds are 100% reproducible.

For the purpose of this PR, though, you can try to compile the codec directly without Docker, provided you have Emscripten/Emsdk installed:

```
export CFLAGS="-O3 -flto -s FILESYSTEM=0"
export CXXFLAGS="${CFLAGS} -std=c++17"
export LDFLAGS="${CFLAGS} -s PTHREAD_POOL_SIZE=navigator.hardwareConcurrency"
emmake make -j`nproc`
```

richgel999:
Thanks a bunch - compiling directly seems to have worked. I have freshly written files in dec/basis_dec.js and enc/basis_enc.js.

richgel999:
Just letting you know I'm still working on this - got distracted by a lot of sun!

surma:
> got distracted by a lot of sun!

I’m 100% supportive (and jealous) of that. 🇬🇧  is not having any of that recently.

richgel999:
Ok - diving in again to track this down. Building Basis is definitely working, which is good, although each iteration takes a few minutes. (After starting the server, when I browse to it there's a delay of around 30-60 seconds for it to come up, not sure why. I get "404 - the requested path could not be found" but eventually it works.)

richgel999:
Looks like the resampler is failing to initialize in its constructor and the caller didn't check for the error (whoops!), which I'm definitely fixing. Tracking down the cause.

richgel999:
Ok, found the problem: The "g_num_resample_filters" global is 0, which is not good. It should never be 0. Looks like basisu_resample_filters.cpp wasn't compiled/linked into the project. (As a side note, with MSVC, my main development platform, this wouldn't have linked due to a missing symbol, but I guess with emscripten there are subtle differences.)

Because it's 0 the resampler can never find any filters, and it will always fail, which is a very low probability failure (which needs to be handled better!).

I'm trying to fix the makefile.

richgel999:
Including basisu_resample_filters.cpp in your makefile fixes the problem:

image

In codecs/basis/Makefile, I added encoder/basisu_resample_filters.cpp to CODEC_CPP_SOURCE_FILES like this:

```
CODEC_CPP_SOURCE_FILES := \
	encoder/basisu_comp.cpp \
	encoder/basisu_enc.cpp \
	encoder/basisu_backend.cpp \
	encoder/basisu_basis_file.cpp \
	encoder/basisu_etc.cpp \
	encoder/basisu_uastc_enc.cpp \
	encoder/basisu_gpu_texture.cpp \
	encoder/basisu_frontend.cpp \
	encoder/basisu_bc7enc.cpp \
	encoder/basisu_pvrtc1_4.cpp \
	encoder/basisu_astc_decomp.cpp \
	encoder/basisu_global_selector_palette_helpers.cpp \
	encoder/basisu_resampler.cpp \
	encoder/basisu_kernels_sse.cpp \
	encoder/basisu_resample_filters.cpp \
	encoder/jpgd.cpp \
	encoder/lodepng.cpp \
	transcoder/basisu_transcoder.cpp
```

I'll modify the code so it checks the Resample constructor's error code (so embarrassing!), so if this happens to anyone else they'll get an error vs. an assert. (I didn't realize that, at least with emscripten, if you don't include this .cpp you would still be able to link at all.)

surma:
Thanks so much for digging into this @richgel999! I am surprised Emscripten didn‘t warn for a missing symbol here. That’s usually something you have to explicitly opt-in to. Looping in @kripken as an FYI.

I’ll adjust the Makefile as you suggested! Thanks again.

surma:
Alright, I updated the PR with more mimapping options and stuff. I think we are fairly close to an MVP :D

@richgel999 I think I have one final question: Playing with the current [staging version](https://deploy-preview-1017--squoosh.netlify.app/), the “Quality” slider (mapped to `m_quality_level`) has some effect when using ETC1S. But even the highest quality is still quite blocky. Is that WAI?

Screen Shot 2021-05-27 at 13 55 26

Related: When using UASTC, the quality slider has _no_ visible effect and doesn’t affect file size at all. Is that WAI? Should we hide the quality slider when people select UASTC?

kripken:
I believe the lack of a warning on an undefined global is the LLVM issue discussed here: https://github.com/emscripten-core/emscripten/issues/14106#issuecomment-833780206

I'll bump there and try to raise the priority.

richgel999:
> @richgel999 I think I have one final question: Playing with the current staging version, the “Quality” slider (mapped to m_quality_level) has some effect when using ETC1S. But even the highest quality is still quite blocky. Is that WAI?

Yes, ETC1S has block artifacts, especially around strong chroma transitions/gradients. UASTC fixes this completely but its bitrate is huge (4-6 bits/texel vs. ~1 bits/texel for ETC1S). We've been working on a 3rd interchangeable texture format that is in between ETC1S/UASTC, which has built-in compression, which would have way less block artifacts.

Fundamentally ETC1S is based off a popular hardware-friendly format on Android with a fixed 4x4 texel block size. The format only has a single RGB 555 base color for each block, and 2-bits for each texel which are used to modulate the intensities of each texel in the block. It's a subset of full ETC1, which subdivides the blocks into 2x4 or 4x2 subblocks, but we don't utilize subblocks to make transcoding to other GPU texture formats much faster. (Basis ETC1S is kind of like hardware friendly gain shape VQ.)

In practice, game developers are very familiar with 4x4 block artifacts and they work around them in various ways. Other popular formats have them, like BC1.

richgel999:
Using a simple pixel shader, I think it's possible to smooth out the ETC1S chroma block artifacts, which is something we've been experimenting with for ETC1S video. The system supports simple ETC1S video (with or without mipmaps) via large texture arrays and skip blocks (aka Conditional Replenishment in MPEG1).

richgel999:
> Related: When using UASTC, the quality slider has no visible effect and doesn’t affect file size at all. Is that WAI? Should we hide the quality slider when people select UASTC?

Yes, UASTC is a different beast from ETC1S. It's intended for extremely high quality textures, comparable to 4x4 LDR ASTC and BC7. The encoder support RDO (rate distortion optimized) UASTC post-processing, which when combined with lossless Zstd compression in KTX2 results in a way for the user to trade off increased distortion for smaller compressed files. In the video game world this is state of the art stuff.

By default, RDO UASTC post-processing is disabled, so there is not direct quality vs. size tradeoff. (There's a quality vs. encoding time tradeoff by setting the pack level in the m_pack_uastc_flags flags parameter, but the quality delta is very small.) RDO post-processing can be enabled by setting m_rdo_uastc to true. The m_rdo_uastc_quality_scalar ("lambda" in RDO literature) controls the quality, where 0=no change and higher values up to around 4.0 or so result in progressively smaller files but more distortion. Good values to try are around .25-1.5. 

RDO post-processing in this codebase is currently pretty slow, especially without multithreading. We have a newer branch that needs to be merged that speeds it up and improves quality, which is coming soon.

surma:
Thanks for all that insight. Super helpful. 

To double-check: Am I understanding you correctly that it would make sense to make the quality slider map to `m_rdo_uastc_quality_scalar` when UASTC is selected? (Let’s ignore the encoding speed cost for now)

richgel999:
> To double-check: Am I understanding you correctly that it would make sense to make the quality slider map to m_rdo_uastc_quality_scalar when UASTC is selected? (Let’s ignore the encoding speed cost for now)

Yes, you would want to remap a quality slider from [0,1] (where 1=highest quality/RDO completely disabled) to [X,0], where X is a lambda around 3.0. Also be sure to enabled Zstd KTX2 supercompression:

params.m_ktx2_uastc_supercompression = basist::KTX2_SS_ZSTANDARD

If you set the m_rdo_uastc_quality_scalar to 0 (no RDO), be sure to just disable RDO entirely.

surma:
Alright folks! 

This is ready for review!

codecs/basis/Makefile
```diff
@@ -0,0 +1,83 @@
+CODEC_URL := https://github.com/BinomialLLC/basis_universal/archive/refs/tags/v1.15_rel2.tar.gz
+CODEC_DIR := node_modules/basis
+CODEC_BUILD_DIR := $(CODEC_DIR)/build
+CODEC_LIB := $(CODEC_BUILD_DIR)/basis.a
+ENVIRONMENT = worker
+
+OUT_JS := enc/basis_enc.js dec/basis_dec.js
+OUT_WASM := $(OUT_JS:.js=.wasm)
+
+COMMON_FLAGS := -O3 -fno-strict-aliasing
```
RReverser, line 10: We already have opt flags in `cpp.Dockerfile`, I'd remove `-O3` here.

codecs/basis/Makefile
```diff
@@ -0,0 +1,83 @@
...
+		-I $(CODEC_DIR)/transcoder \
+		${CXXFLAGS} \
+		${LDFLAGS} \
+		--closure 1 \
+		--bind \
+		-s ALLOW_MEMORY_GROWTH=1 \
+		-s MODULARIZE=1 \
+		-s TEXTDECODER=2 \
+		-s ENVIRONMENT=$(ENVIRONMENT) \
+		-s EXPORT_ES6=1 \
```
RReverser, line 62:
```suggestion
		-s ENVIRONMENT=$(ENVIRONMENT) \
		-s EXPORT_ES6=1 \
```

Some of those flags are already specified in Dockerfile, while `MODULARIZE` is auto-implied by `EXPORT_ES6`.

src/features/decoders/basis/worker/basisDecode.ts
```diff
@@ -0,0 +1,32 @@
...
+ * limitations under the License.
+ */
+import type { BasisModule } from 'codecs/basis/dec/basis_dec';
+import { initEmscriptenModule, blobToArrayBuffer } from 'features/worker-utils';
+
+let emscriptenModule: Promise<BasisModule>;
+
+export default async function decode(blob: Blob): Promise<ImageData> {
+  if (!emscriptenModule) {
+    const decoder = await import('codecs/basis/dec/basis_dec');
```
RReverser, line 20: Better to chain those promises with `.then` so that `emscriptenModule` is assigned asap instead of `await`-ing; otherwise you risk that another call will come in, see that `emscriptenModule` is still not assigned and instantiate a separate copy.

This probably doesn't happen today, but better to be safe.

RReverser: > Better to chain those promises with .then

or use a separate `init` function like you do in other files

RReverser:
Left few nits, otherwise looks good.

src/features/encoders/basis/client/index.tsx
```diff
@@ -0,0 +1,214 @@
...
+    ) as HTMLFormElement;
+    const { options } = this.props;
+
+    const uastc = form.mode.value === '1';
+    let quality = inputFieldValueAsNumber(form.quality, options.quality);
+    if (uastc) {
+      quality = clamp(0, quality, 4);
+    } else {
+      quality = Math.floor(clamp(0, quality, 255));
+    }
```
jakearchibald, line 53: Are these different things to the user? I wonder if we should give them seperate state here.

https://github.com/GoogleChromeLabs/squoosh/blob/dev/src/features/encoders/avif/client/index.tsx - I think I developed a better pattern for the AVIF UI, where the incoming settings are copied to state and translated back to encoders settings in the change event.

In this case it means these two qualities have can different defaults.

src/features/encoders/basis/client/index.tsx
```diff
@@ -0,0 +1,214 @@
...
+                {options.mipmap ? (
+                  <Fragment>
+                    <div class={style.optionOneCell}>
+                      <Range
+                        min="0"
+                        max="10"
+                        name="mipmap_min_dimension"
+                        value={Math.floor(
+                          Math.log2(options.mipmap_min_dimension),
+                        )}
```
jakearchibald, line 162: The suggestion above (storing the state locally) also means this kind of translation wouldn't need to be done per render.

src/features/encoders/basis/client/index.tsx
```diff
@@ -0,0 +1,214 @@
...
+        </label>
+        <label class={style.optionReveal}>
+          <Revealer
+            checked={showAdvanced}
+            onChange={linkState(this, 'showAdvanced')}
+          />
+          Advanced settings
+        </label>
+        <Expander>
+          {showAdvanced ? (
```
jakearchibald, line 134: fwiw, this can be `{showAdvanced && (` and then doesn't need the `: null}` at the end. It's a bit of a 'trick' but it seems common in JSX.

jakearchibald (changes requested):
Some UI observations:

- Not sure what 'compression' does. Is this 'effort'? If so, let's name it that so it's consistent with other codecs.
- 'compression' doesn't seem to do anything in 'quality' mode.
- The 'quality' slider seems backwards in 'quality' mode. Lower quality values have a bigger size and higher visual quality.
- 'perceptual distance metric' doesn't seem to do anything in 'quality' mode.

jakearchibald:
Would it be useful to get feedback from someone who uses this codec? Ada maybe?

surma:
> Would it be useful to get feedback from someone who uses this codec? Ada maybe?

Good shout. I’ll see if I can summon a few folks to give this a quick test before I address your other remarks. Here we go: 

@toji @mrdoob @AdaRoseCannon would you mind taking a look at the [staging version of Squoosh](https://deploy-preview-1017--squoosh.netlify.app/) to see if it is useful for your usual work with WebGL/WebXR/ThreeJS? Feel free to loop in other folks who might have opinions!

toji:
Overall this is really great!

I can't comment on the sliders, as I'm not super familiar with what effect they are supposed to have. The "Compression (ETC1)"/"Quality (UASTC)" drop down could be slightly confusing, because "Quality" is still a form of GPU compression and even at its highest settings will probably be a perceptual downgrade from JPEG. I would say that labeling them "Higher Compression (ETC1)" and "Higher Quality (UASTC)" respectively would be more accurate.

FWIW: The general strategy with GPU compression is to accept the quality loss and use the VRAM it frees up to use either larger version of the same texture or more textures in the same scene, either one being an overall win on visual quality. If you aren't aware of this then it's easy to be disappointed by the compressed quality when viewing the sort of 1:1 comparisons Squoosh visualizes so well. I don't really know how to compensate for that?

For the types of usage I would typical aim for (Realtime 3D scenes), having "Embed Mipmaps" enabled by default would almost always be the right answer. Also, I feel like defaulting the mipmaps to use sRGB Mipmapping is probably most frequently the right thing to do, but ultimately that depends on if the image is meant as a color texture or something like a normal map. It's really going to depend on who's using it. 🤷 Also to be honest, someone with my target use case would probably be more drawn to using the squoosh page to help figure out the right commands to send to the CLI so I can automatically run it over my entire project. (I noticed the page says that Basis isn't supported by the CLI yet?)

mrdoob:
I think @donmccurdy may have some good feedback too.

donmccurdy:
### Input/output dimensions

**(1)** When the compressed texture does not have power-of-two dimensions, it will (in most cases) not work as users expect with WebGL 1.0. **(2)** When the compressed texture does not have multiple-of-four dimensions, it may not work at all. For these reasons it might be best to show some kind of warning or prompt on non-POT textures.  In [my own CLI implementation](https://github.com/donmccurdy/glTF-Transform/blob/02a33e65bdd314d0eb72ec5d19b7c73b2e0da9ca/packages/cli/src/transforms/toktx.ts#L268-L282) I enforce multiple-of-four dimensions, and provide a warning and an option for non-POT dimensions.

### Compression option

> Not sure what 'compression' does. Is this 'effort'?

To my understanding this is a misnomer, it's an encoding speed vs. visual quality tradeoff, and has little effect on compressed size. The "quality" slider is definitely the place most users should start tuning. In the [guide Khronos provided to artists](https://github.com/KhronosGroup/3D-Formats-Guidelines/blob/main/KTXArtistGuide.md#compressing-to-ktx-with-gltf-transform), the compression option does not even make an appearance for ETC1S mode; for UASTC mode it is named `--level` and can be more helpful. I might consider moving it to the advanced section, or hiding it entirely for ETC1S, for these reasons.

### Color / non-color data

> ... ultimately that depends on if the image is meant as a color texture or something like a normal map.

This is tricky. :/ ... there are a handful of settings that would ideally be used for "non-color" textures like normal maps or O/R/M data... I'm not sure how to accommodate these within the UI, they're easy enough to infer when processing a whole glTF file with referenced textures, based on context, but harder — without some pretty niche knowledge — to get right on isolated textures. It's not so easy to discover the `Linear RGB` checkbox under the "Resize" expander either... Blender provides a "Color Space" dropdown with "sRGB" and "Non-Color" as options, artists may know those.

Here are those the settings that relate to non-color data as they would be named in the KTX-Software CLI...

https://github.com/donmccurdy/glTF-Transform/blob/02a33e65bdd314d0eb72ec5d19b7c73b2e0da9ca/packages/cli/src/transforms/toktx.ts#L251-L266

... I'm not exactly sure how these translate into Basis Universal options, perhaps @richgel999 can help with that. I've already written a CLI utility (`gltf-transform ktxfix in.glb out.glb`) that scans a glTF file and fixes incorrect KTX2 color primaries, inferring primaries from their context in the glTF file, but this may be harder for developers using the textures standalone in a WebGL application. The R/RG channel information enables clients to choose a more optimal transcoding target.

@lexaknyazev any idea how we might infer these settings, or what the practical implications are for getting them wrong (e.g. on a normal map)?

donmccurdy:
Aside – is this something I can import and use as a JS/WASM dependency? I've been hoping to support processing whole glTF/GLB files in the browser for KTX2/Basis compression, but currently just have a CLI implementation based on KTX-Software. Or should I just start with the encoder in the Basis Universal repository?

grandpa1946:
why no merge?
this repo has 25 branches made by dependabot, maybe give it merge permission?

PR #1018, opened 2021-05-18
Add a call-out to Web Codecs API during decoding

surma:
Added a code path that uses [Web Codecs API](https://w3c.github.io/webcodecs) during decoding.


src/client/lazy-app/Compress/index.tsx
```diff
@@ -109,8 +110,12 @@ async function decodeImage(
       if (mimeType === 'image/webp2') {
         return await workerBridge.wp2Decode(signal, blob);
       }
-      // If it's not one of those types, fall through and try built-in decoding for a laugh.
     }
+    // If none of those work, let’s see if Web Codecs API is available
+    if (await WebCodecs.isTypeSupported(mimeType)) {
+      return await abortable(signal, WebCodecs.decode(blob, mimeType));
```
surma: I am not returning to the hail-mary path below because Web Codecs API will support all the image formats that built-in canvas supports.

dalecurtis: Just a slight caveat there, we don't support ICO/CUR and SVG presently. SVG requires DOM, so can't be done in a worker context. ICO/CUR function oddly so we've disabled support in the mean time. Either way the isTypeSupported() will give you a valid answer.

src/client/lazy-app/util/web-codecs/index.ts
```diff
@@ -0,0 +1,26 @@
...
+): Promise<ImageData> {
+  if (!hasImageDecoder) {
+    throw Error(
+      `This browser does not support ImageDecoder. This function should not have been called.`,
+    );
+  }
+  const decoder = new ImageDecoder({
+    type: mimeType,
+    // Non-obvious way to turn an Blob into a ReadableStream
+    data: new Response(blob).body!,
```
jakearchibald, line 22: Fwiw, `blob.stream()` has decent support now. But maybe we should wait a couple more Safari versions before using it. https://caniuse.com/?search=blob%20stream

jakearchibald (changes requested):
It feels too early to land this since it's an evolving API (the demos in https://twitter.com/ChromiumDev/status/1255870582442405889 no longer work) and there isn't a spec yet.

Should we make this a draft?

surma:
Oh I thought this was stable-ish. Yeah, happy to let this sit for a while then.

dalecurtis:
ImageDecoder is now stable (updates were tracked here: https://bugs.chromium.org/p/chromium/issues/detail?id=1182435). I've ping @tomayac about updating the demo. The spec is here https://w3c.github.io/webcodecs/#image-decoding

tomayac:
> ImageDecoder is now stable (updates were tracked here: https://bugs.chromium.org/p/chromium/issues/detail?id=1182435). I've ping @tomayac about updating the demo. The spec is here https://w3c.github.io/webcodecs/#image-decoding

✅ Demo updated: https://imagedecoder.glitch.me/.

jakearchibald:
@dalecurtis my bad. Maybe update https://github.com/dalecurtis/image-decoder-api? That page has good SEO 😀 and there's no link to the spec.

dalecurtis:
Ah! I thought I had already done that, thanks for calling that out. Done.

src/client/lazy-app/Compress/index.tsx
jakearchibald: Are there any other differences we should be aware of vs `createImageBitmap`?

jakearchibald: @surma shouldn't this code be rolled into `builtinDecode`? Since it's using the built-in decoders

jakearchibald:
I've got a vague worry that we're moving from an API where the mimetype doesn't need to be accurate, to one where it does. Like, in Squoosh right now we could remove a bunch of our mime sniffing code, since we only need to sniff for the encoders we ship.

However, with this PR, we can't do that, as Chrome would no longer be able to load a `.png` that was actually a JPEG.

Maybe it's fine. I'm just worried about adding the complexity for little/no gain.

In future this would let us support animated formats though, which is nice.

surma:
Would you prefer if we continued to fall through to the hail-mary canvas decoding when WebCodecs fails?

jakearchibald:
Yeah let's do that for safety, just in case it turns out our WebP sniffing is wrong for some exotic flavour of WebP or whatever.

We should keep our current set of file sniffing, as we'd need it if we wanted to support animated types.

jakearchibald:
Filed https://github.com/w3c/webcodecs/issues/252

surma:
Done!

jakearchibald:
I still think this code should go in `builtinDecode`, since that's what it is

surma:
Done. I added a couple of aborted checks to avoid unnecessary work. Is that the idiomatic way to do that with `AbortController`?

src/client/lazy-app/util/index.ts
```diff
@@ -216,13 +237,27 @@ export function drawableToImageData(
...
+  signal: AbortSignal,
+  blob: Blob,
+  mimeType: string,
+): Promise<ImageData> {
+  // If WebCodecs are supported, use that.
+  if (await WebCodecs.isTypeSupported(mimeType)) {
+    if (signal.aborted) return Promise.reject('Aborted');
+    try {
+      return await WebCodecs.decode(blob, mimeType);
+    } catch (e) {}
```
jakearchibald:
```suggestion
    assertSignal(signal);
    try {
      return await abortable(signal, WebCodecs.decode(blob, mimeType));
    } catch (e) {}
```

src/client/lazy-app/util/index.ts
```diff
@@ -216,13 +237,27 @@ export function drawableToImageData(
...
+  }
+  if (signal.aborted) return Promise.reject('Aborted');
+
   // Prefer createImageBitmap as it's the off-thread option for Firefox.
   const drawable =
     'createImageBitmap' in self
       ? await createImageBitmap(blob)
       : await blobToImg(blob);
 
+  if (signal.aborted) return Promise.reject('Aborted');
```
jakearchibald:
```suggestion
  assertSignal(signal);
  
  // Prefer createImageBitmap as it's the off-thread option for Firefox.
  const drawable = await abortable(
    signal,
    'createImageBitmap' in self ? createImageBitmap(blob) : blobToImg(blob),
  );
```

jakearchibald (changes requested):
We've got a nice little `abortable` helper 😄 

(I haven't tested these changes)

src/client/lazy-app/util/index.ts
surma: Right, but I wanted to avoid starting `WebCodecs.decode()` to run if we already got an abort signal.

src/client/lazy-app/util/index.ts
surma: I applied the same thinking here: Couldn’t there be a scenario where `createImageBitmap` or `blobToImg` start doing work even though we already got an abort signal?

src/client/lazy-app/util/index.ts
jakearchibald: Updated

src/client/lazy-app/util/index.ts
jakearchibald: Also updated

jakearchibald (approved):
LGTM!

jakearchibald:
I've filed https://github.com/GoogleChromeLabs/squoosh/issues/1028 for me to do something better for abortables

dalecurtis:
@jakearchibald WDYT about having AbortSignal as a second parameter to decode()? (In the standard) Or as a member in the ImageDecodeOptions dictionary.

surma:
I know you didn’t ask me, but considering that even decoding can take any amount of time in the context of WebCodecs (as we don’t curate the codecs anymore), that seems like a very good idea to me.

dalecurtis:
All opinions welcome :) Filed https://github.com/w3c/webcodecs/issues/254

PR #1019, opened 2021-05-20
[CLI] Handle subdirectories without failing

atjn:
This is a rebase of #991 - I apparently still have some things to learn when it comes to git commands 😅

When you ask the CLI to minify an entire directory - something like this:
```
npx @squoosh/cli --webp "auto" /path/to/folder/
```
...and the directory contains subdirectories, squoosh will try to read the directories as files, and will fail spectacularly.

In the future, you might want to let the user choose to read the subdirectories recursively or something, but for the time being, this PR fixes the problem by making squoosh ignore any subdirectories.

### DIFF
--- a/cli/src/index.js
+++ b/cli/src/index.js
@@ -75,7 +75,7 @@ async function getInputFiles(paths) {
 
   for (const inputPath of paths) {
     const files = (await fsp.lstat(inputPath)).isDirectory()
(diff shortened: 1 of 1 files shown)


surma (approved):
Thanks for this. I plan to switch to globs here at some point, but this is a good fix for now. Thank you!

PR #1023, opened 2021-05-21
<title> to indicate progress/error

hchiam:
Attempt at #644. Maybe there's a cleaner way to use the preact state, but this looks like it's working so far.


hchiam (commented):
I basically looked for `loading: true` and `loading: false`, decided where it likely made sense to call `updateDocumentTitle` with indicators, tested different controls, and hard-coded some errors to throw to see how it displayed.

jakearchibald:
Thanks for taking a look at this!

I'm not sure this is right, as it's clearing loading state when only one side completes, even though the other side may still be in progress.

I think it's best to calculate this in `componentDidUpdate`. That gives you access to the previous state in `prevState`, and the current state in `this.state`.

Use this to figure out if the state 'was loading' and 'is loading'. Remember there are 3 places with loading state. `state.loading`, `state.sides[0].loading`, and `state.sides[1].loading`. If 'was loading' and 'is loading' are different, you know you need to update the document title.

I wouldn't worry about error state, loading is enough.

src/client/lazy-app/Compress/index.tsx
```diff
@@ -374,6 +376,15 @@ export default class Compress extends Component<Props, State> {
   }
 
   componentDidUpdate(prevProps: Props, prevState: State): void {
+    const isLoading =
+      this.state.loading ||
+      this.state.sides[0].loading ||
+      this.state.sides[1].loading;
+    if (isLoading) {
```
hchiam: alternatively, i could do this, but it seemed unnecessary:
```suggestion
    const wasLoading = prevState.loading;
    const isLoading = this.state.loading || this.state.sides[0].loading || this.state.sides[1].loading;
    const changedLoadingState = wasLoading !== isLoading;
    if (changedLoadingState) {
```
(both ways seemed to work btw, but let me know which is more correct)

jakearchibald: This seems better to me, as it avoids going though the title setter if the title doesn't need to change.

hchiam:
@jakearchibald sorry for the multiple reviews, it should be ready now - I used the suggestion and did a sanity check with a console log to make sure `changedLoadingState` was correctly toggling to `true` and back to `false`

src/client/lazy-app/Compress/index.tsx
```diff
@@ -374,6 +376,14 @@ export default class Compress extends Component<Props, State> {
   }
 
   componentDidUpdate(prevProps: Props, prevState: State): void {
+    const wasLoading = prevState.loading;
+    const isLoading = this.state.loading || this.state.sides[0].loading || this.state.sides[1].loading;
```
jakearchibald: `this.state` and `prevState` are the same type, so you need to check all three loading points on `prevState` too.

src/client/lazy-app/Compress/index.tsx
```diff
@@ -374,6 +376,14 @@ export default class Compress extends Component<Props, State> {
...
+    const changedLoadingState = wasLoading !== isLoading;
+    if (changedLoadingState) {
+      updateDocumentTitle(loadingIndicator + this.state.source?.file.name);
```
jakearchibald: Just because the loading state has changed doesn't mean we're now in the loading state. It could have changed from loading to not-loading.

jakearchibald (changes requested):
Getting close!

src/client/lazy-app/Compress/index.tsx
hchiam: Ah ok it looked like it was working but that'll make the code make more sense, will fix.

hchiam:
Ok, ready for another review - thanks for guiding me thru this btw

src/client/lazy-app/Compress/index.tsx
```diff
@@ -374,6 +376,21 @@ export default class Compress extends Component<Props, State> {
...
+      this.state.loading ||
+      this.state.sides[0].loading ||
+      this.state.sides[1].loading;
+    const startedLoading = !wasLoading && isLoading;
+    const stoppedLoading = wasLoading && !isLoading;
+    if (startedLoading) {
+      updateDocumentTitle(loadingIndicator + this.state.source?.file.name);
+    } else if (stoppedLoading) {
+      updateDocumentTitle(this.state.source?.file.name);
+    }
```
jakearchibald:
```suggestion
    if (wasLoading !== isLoading) {
      if (isLoading) {
        updateDocumentTitle(loadingIndicator + this.state.source?.file.name);
      } else {
        updateDocumentTitle(this.state.source?.file.name);
      }
    }
```

Once you know the loading state has changed, `isLoading` gives you everything you need.

jakearchibald:
This is working, except for the initial loading state (when you first select an image).

I think this is because the `updateDocumentTitle` call in `updateImage` is removing the loading indicator.

I'm open to better suggestions, but I think `updateDocumentTitle(filename)` should become something like `updateDocumentTitle({ loading, filename })`, and we should only ever call this in `componentDidUpdate`.

Right now we're checking if 'loading' has changed before calling `updateDocumentTitle`, but we should also check if `state.source` has changed too.

@hchiam are you ok to keep going with this? Totally understand if you're sick of the churn here

src/client/lazy-app/Compress/index.tsx
```diff
@@ -374,6 +376,22 @@ export default class Compress extends Component<Props, State> {
...
+    const wasLoading =
+      prevState.loading ||
+      prevState.sides[0].loading ||
+      prevState.sides[1].loading;
+    const isLoading =
+      this.state.loading ||
+      this.state.sides[0].loading ||
+      this.state.sides[1].loading;
+    const sourceChanged = prevState.source !== this.state.source;
+    if (wasLoading !== isLoading || sourceChanged) {
```
hchiam, line 396: or: 
```suggestion
    const sourceChanged = prevState.source !== this.state.source;
    const leftImageChanged = prevState.sides[0].file !== this.state.sides[0].file;
    const rightImageChanged = prevState.sides[1].file !== this.state.sides[1].file;
    const imageChanged = sourceChanged || leftImageChanged || rightImageChanged;
    if ((wasLoading !== isLoading) || imageChanged) {
```

hchiam:
I'll think about `updateDocumentTitle({ loading, filename })` and try one more thing.

src/client/lazy-app/Compress/index.tsx
jakearchibald: Nah, what you've got here is good. When the sides change, they don't change filename.

src/client/lazy-app/Compress/index.tsx
```diff
@@ -257,12 +262,16 @@ function processorStateEquivalent(a: ProcessorState, b: ProcessorState) {
...
-function updateDocumentTitle(filename: string = ''): void {
-  document.title = filename
-    ? `${filename} - ${originalDocumentTitle}`
-    : originalDocumentTitle;
+function updateDocumentTitle(loadingFileInfo: LoadingFileInfo): void {
+  const { loading, filename } = loadingFileInfo;
+  document.title =
+    filename && loading
+      ? `${filename} - ${originalDocumentTitle}`
+      : originalDocumentTitle;
```
jakearchibald: Since this function is formatting the document title, it should take care of the loading indicator. It shouldn't be part of the filename.

```suggestion
  let title = '';
  if (loading) title += loadingIndicator;
  if (filename) title += filename + ' - ';
  title += originalDocumentTitle;
  document.title = title;
```

src/client/lazy-app/Compress/index.tsx
```diff
@@ -366,14 +375,30 @@ export default class Compress extends Component<Props, State> {
...
+      this.state.loading ||
+      this.state.sides[0].loading ||
+      this.state.sides[1].loading;
+    const sourceChanged = prevState.source !== this.state.source;
+    if (wasLoading !== isLoading || sourceChanged) {
+      updateDocumentTitle({
+        loading: isLoading,
+        filename:
+          (isLoading ? loadingIndicator : '') + this.state.source?.file.name,
+      });
```
jakearchibald: As above, the loading indicator isn't part of the file name.

```suggestion
      updateDocumentTitle({
        loading: isLoading,
        filename: this.state.source?.file.name,
      });
```

jakearchibald (changes requested):
Nearly there I think…

jakearchibald (approved):
Looks good to me!

surma:
Thank you @hchiam! I really appreciate you seeing this one through to the end 🎉

jakearchibald:
Let's land and ship this! Thanks @hchiam!

jakearchibald:
And it's live!

PR #1025, opened 2021-05-24
Remove SIMD for now

RReverser:
This is by no means a proper fix, but Chrome 91 (stable tomorrow) brings some breaking changes to SIMD, and I'd rather disable SIMD for now to avoid breaking Squoosh altogether once it hits stable, and work on a proper fix in a separate branch.


jakearchibald (approved):
LGTM. Filed https://github.com/GoogleChromeLabs/squoosh/issues/1026

PR #1027, opened 2021-05-24
Bundle URLs as assets

RReverser:
- Upgrade Emscripten to 2.0.21. Few notes:
	   - Lots of deprecated SIMD intrinsic warnings & errors in JPEG-XL -> Highway; had to suppress erorrs to make project build. https://github.com/google/highway/issues/206
	   - Moved couple of common link flags to cpp.Dockerfile (note: can't move `EXPORT_ES6` otherwise `configure` will fail).
	   - MODULARIZE=1 is no longer necessary and implied by EXPORT_ES6.
	   - EXPORT_NAME=... is no longer necessary in EXPORT_ES6.
	   - Changed visdif to also use EXPORT_ES6 and ENVIRONMENT=node instead of generic JS.
 - Update wasm-bindgen.
 - Bundle URL assets as dependencies

	  I've added support for `new URL(..., import.meta.url)` pattern accepted by many bundlers to wasm-bindgen and Emscripten, both for Wasm files as well as for side Workers autogenerated to support multithreading.
	  
	  On the apps side like Squoosh this means we no longer have to manually specify URLs to be passed to wasm-bindgen or Emscripten init functions. In this PR I'm primarily doing just that - removing all the manual Wasm URL locators, as well as the associated logic.
	  
	  One notable change on the build system side is that, in order for Wasm and multithreading-related Workers to be detected as dependencies, I'm now walking over `referencedFiles` in addition to `imports` when collecting list of dependencies in `getDependencies` function (client-bundle-plugin.js).
	  
	  As a side effect, it also included other linked assets - images and CSS - which simplified quite a few caching lists in to-cache.ts as well. Instead of having to manually specify each asset that needs to be included, the logic is inverted and only few extra assets had to be excluded to keep the list of in items cached on the first load as low as it was before the change.
	  
	  All in all, I think this simplification is worth it and even helped to accidentally uncover & fix one caching bug where WP2 regular JS was cached in addition to WP2 multithreaded JS simultaneously.


RReverser:
I'll recheck the Wasm sizes, those look wrong.

rollup.config.js
```diff
@@ -106,6 +107,7 @@ export default async function ({ watch }) {
           plugins: [
             { resolveFileUrl },
             OMT({ loader: await omtLoaderPromise }),
+            importMetaAssets(),
```
surma, line 110: Gonna flag this one for discussion. This automatically turns any `new URL("some-file.png", import.meta.url)` into an asset. That means that Rust and Emscripten glue code refer to the correct `.wasm` out of the box and we don’t need to pass it around manually. I’m not sure I like that pattern, as it’s not in line with how the platform handles this without. But I’m gonna yield to @jakearchibald here, it’s his area more than mine.

(As in, if it was up to me, I’d prefer we continue to explicitly import assets with `url:` prefix.)

surma:
Okay so I am all in favor of upgrading to Emscripten 2.0.21.

As for the `URL("some-file.png", import.meta.url)` pattern to signal an asset, I am not sure. I really dislike that it diverges from the semantics of the platform here. But Imma yield to @jakearchibald here because it’s his area.

(As in, if it was up to me, I’d prefer we continue to explicitly import assets with `url:` prefix.)

RReverser:
> I really dislike that it diverges from the semantics of the platform here

How so? It aligns with the platform in that the `new URL` is valid both before and after the bundling; the job of bundler is just to keep code working, and this is a pattern they currently agreed on as statically analysable.

RReverser:
> I'll recheck the Wasm sizes, those look wrong.

I've rechecked, those reproduce even with just Emscripten upgrade with no other changes.

I can only presume that the `-O3` which we use for speed reasons (it gave significant boost to codecs like AVIF in the past) got "better" at inling or unrolling loops or something else which now causes it to produce a much bigger code - given that there was a major LLVM bump in-between, this seems likely but cc @kripken in case 20-30% size increase is suspicious.

surma:
> the job of bundler is just to keep code working

That’s an interesting interpretation. I can get behind that. I was worried about accidentally triggering asset creation where I didn’t mean to, but that still doesn’t affect runtime performance really...

RReverser:
> I was worried about accidentally triggering asset creation where I didn’t mean to

Ah yeah, I think it's pretty hard to do that by accident given how explicit the pattern is. I'd agree if it was crawling all the strings that look like URLs, but `new URL('...', import.meta.url)` is almost like static syntax in terms of explicitness.

I hope `import.meta.resolve` proposal comes through and pushes it even further in future though.

kripken:
> those reproduce even with just Emscripten upgrade with no other changes.
> 
> I can only presume that the `-O3` which we use for speed reasons (it gave significant boost to codecs like AVIF in the past) got "better" at inling or unrolling loops or something else which now causes it to produce a much bigger code - given that there was a major LLVM bump in-between, this seems likely but cc @kripken in case 20-30% size increase is suspicious.

In 2.0.17 LLVM switches to the new pass manager, which might explain this. Can maybe do a quick [bisect](https://emscripten.org/docs/contributing/developers_guide.html#bisecting) to verify.

RReverser:
> In 2.0.17 LLVM switches to the new pass manager, which might explain this.

Do you think the explanation would be "it's a new optimisation" or "it's a regression bug / side-effect of the new pass manager"?

I'll try before/after 2.0.17 to verify.

kripken:
Could be either, really... new pipeline of optimizations, with new heuristics. Sometimes better, sometimes worse.

RReverser:
I just checked the speed of AVIF with standard params, and looks like it barely changed (~7-8% improvement but could be just noise since those are one-off runs). Maybe some more niche code paths did get better, idk.

@kripken I guess what I meant to ask is, whether this sort of size increase (20-30% seems pretty significant) is considered a regression or a problem at all by LLVM, or whether size increases at O3 are ignored as a natural artifact of optimisations and aren't tracked closely. Or, to rephrase, is this something worth of filing a bug?

kripken:
I don't work enough with LLVM and the LLVM community to be sure of how they are treating regressions. We saw a few (smaller) regressions on the emscripten benchmark suite, but decided not to file issues as the results were mixed (some things did get faster, and some got smaller). But 20-30% larger with no benefit might be worth a bug - I'd file one, worst case it won't get much attention. If you file one, a bloaty diff might be interesting to help the discussion there.

RReverser:
Sounds reasonable. Just to double-check - by "a bloaty diff" do you mean https://github.com/google/bloaty?

kripken:
Yes.

RReverser:
> In 2.0.17 LLVM switches to the new pass manager, which might explain this.

Ha, that doesn't seem to be it. Looks like 2.0.16 produced even slightly larger binaries. Continuing...

RReverser:
Okay, finished bisect. Looks like there was a bunch of size changes over the versions.

Over the whole 2.0.12-2.0.21 range increased `avif_enc.wasm` increased by 122KB total. However, a single 2.0.11-2.0.12 bump contributed 311KB and looks most suspicious.

RReverser:
Hmm, 2.0.12 Changelog doesn't look like much, but I wonder if it's the CMake change that caused this? By the sound of it, it started passing `-O3` to the CMake where it would previously pass `-O2`?

Although our `CFLAGS` have `-O3` in both versions, and I assume it would take precedence over the Emscripten default?

kripken:
Yes, if you set it, then the default should be ignored. So maybe it's not that.

Can you bisect into the release? The link I gave before shows how. It's pretty easy I think, basically just do `emsdk install HASH` a few times as you bisect over a side repo. That would give us the specific roll that regressed here. It's a super-useful technique that we are using in more and more bug reports.

https://emscripten.org/docs/contributing/developers_guide.html#bisecting

RReverser:
I think I saw Sam mentioning somewhere that emsdk install doesn't work in Docker? I guess I could try to build outside by copying the flags, just want to make sure I don't introduce impurities into comparisons.

kripken:
I'm not aware of such an issue, but I do not know the details of how the emscripten docker image works, so it could be.

RReverser:
I don't think I can spend time going into bisecting individual Emscripten commits, so raised an issue upstream instead now that it's confirmed that the size increase is coming somewhere from Emscripten itself: https://github.com/emscripten-core/emscripten/issues/14282

lib/client-bundle-plugin.js
```diff
@@ -20,18 +20,31 @@ const allSrcPlaceholder = 'CLIENT_BUNDLE_PLUGIN_ALL_SRC';
...
 
+  // Merge referencedFiles as regular deps.
+  //
+  // Didn't do this as part of the main loop since their `chunk` can't have
+  // nested deps and sometimes might be missing altogether, depending on type.
+  for (const dep of referencedFiles) {
+    crawlDependencies.add(dep);
+  }
+
   return [...crawlDependencies];
```
jakearchibald: Seems a bit weird to add referenced files to `crawlDependencies`. Does this work?

```suggestion
  return [...crawlDependencies, ...referencedFiles];
```

lib/client-bundle-plugin.js
```diff
@@ -20,18 +20,31 @@ const allSrcPlaceholder = 'CLIENT_BUNDLE_PLUGIN_ALL_SRC';
...
   for (const fileName of crawlDependencies) {
     const chunk = clientOutput.find((v) => v.fileName === fileName);
 
     for (const dep of chunk.imports) {
       crawlDependencies.add(dep);
     }
+
+    for (const dep of chunk.referencedFiles) {
+      referencedFiles.add(dep);
+    }
```
jakearchibald, line 34:
```suggestion
    // These don't go into crawlDependencies as their `chunk` can't have
    // nested deps and sometimes might be missing altogether, depending on type.
    for (const dep of chunk.referencedFiles) {
      referencedFiles.add(dep);
    }
```

lib/entry-data-plugin.js
```diff
@@ -17,10 +17,14 @@ const prefix = 'entry-data:';
...
-const filenamePrefix = 'static/';
+
+/** @param {string} fileName */
+export function fileNameToURL(fileName) {
+  return fileName.replace(/^static\//, '/');
+}
 
 export default function entryDataPlugin() {
-  /** @type {string} */
+  /** @type {number} */
```
jakearchibald, line 27: lol, thanks

lib/entry-data-plugin.js
```diff
@@ -69,15 +73,11 @@ export default function entryDataPlugin() {
...
-                chunk.fileName.slice(filenamePrefix.length),
-              );
+              return JSON.stringify(fileNameToURL(chunk.fileName));
             }
 
             return JSON.stringify(
-              getDependencies(chunks, chunk).map((item) =>
-                item.slice(filenamePrefix.length),
-              ),
+              getDependencies(chunks, chunk).map(fileNameToURL),
```
jakearchibald:
```suggestion
              getDependencies(chunks, chunk).map((filename) => fileNameToURL(filename)),
```

https://jakearchibald.com/2021/function-callback-risks/ 😀

rollup.config.js
jakearchibald, line 122: Is the intention that we should be replacing our current method of bundling assets with this method?

If so, this should go in the `commonPlugins` so it works for our static build, not just client JS. Right now it'll produce a `file:` URL in the static build.

TypeScript doesn't seem to like references to `import.meta.url` right now.

src/sw/to-cache.ts
```diff
@@ -13,92 +13,83 @@ function subtractSets<T extends any>(set1: Set<T>, set2: Set<T>): Set<T> {
...
 // WP2
-import * as wp2EncMtWorker from 'entry-data:codecs/wp2/enc/wp2_enc_mt.worker.js';
+import * as wp2EncMtSimd from 'entry-data:codecs/wp2/enc/wp2_enc_mt_simd';
 import * as wp2EncMt from 'entry-data:codecs/wp2/enc/wp2_enc_mt';
-import wp2EncMtWasm from 'url:codecs/wp2/enc/wp2_enc_mt.wasm';
 import * as wp2Enc from 'entry-data:codecs/wp2/enc/wp2_enc';
-import wp2EncWasm from 'url:codecs/wp2/enc/wp2_enc.wasm';
+
+export function shouldCacheDynamically(url: string) {
+  return url.startsWith('/c/demo-');
```
jakearchibald, line 52: I'm a bit worried about this. Our old build system behaved like this and it caught us out - things started caching differently after a file rename, and that wasn't expected. I deliberately changed it to a system where you'd get a failure if you renamed a file but didn't also change this file, rather than a behaviour change creeping in.

jakearchibald (changes requested):
Mostly blocking on the increase in wasm sizes.

This seems ok. I'm still unsure about `new URL(…, import.meta.url)` meaning "add an asset to the bundle, and hash it", and I wonder if it'll trip us up later if we just want to create a module-relative URL.

There's some fragility introduced into the caching logic **edit** no there isn't

lib/client-bundle-plugin.js
RReverser: Not really, because some JS can be simultaneously a regular entry/dependency and a referenced file (like auto-generated Wasm workers). I could create another intermediate `Set` if you want though.

lib/entry-data-plugin.js
RReverser: This function is specifically designed to be a callback though :P

src/sw/to-cache.ts
RReverser: There was already this `.startsWith('/c/demo-')` in the code, I only extracted it into a function to avoid hardcoding in two places instead of one.

rollup.config.js
RReverser: Hmm what is "static build" / what is the scenario where it wouldn't work? I've included it in the same place as `OMT` because conceptually both handle similar code paths (OMT handles `new Worker(new URL(..., import.meta.url))` already, and this plugin handles the rest of `new URL(..., import.meta.url)` not covered by OMT).

src/sw/to-cache.ts
jakearchibald: Hah, then it was I that did a bad job. I should fix that at some point, but no worried for now.

lib/client-bundle-plugin.js
RReverser: Is this meant instead of my comment or in addition to it?

lib/entry-data-plugin.js
jakearchibald: It doesn't look like it. It's called in other places not-as-a-callback.

lib/client-bundle-plugin.js
jakearchibald: Sticking with the current code but adding a comment making the above clear seems fine

src/sw/to-cache.ts
RReverser: Check src/sw/index.js change: https://github.com/GoogleChromeLabs/squoosh/pull/1027/files#diff-46250d1a173a5036cad883e5b6161d0e996c5a669cccc797034e5e776dbd4525

rollup.config.js
jakearchibald: `src/shared/prerendered-app/Intro/index.tsx` for example. This file is used to generate our initial HTML. If I change:

```js
import deviceScreenIcon from 'url:./imgs/demos/icon-demo-device-screen.jpg';
```

to

```js
const deviceScreenIcon = new URL('./imgs/demos/icon-demo-device-screen.jpg', import.meta.url).href;
```

(and work around the typescript issues) it works fine on the client, but the initial render contains a `file:` URL.

lib/client-bundle-plugin.js
jakearchibald: It was linked to the other suggested change above. Since we're sticking with the original, this can be ignored too.

rollup.config.js
RReverser: Ah, interesting. How come we didn't run into that with OMT? E.g. I already rely on it bundling the `new Worker(...)` from wasm-bindgen-rayon, I guess that one also results in `file:` URL on `dev` branch right now.

lib/entry-data-plugin.js
RReverser: Hm. I mean, I specifically wrote it in this form so I could pass it to `.map` (rather than using e.g. object as a param). How do you distinguish between those formats / what should I do to make it more explicit it's intended to be used that way?

rollup.config.js
jakearchibald: `new Worker` will only appear in JS aimed at the client

lib/entry-data-plugin.js
jakearchibald: `fileNameToURLMapper`? Or just use an inline function like above to remove the risk.

rollup.config.js
RReverser: Ok, sorry, I'm very confused again. What is "client" in this scenario? Browser? If so, what else can JS be aimed at?

lib/client-bundle-plugin.js
RReverser:
```suggestion
  // Merge referencedFiles as regular deps. They need to be in the same Set as
  // some JS files might appear in both lists and need to be deduped too.
  //
  // Didn't do this as part of the main loop since their `chunk` can't have
  // nested deps and sometimes might be missing altogether, depending on type.
  for (const dep of referencedFiles) {
    crawlDependencies.add(dep);
  }

  return [...crawlDependencies];
```

rollup.config.js
jakearchibald: Client is the browser, yeah. Our build system is also JavaScript, so the code that runs to generate the static html is JavaScript, but not running in the browser (running in Node).

We have three root component directories:

- `static-build` - these components only execute as part of the build
- `client` - these components only execute in the browser
- `shared` - these components execute in both

Right now the URL pattern only works in the client code. This is a bit of a gotcha with the shared code, since it's easy to miss that it isn't working.

It might just be a matter of moving it to the common plugins.

RReverser: Oh ok, I see, so this is not something that's happening now (that's why I didn't see the issue), but something that might happen if we convert more of those URL imports to this pattern. Yeah, makes sense to move to common plugins to prevent this issue.

RReverser: Hm, I actually don't see that happening even now in static HTML. If I add

```js
//@ts-ignore
console.log(new URL('./imgs/demos/icon-demo-large-photo.jpg', import.meta.url).href);
```

to `src/shared/prerendered-app/Intro/index.tsx` as you suggested, then the `.tmp/build/static/index.html` output contains:

```js
console.log(new URL("/c/icon-demo-large-photo-18da387a.jpg", module.uri).href);
```

Which seems fine and isn't a `file:` URL.

Did you mean a different output mode / file?

RReverser: (correspondingly, moving to `commonPlugins` doesn't seem to make a difference)

jakearchibald: Here's a demo https://deploy-preview-1038--squoosh.netlify.app/ (PR: https://github.com/GoogleChromeLabs/squoosh/pull/1038).

If you view-source you'll see:

```html
<img class="_demo-icon_5wn1n_222" src="file:///opt/build/repo/.tmp/build/shared/prerendered-app/Intro/imgs/demos/icon-demo-device-screen.jpg" alt="Device screen" />
```

If I move the plugin to `commonPlugins` then I get:

```html
<img class="_demo-icon_5wn1n_222" src="file:///c/icon-demo-device-screen-5d52d8b9.jpg" alt="Device screen" />
```

…which is closer, but still not right.

RReverser: Interesting. Which folder is that HTML generated in? I only did "search in files" for "file:" and couldn't find this one.

RReverser: Oh interesting, I can repro with your change. I wonder why it inlines `file:` when it's in HTML but correct URL when it's something like my `console.log`...

RReverser: I suppose it's because Rollup generates stuff like

```js
const deviceScreenIcon = new URL('./imgs/demos/icon-demo-device-screen.jpg', (typeof document === 'undefined' ? new (require('u' + 'rl').URL)('file:' + __filename).href : (document.currentScript && document.currentScript.src || new URL('shared/prerendered-app/Intro/index.js', document.baseURI).href))).href;
```

so in browser where `document` is defined, it works fine, but in static render it falls back to the `file:` branch.

I'll look into it a bit more.

jakearchibald: Where were you looking for the log? In the console of the build, or in the browser?

RReverser: In the browser + looking manually in files. I'm not sure how to invoke the build itself (it's not just `npm run build` IIUC?).

RReverser: Ah it is. I guess I just wasn't looking for that log line in the CLI, I see it now.

jakearchibald: I guess we could fix that bit in another PR. It's less important than the wasm bloat.

RReverser: Let me take another look and if not, yeah, let's postpone.

At least now I see what's happening here. The `new URL(..., import.meta.url)` is meant to transform in a way where 2nd arg becomes a valid fully-specified URL. However, there is no such URL Rollup can substitute in Node.js since it doesn't know the domain application will end up with, and substituting with just relative URL like `/` would make `new URL` invocation invalid.

I'm looking at options of the plugin to see if we can work around this issue somehow... If not, maybe we'll be able to provide target domain at compile-time (like take it from Netlify or dev environment).

RReverser: > I'm looking at options of the plugin to see if we can work around this issue somehow...

Yeah it doesn't seem like plugin supports that.

RReverser: Okay so I can't think of a good solution to support this right now, but at least I've added a custom `resolveImportMeta` handler to static build that prevents usages like the one in your example and throws a build-time error for now.

It still accepts all the usages in this PR, so it seems like a good balanced solution until we find a way to support `new URL` in static HTML too. WDYT?

jakearchibald: Sounds good

RReverser:
Ok so what should we do here next? It looks like sizes are an unfortunate but legitimate side-effect of upgrade (and different optimisations in LLVM upstream).

We can switch back to optimisation mode with smaller sizes, but then we lose encoding speed. Should I try and experiment with more modes and wait with this PR for now?

RReverser:
> Should I try and experiment with more modes and wait with this PR for now?

Trying one more thing, fingers crossed...

RReverser:
Yeah no, tried to switch to the legacy pass manager, but that increased size even more.

jakearchibald:
> Ok so what should we do here next? It looks like sizes are an unfortunate but legitimate side-effect of upgrade (and different optimisations in LLVM upstream).

Can you summarise why the increases are legitimate? As in, what are we winning for the size increase?

@surma what do you think? A 10%-30% increase on pretty much all our wasm seems pretty hefty, especially if we don't gain anything from it. But if it's our only way to stay up to date with emscripten, and they don't see the size increase as an issue, I'm not sure we have any other choice.

RReverser:
> Can you summarise why the increases are legitimate? As in, what are we winning for the size increase?

Maybe "legitimate" is not the right word. What I meant is that it's not a bug in the build system, but rather LLVM trying to optimise even more in newer versions, and inlining more functions into each other. Unfortunately, this time we don't win any noticeable performance gains (at least as far as I can tell from my quick tests at default settings).

I don't like the size increase either, which is why I raised the issue on Emscripten side which was tracked down to LLVM upgrade, but then there's not much we can do because size increase is expected at `-O3`.

We can revert to `-Os` like before https://github.com/GoogleChromeLabs/squoosh/pull/838 & co., but then we lose the original speed gains as well. I'd love to figure out if there's a specific LLVM pass we could disable to get previous perf gains without the size increase, but doesn't look so either - it's just that implementation of inlining changed under the hood, so it's "all or nothing" sort of situation.

surma:
Yeah, this is a tough one... It’s a one-off download that we cache, it’s not part of the critical loading path, we don’t update codecs _that_ frequently. I think we can get away with it. Especially because codecs are our main performance bottle neck right now.

(Part of the problem is probably that in native land — and therefore from LLVMs perspective — bigger binary size for speed is a cheap tradeoff.

jakearchibald:
I'm happy if you're happy @surma. Eventually we should make codec downloading opt-in, especially if they continue to get bigger, and we add more.

RReverser:
> Eventually we should make codec downloading opt-in

At the very least, we could probably respect the Save-Data mode (that Chrome Lite & other similar implementations send) and download things lazily in that mode instead of trying to make Squoosh work offline. But it's out of scope of this PR.

RReverser:
> I'm happy if you're happy @surma.

@jakearchibald just fyi - I can't merge this until you change your review on the PR.

jakearchibald (approved):

PR #1030, opened 2021-05-26
Update JPEG XL to use github repos.

veluca93:
Also update to latest main.

PR #1039, opened 2021-06-02
JPEG-XL: Better 'effort' and adding noise synth

jakearchibald:
This builds on https://github.com/GoogleChromeLabs/squoosh/pull/1030.

I changed our effort setting so it's the same as JXL's (they've switched from backward-speed to effort).

Also added noise synthesis. 

This shouldn't be merged before #1027


surma (approved):

jakearchibald:
Hold up on this, I'm bashing it into shape

jakearchibald:
@surma ready! Sorry about that, I didn't think you'd be awake 😄

src/client/lazy-app/Compress/Options/Range/style.css
```diff
@@ -35,7 +35,7 @@
   text-decoration-style: dotted;
   text-decoration-color: var(--main-theme-color);
   text-underline-position: under;
-  width: 48px;
+  width: 54px;
```
jakearchibald, line 38: Making room for bigger numbers

src/client/lazy-app/Compress/Options/Range/custom-els/RangeInput/index.ts
```diff
@@ -112,18 +112,24 @@ class RangeInputElement extends HTMLElement {
     this.dispatchEvent(retargetted);
   };
 
+  private _getDisplayValue(value: number): string {
+    if (value >= 10000) return `${(value / 1000).toFixed(1)}k`;
```
jakearchibald: Handling the display of long values in the range bubble

jakearchibald:
Actually, hold off on this. Chatting it through with JXL folks

jakearchibald:
Aaaaaand we're good to go. Sorry about that. @surma: review time!

src/client/lazy-app/Compress/Options/Range/custom-els/RangeInput/index.ts
```diff
@@ -112,18 +112,24 @@ class RangeInputElement extends HTMLElement {
     this.dispatchEvent(retargetted);
   };
 
+  private _getDisplayValue(value: number): string {
+    if (value >= 10000) return (value / 1000).toFixed(1) + 'k';
+
+    const labelPrecision =
+      Number(this.labelPrecision) || getPrescision(this.step) || 0;
```
RReverser: This function was there before, but nit: there's a typo in `getPrescision`.

jakearchibald: Good catch. Fixed!

src/features/encoders/jxl/client/index.tsx
```diff
@@ -199,6 +199,16 @@ export class Options extends Component<Props, State> {
                   Optimise for decoding speed (worse compression):
                 </Range>
               </div>
+              <div class={style.optionOneCell}>
+                <Range
+                  min="0"
+                  max="50000"
```
surma, line 205: Coming from this with a slight photography angle, I wonder if we should have `step=100`. Ideally, we’d actually make this a log2 scale, but happy to defer that to whenever we give our range sliders support for showing the actual value. WDYT?

surma (approved):
One remark, otherwise LGTM.

src/features/encoders/jxl/client/index.tsx
jakearchibald: Yeah, I'll add the step.

This setting is kinda hard to use right now since it's sensitive to the size of the image (since it maps image size to sensor megapixels). A large value looks stupid on a large image, but you need a large value to make a difference to a small image.

I've given that feedback to the JXL folks.

Hopefully eventually it'll automatically detect noise size and amplitude, and smooth the image accordingly.

PR #1054, opened 2021-06-15
Fix memory leaks

### LINKED ISSUE #1048: Memory leak when going from compress screen to home screen
I don't know this is whether or not a bug
once start compress image process will greatly improve memory usage

any solution can reduce memoery usage or not?

as you can see I just opened the [website](https://squoosh.app/)
the tab take about 30mb of memory 
image
When I upload a image about 1.8m, the memory is straight up, and then wait for compress done (using default options)
the memory is about 800mb
and then i close the website image editor, the memory is about 350mb usage, and cant no longer reduce
image
this is right or not? and there is a way to solve this or not? thank you guys

jakearchibald:
Fixes #1048

### DIFF
--- a/src/client/lazy-app/Compress/Output/custom-els/TwoUp/index.ts
+++ b/src/client/lazy-app/Compress/Output/custom-els/TwoUp/index.ts
@@ -34,6 +34,8 @@ export default class TwoUp extends HTMLElement {
    */
   private _everConnected = false;
 
+  private _resizeObserver?: ResizeObserver;
+
   constructor() {
     super();
     this._handle.className = styles.twoUpHandle;
@@ -44,13 +46,6 @@ export default class TwoUp extends HTMLElement {
     new MutationObserver(() => this._childrenChange()).observe(this, {
       childList: true,
     });
-
-    // Watch for element size changes.
-    if ('ResizeObserver' in window) {
-      new ResizeObserver(() => this._resetPosition()).observe(this);
-    } else {
-      window.addEventListener('resize', () => this._resetPosition());
-    }
 
     // Watch for pointers on the handle.
     const pointerTracker: PointerTracker = new PointerTracker(this._handle, {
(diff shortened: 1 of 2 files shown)


src/client/lazy-app/Compress/Output/custom-els/TwoUp/index.ts
```diff
@@ -84,20 +77,36 @@ export default class TwoUp extends HTMLElement {
...
+
     if (!this._everConnected) {
       this._resetPosition();
       this._everConnected = true;
     }
   }
 
+  disconnectedCallback() {
+    window.removeEventListener('keydown', this._onKeyDown);
+    window.removeEventListener('resize', this._onResize);
```
jakearchibald, line 98: We should have caught these at review time 😄. I guess we would have if it was a Preact component, but we're less experienced with web components.

src/client/lazy-app/Compress/Output/custom-els/TwoUp/index.ts
```diff
@@ -84,20 +77,36 @@ export default class TwoUp extends HTMLElement {
...
+    if (this._resizeObserver) this._resizeObserver.disconnect();
```
jakearchibald, line 99: I really don't think we should have to do this, but it's a direct cause of the leak. Filed https://bugs.chromium.org/p/chromium/issues/detail?id=1220041.

src/client/lazy-app/Compress/index.tsx
```diff
@@ -377,6 +377,7 @@ export default class Compress extends Component<Props, State> {
 
   componentWillUnmount(): void {
     updateDocumentTitle({ loading: false });
+    this.widthQuery.removeListener(this.onMobileWidthChange);
```
jakearchibald, line 380: This doesn't seem to be causing a leak. Or, if it is, it's small. But, good to add it for safety.

surma (approved):
Nice work. LGTM.

src/client/lazy-app/Compress/index.tsx
jakearchibald: (I was wrong here, it was causing a leak, it was just hard to see due to the resizeObserver issue)

PR #1067, opened 2021-06-20
Implement a lossless signal

### LINKED ISSUE #1060: [lib] New feature: Signal if an image is lossless or lossy
I have a use case where I need to know if an input image is compressed with a lossy or lossless algorithm. Figuring this out is easy for codecs like png and jpg which only support one type, but it's not that simple when I want to test a webp or jxl image.
Ideally, I would be able to get this information directly from Squoosh.

I have build a draft implementation in Squoosh, and it seems relatively simple to add this. I am detecting lossy/lossless encoding by reading the first two chunks of every image. Please let me know if there's a smarter way to get this info :D
The signal would be added to `libSquoosh` as `Image.decoded.lossless: true | false`.

Apart from helping me out, I also think this could be very useful for Squoosh itself. Currently, Squoosh will default to encoding any image lossy, even if it's an SVG. With this signal, it should be possible to automatically toggle the lossy/lossless switch in the target encoder, matching the encoding type for the input image, resulting in much better defaults.

Adding a system to change defaults based on the input image seems so complex that it's probably better to let the core team do it (or do you disagree?), but if I can get a green light on this, I'll be happy to implement the signal :)

atjn:
Closes #1060, please read that first.

This requires some more research/testing before I'm certain that it works properly, but assuming I get the green light, this is pretty much what the final PR will look like :)

### DIFF
--- a/libsquoosh/README.md
+++ b/libsquoosh/README.md
@@ -117,6 +117,7 @@ console.log(await image.decoded);
     width: 4606,  //pixels
     height: 2591  //pixels
   },
+  lossless: true,  //whether the source image was encoded losslessly
   size: 2467795  //bytes
 }
 ```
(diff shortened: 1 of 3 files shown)


atjn:
This seems unlikely to ever be included. I am closing the PR, but #1060 remains an open suggestion until a maintainer closes it :)

PR #1072, opened 2021-06-24
Add charset utf-8 meta tag

coliff:


surma:
Hm, I’m not sure we need that provided we send a `content-type: text/html; charset=UTF-8` HTTP header.

@jakearchibald wdyt?

coliff:
Ah ok. It's recommended in the Google HTML style guide so I thought it should always be included:
https://google.github.io/styleguide/htmlcssguide.html#Encoding

coliff:
should this be merged or shall I close it?

PR #1074, opened 2021-06-24
[libsquoosh] Add thread support to AVIF in Node

surma:


libsquoosh/src/codecs.ts
```diff
@@ -1,5 +1,13 @@
 import { promises as fsp } from 'fs';
 import { instantiateEmscriptenWasm, pathify } from './emscripten-utils.js';
+import { threads } from 'wasm-feature-detect';
+import { cpus } from 'os';
+
+// We use `navigator.hardwareConcurrency` for Emscripten’s pthread pool size.
+// This is the only workaround I can get working without crying.
```
nucliweb, line 7: To maintain consistency with the previous sentence
```suggestion
// This is the only workaround we can get working without crying.
```

libsquoosh/src/emscripten-utils.ts
```diff
@@ -10,10 +10,14 @@ export function pathify(path: string): string {
 export function instantiateEmscriptenWasm<T extends EmscriptenWasm.Module>(
   factory: EmscriptenWasm.ModuleFactory<T>,
   path: string,
+  workerWasm: string = '',
```
RReverser: As far as I can tell, this should be called `workerJS` not `workerWasm`.

libsquoosh/src/emscripten-utils.ts
```diff
@@ -10,10 +10,14 @@ export function pathify(path: string): string {
...
 ): Promise<T> {
   return factory({
-    locateFile() {
-      return pathify(path);
+    locateFile(requestPath) {
+      if (requestPath.endsWith('.wasm')) return pathify(path);
```
RReverser, line 21: I'll admit I don't understand this logic / why we need different logic for different types of files here. Maybe a comment would be helpful?

surma: The build system for libSquoosh doesn’t understand `new URL(..., import.meta.url)` for assets yet, so we need to manually inject the updated paths here.

libsquoosh/src/emscripten-utils.ts
surma: Good point

libsquoosh/src/emscripten-utils.ts
surma: I’ll add a comment

Issue #1012, opened 2021-05-11
[CLI] Limit the number of decoded images in memory


illycz:
I'm trying edit folder with approx 200 images within `npx @squoosh/cli --mozjpeg '{"quality":80}' -d . ./*.jpg` and get memory error:
`1.54MBRangeError: WebAssembly.Memory(): could not allocate memory
    at /Users/illy/.npm/npx/83877/lib/nodemodules/@squoosh/cli/build/index.js:7:176011
    at instantiateEmscriptenWasm (/Users/illy/.npm/npx/83877/lib/nodemodules/@squoosh/cli/build/index.js:7:153532)
    at Object.dec (/Users/illy/.npm/npx/83877/lib/nodemodules/@squoosh/cli/build/index.js:7:378727)
    at decodeFile (/Users/illy/.npm/npx/83877/lib/nodemodules/@squoosh/cli/build/index.js:7:520969)
    at async MessagePort.<anonymous> (/Users/illy/.npm/npx/83877/lib/nodemodules/@squoosh/cli/build/index.js:7:493320)
Worker error:  Error [RuntimeError]: abort(RangeError: WebAssembly.Memory(): could not allocate memory). Build with -s ASSERTIONS=1 for more info.`

Terminal blinking and process never ending.

Thanks

illycz:
Few images working as expected...

user69nice:
> I'm trying edit folder with approx 200 images within `npx @squoosh/cli --mozjpeg '{"quality":80}' -d . ./*.jpg` and get memory error:
[...]

PR #1075, opened 2021-06-25
Improve error messages when decoding using mozjpeg

jcao219:
Aims to make errors from mozjpeg decode/encode more visible.

Some more questions:
- Is it okay for the whole batch to fail if a single image fails encoding/decoding?
- ~The web app code might also need an update to handle the errors from the C++ code?~


libsquoosh/src/index.js
```diff
@@ -35,6 +35,10 @@ async function decodeFile({ file }) {
     throw Error(`${file} has an unsupported format`);
   }
   const rgba = (await encoders[key].dec()).decode(new Uint8Array(buffer));
+  // The C++ code returns a string if there's an error.
```
jcao219: Not sure of a better way to return errors from the C++ code.

cli/src/index.js
```diff
@@ -114,19 +114,34 @@ async function processFiles(files) {
...
-        file,
-        size: (await image.decoded).size,
-        outputs: [],
-      });
-      progress.setProgress(++decoded, files.length);
-      return image;
-    }),
-  );
+  let decodedFiles;
+  try {
```
jcao219: The mozjpeg error kept causing UnhandledPromiseRejections, since it bubbles up to this top level, which does not have an error handler. I'm thinking we should add a top-level `try { } catch {}` that wraps this `processFiles` function.

jcao219:
The C++ error handling code was mostly taken from the example here: https://github.com/mozilla/mozjpeg/blob/323469818c94374c200df2fdd11695dc4ece5b6f/example.txt#L242

libsquoosh/src/index.js
```diff
@@ -8,6 +8,16 @@ import { autoOptimize } from './auto-optimizer.js';
 
 export { ImagePool, encoders, preprocessors };
 
+// The C++ code returns a string when an error occurs.
+// This function is used to evaluate return values from wasm,
+// and throw an error if applicable.
+function handleStringError(operation, value) {
```
surma: This seems like an asserting function. It should probably be called something like `assertNoCPPError` or something. `value` as first parameter, error message as second parameter, somewhat analogous to `console.assert()`. WDYT?

libsquoosh/src/index.js
```diff
@@ -66,15 +77,15 @@ async function encodeImage({
     const optionToOptimize = encoders[encName].autoOptimize.option;
     const decoder = await encoders[encName].dec();
     const encode = (bitmapIn, quality) =>
-      encoder.encode(
+      handleStringError('encode', encoder.encode(
```
surma: Let’s not nest these functions calls. That seems confusing.

surma (approved):
Overall this seems reasonable. I’ll tag @atjn in case he has some time to take a look at the CLI/libSquoosh changes.

Also taggin @RReverser to take a look at the usage of `setjmp`. I vaguely remember that `setjmp` had some gotchas with Emscripten, but maybe that’s been resolved since the last time I looked at it.

surma (changes requested):
Clicked the wrong button 😅

libsquoosh/src/index.js
jcao219: Good point -- it really is an assertion function!

libsquoosh/src/index.js
jcao219: Ah, I agree. Fixed!

atjn (commented):
I haven't done any thorough testing, but this seems good to me. It's really great to see some improvements to error handling 🥳

I have two minor comments:

IMO the code would become much more readable if we renamed `assertNoCPPError` to `detectCPPError`. You already renamed this once @surma, do you agree with me?

I think it's dangerous that any returned string is now implicitly detected as a CPP error. Ideally, it should be explicitly called out, perhaps by prepending the string with `CPP_ERROR:` or by instead returning an object with a `CPP_ERROR` key.

atjn:
> Is it okay for the whole batch to fail if a single image fails encoding/decoding?

Not if you ask me. I think it would make a lot of sense to implement non-blocking per-image error handling at the same time as fixing #1012.

RReverser:
> I vaguely remember that `setjmp` had some gotchas with Emscripten, but maybe that’s been resolved since the last time I looked at it.

Yeah it doesn't officially support jumping up the stack as that's UB, which is exactly what we do here.

Instead, I'd just throw a C++ exception (which will become a regular JS error) in that exit callback. Just need to make sure to also deallocate any manually allocated resources.

jcao219:
> > I vaguely remember that `setjmp` had some gotchas with Emscripten, but maybe that’s been resolved since the last time I looked at it.
> 
> Yeah it doesn't officially support jumping up the stack as that's UB, which is exactly what we do here.
> 
> Instead, I'd just throw a C++ exception (which will become a regular JS error) in that exit callback. Just need to make sure to also deallocate any manually allocated resources.

I'm in favor of throwing a C++ exception. That way, we only need try/catch in the caller JS code. The only thing to watch out for is we would have to add the `-fexceptions` build flag in order to enable exceptions support.

RReverser:
> The only thing to watch out for is we would have to add the `-fexceptions` build flag in order to enable exceptions support.

AFAIK (but I might be wrong) throwing works even without it; if it doesn't, you can throw JS exception directly via `EM_ASM({ throw new Error(...) });` instead - it would be lower overhead.

atjn (approved):

jcao219:
> > The only thing to watch out for is we would have to add the `-fexceptions` build flag in order to enable exceptions support.
> 
> AFAIK (but I might be wrong) throwing works even without it; if it doesn't, you can throw JS exception directly via `EM_ASM({ throw new Error(...) });` instead - it would be lower overhead.

I wasn't able to get throwing to work. However, the `EM_ASM` idea works really well. It's exactly the mechanism I was looking for in order to pass an error in a more streamlined way. Thanks for the tip!

Next I need to look into whether all resources are properly deallocated after the callback throws the JS error.

RReverser:
> Next I need to look into whether all resources are properly deallocated after the callback throws the JS error.

Note: you might need to pay special caution when dealing with input params - `std::string` input would normally be deallocated when going out of scope, but in this case we need to do something manual.

Alternatively, we could just create a new Wasm instance per input file. This will create a bit of overhead but otherwise will be a lot easier to handle, as entire Wasm memory will be deallocated all at once.

Or, finally, we could just use `-fexceptions` in which case C++ takes care of deallocating at least managed objects.

jcao219:
> > Next I need to look into whether all resources are properly deallocated after the callback throws the JS error.
> 
> Note: you might need to pay special caution when dealing with input params - `std::string` input would normally be deallocated when going out of scope, but in this case we need to do something manual.
> 
> Alternatively, we could just create a new Wasm instance per input file. This will create a bit of overhead but otherwise will be a lot easier to handle, as entire Wasm memory will be deallocated all at once.
> 
> Or, finally, we could just use `-fexceptions` in which case C++ takes care of deallocating at least managed objects.

Thanks for the tip about the `std::string` input! Would definitely not want to leak that.

I'm still investigating the options to solve the issue. I don't like `-fexceptions` because of the extra overhead in generated code, and because we'd need an extra C++ function to extract the error message as a string from the exception pointer returned to JS. 

One question regarding creating a new Wasm instance per input file. Is that not what's already happening with https://github.com/GoogleChromeLabs/squoosh/blob/ad5002c79cccab7291b7bbb5d9f6270087ca6d4e/libsquoosh/src/index.js#L37 calling https://github.com/GoogleChromeLabs/squoosh/blob/ad5002c79cccab7291b7bbb5d9f6270087ca6d4e/libsquoosh/src/codecs.ts#L256 ?

surma:
> One question regarding creating a new Wasm instance per input file. Is that not what's already happening?


In the CLI: yes. In the PWA that might not always be the case.

PR #1081, opened 2021-06-30
Remove unused variable

atjn:
Nitpick: found a lone variable hanging around in the code.

As far as I can tell, this isn't needed and can be safely removed, but I could be missing something.


nucliweb (approved):

surma (approved):
Yeah, I think `_nextWorker()` is supposed to return `undefined` when the worker pool is closed, so that seems correct :)

PR #1082, opened 2021-07-01
fix: Don't overwrite value while typing

### LINKED ISSUE #316: Numbers cannot start with 1 when editing colors for OptiPNG
Steps to reproduce:
1. In `Compress` mode, switch to `OptiPNG`.
2. Toggle to check on `Reduce palette` checkbox.
3. Focus on or click the `input[type=number]` element of `Colors:`
4. And type `17` and see the value 'jumps' to `27` instead.

Because of the input validation that only allows a minimum value of `2`, this breaks the final value when user tries to enter anything starts with `1` but less than `2`.

bfanger:
Allows entering number like "16" when the minimum is set to 2

Fixes #316

after-fix


jakearchibald (approved):
LGTM! Nice one!

Issue #984, opened 2021-04-01
Be able to tell more easily when codec INCREASES filesize


tomByrer:
**Is your feature request related to a problem? Please describe.**
Tricky to tell at a glance when the filesize is increased by a codec.

**Describe the solution you'd like**
A more distinct `+` symbol, italics, &/or different color of larger values

**Additional context**
discovery: https://github.com/GoogleChromeLabs/squoosh/issues/960#issuecomment-811880408

Thanks for your work & consideration! :)

kosamari:
I do agree up and down arrows are very subtle. 

![up](https://user-images.githubusercontent.com/4581495/124143014-c0b50c80-da58-11eb-8ada-e4549a0f88be.png)
![down](https://user-images.githubusercontent.com/4581495/124143027-c3176680-da58-11eb-9514-ce6dcc92febc.png)

I think this is a good suggestion - @argyleink can you chime in on UI design? :)

argyleink:
Chimed! 🙂

PR #1096, opened 2021-07-09
Prototyping alternative size change indicators

argyleink:
This PR is thoughts on #984

I ended up preferring a `+` for when the image will be larger in size, and the `↓` unicode character for when it's smaller:

Screen Shot 2021-07-09 at 10 40 12 AM
Screen Shot 2021-07-09 at 10 40 19 AM

I also switched to the monospace font for the arrow and plus, which in general should use a bolder symbol across platforms. if this arrow is problematic, we could use an inline SVG instead and own the vector ourselves to normalize it. 

All in all, _I felt_ switching to a plus instead of an up arrow was enough to take the subtly out, what do y'all think?

### DIFF
--- a/src/client/lazy-app/Compress/Results/index.tsx
+++ b/src/client/lazy-app/Compress/Results/index.tsx
@@ -110,7 +110,7 @@ export default class Results extends Component<Props, State> {
               <div class={style.percentOutput}>
                 {diff && diff !== 1 && (
                   <span class={style.sizeDirection}>
-                    {diff < 1 ? '↓' : '↑'}
+                    {diff < 1 ? '↓' : '+'}
                   </span>
                 )}
                 <span class={style.sizeValue}>{percent || 0}</span>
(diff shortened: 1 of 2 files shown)


nucliweb (approved):
👏

surma:
I like the “+” a lot. But I am not sure I like mixing “+” and “↓” ... Gonna wait for @jakearchibald to weigh in here.

mikedelgaudio:
Definitely like the +, I'd agree with @surma maybe we can get creative on a "-" sign? 

I'm wondering could contrasting colors help us here to provide a clearer distraction between a positive or negative change?
One caveat to this though is Squoosh _reduces_ image size so a negative sign would be a positive color of some sort...

jakearchibald:
I still prefer 100% to mean no change, 50% to mean half the size, and 200% to mean double the size… but I realise I'm in the minority there 😄 

I agree the symbol change works well. I don't even mind the switch between the + and arrow. Switching between + and - might look pretty subtle without something like a colour change to emphasize things.

PR #1110, opened 2021-07-27
Convert remaining JS to TS in libSquoosh

styfle:
Follow up to @ergunsh's work to convert remaining src code for libSquoosh to TypeScript.

There is more usage of `any` than I would like, so I'm open to suggestions here.

cc @surma


surma:
@ergunsh Do you mind taking a look?

I am okay with taking some shortcuts for now (like `any`) and fixing those in later PRs!

libsquoosh/src/codecs.ts
```diff
@@ -9,6 +9,23 @@ import { cpus } from 'os';
   hardwareConcurrency: cpus().length,
 };
 
+interface DecodeModule extends EmscriptenWasm.Module {
+  decode: (data: Uint8Array) => any;
```
ergunsh: I think we know the return value of the `decode` which is ImageData. So, we can use ImageData here as the return type, wdyt?

libsquoosh/src/codecs.ts
```diff
@@ -9,6 +9,23 @@ import { cpus } from 'os';
...
+}
+
+interface EncodeModule extends EmscriptenWasm.Module {
```
ergunsh: I think types for encode modules for the codecs already exists: so maybe we can use them instead of using this EncodeModule (so that opts will be correctly typed for each codec).

Here is an example import for the MozJPEG codec:
```
import type {
  MozJPEGModule as MozJPEGEncodeModule,
  EncodeOptions as MozJPEGEncodeOptions,
} from '../../codecs/mozjpeg/enc/mozjpeg_enc';
```

libsquoosh/src/codecs.ts
```diff
@@ -421,7 +450,7 @@ export const codecs = {
       await oxipngPromise;
       return {
         encode: (
-          buffer: Uint8Array,
+          buffer: Uint8ClampedArray | ArrayBuffer,
```
ergunsh, line 459: This is more of a question than the review comment, why did we need to change this to `Uint8ClampedArray | ArrayBuffer`?

libsquoosh/src/index.ts
```diff
@@ -5,10 +5,18 @@ import { promises as fsp } from 'fs';
 import { codecs as encoders, preprocessors } from './codecs.js';
 import WorkerPool from './worker_pool.js';
 import { autoOptimize } from './auto-optimizer.js';
+import type ImageData from './image_data';
 
 export { ImagePool, encoders, preprocessors };
+type EncoderKey = keyof typeof encoders;
+type PreprocessorKey = keyof typeof preprocessors;
+type FileLike = Buffer | ArrayBuffer | string | { buffer: Buffer };
```
ergunsh: I think we need `ArrayBufferView` in here as well (ArrayBuffer.isView is a type guard for `ArrayBufferView`). Then, we can remove `{ buffer: Buffer }` typing from FileLike since ArrayBufferView already contains `buffer` property. WDYT?

libsquoosh/src/index.ts
```diff
@@ -107,26 +131,35 @@ async function encodeImage({
...
-// both decoding and encoding go through the worker pool
-function handleJob(params) {
-  const { operation } = params;
-  switch (operation) {
+type EncodeParams = { operation: 'encode' } & Parameters<typeof encodeImage>[0];
+type DecodeParams = { operation: 'decode' } & Parameters<typeof decodeFile>[0];
+type PreprocessParams = { operation: 'preprocess' } & Parameters<
+  typeof preprocessImage
+>[0];
+export type JobMessage = EncodeParams | DecodeParams | PreprocessParams;
```
ergunsh: Instead of exporting this from here and using it in WorkerPool, would it make sense to make WorkerPool accept a generic `JobMessage` type so that we can isolate its implementation and definition from its usage?

libsquoosh/src/index.ts
```diff
@@ -161,14 +195,20 @@ class Image {
   /**
    * Define one or several encoders to use on the image.
    * @param {object} encodeOptions - An object with encoders to use, and their settings.
-   * @returns {Promise<undefined>} - A promise that resolves when the image has been encoded with all the specified encoders.
+   * @returns {Promise<void>} - A promise that resolves when the image has been encoded with all the specified encoders.
    */
-  async encode(encodeOptions = {}) {
+  async encode(
+    encodeOptions: {
```
ergunsh, line 216: I'm not sure if this is the correct typing for the `encode` method. For example, this is how it is called in the example in readme:
```
const encodeOptions = {
  mozjpeg: {}, //an empty object means 'use default settings'
  jxl: {
    quality: 90,
  },
}
await image.encode(encodeOptions);
```

It should be something like:
```
{
optimizerButteraugliTarget?: number;
maxOptimizerRounds?: number;
} & {
    [key: EncoderKey]: any; // any is okay for now
}
```

I'm not sure about the `[key: EncoderKey]` syntax though 😅

libsquoosh/src/worker_pool.ts
```diff
@@ -1,13 +1,15 @@
 import { Worker, parentPort } from 'worker_threads';
+// @ts-ignore
```
ergunsh: I think `TransformStream` should already have types. Did you encounter any errors while using it?

libsquoosh/src/worker_pool.ts
```diff
@@ -1,13 +1,15 @@
...
 import { TransformStream } from 'web-streams-polyfill';
+import type { JobMessage } from './index';
```
ergunsh: I added the comment above but wanted to add here as well: wdyt about not importing something from `index` for the WorkerPool definition and implementation; to make it easier to reason about?

I think we can have a generic WorkerPool definition (essentially it dispatches a message (input) and returns a result (output), so it can be something like `class WorkerPool<I, O>`). Then we can use it in index as `WorkerPool<JobMessage, any>`. WDYT?

ergunsh (commented):
The changes here (including the any usages) make a lot of sense for enabling us to move forward with making libSquoosh available on web!

However, I think, before exposing these types to the users; we need to get rid of `any` usages and provide better types overall. (which I can work on on top of this 😊)

So, thanks a lot! 🥳

libsquoosh/src/codecs.ts
surma: +1, decoders will always return `ImageData`
```suggestion
  decode: (data: Uint8Array) => ImageData;
```

libsquoosh/src/codecs.ts
surma: Technically `ImageData` always gives `Uint8ClampedArray`, not sure why this wasn’t a type error before.

libsquoosh/src/codecs.ts
styfle: I tried that but there seems to be a mismatch with the ImageData in this file and the one in `index.ts`

```ts
interface DecodeModule extends EmscriptenWasm.Module {
  decode: (data: Uint8Array) => ImageData;
}
```

image

libsquoosh/src/codecs.ts
styfle: That seems like the correct solution but it won't work for me here:

image

libsquoosh/src/codecs.ts
styfle: This was to make sure all the `encode()` signatures are the same for each codec

libsquoosh/src/index.ts
styfle: ArrayBufferView didn't work with the existing code

image

libsquoosh/src/worker_pool.ts
styfle: image

styfle:
@ergunsh If you want to make changes to this branch and push them, feel free (that might be easier for you)

libsquoosh/src/index.ts
styfle: Yeah that would be ideal. The code as written today has a static method (useThisThreadAsWorker) which can't use generics from the class. It also has a helper function outside the class (jobPromise) which needs to use the message as well.

styfle:
@surma @ergunsh Thoughts?

We could also chat in Slack. Send me an email (steven at vercel.com) and we'll invite you to the Vercel workspace 🤗

libsquoosh/src/codecs.ts
ergunsh: Can you try channg the
'''
ImageData: import(…)
'''

To 'ImageData: import(…).ImageData' since and then using the ImageData in decoder?

ergunsh:
> @surma @ergunsh Thoughts?
> 
> 
> 
> We could also chat in Slack. Send me an email (steven at vercel.com) and we'll invite you to the Vercel workspace 🤗 

Hi @styfle, I'll be able to review this tomorrow. I'm so sorry for the delay 😔 

Thanks a lot for the comments/answers!

libsquoosh/src/codecs.ts
ergunsh: I think I have made a mistake there while defining the global `ImageData`. We need to use this:
```
  type ImageData = import('./image_data.js').default;
```

If we do this, it works with ImageData return type above.

libsquoosh/src/codecs.ts
ergunsh: Hihiiim, it is strange. Here is what I tried and the only error we get is the return value says `Uint8Array | null`. So, we might need to handle `null` for the `encode` calls as well.

Here is my version of `decode.ts`: https://gist.github.com/ergunsh/93d0ddadd72d7f383fde952bbd45de31

WDYT?

libsquoosh/src/index.ts
ergunsh: Yes, that's strange and I didn't understand correctly why. However, I think it would be better to use `ArrayBufferView` in type definition and do a type assertion to `any` for `file` in the place with a TODO to figure it out why we need that. (i.e. maybe a type guard in the if clauses also checks for Buffer and we might not need `file instanceof Buffer` check in there) WDYT?

libsquoosh/src/worker_pool.ts
ergunsh: That is so strange, I do not receive that error. Did you check if you installed the dependencies and can you also check if the `build` fails with the same error?

libsquoosh/src/index.ts
ergunsh: I think, we can move `jobPromise` inside the class and for now, add the generic parameters to `useThisThreadAsWorker` like `static useThisThreadAsWorker<I, O>(cb: (msg: I) => O)`. Then, we can find a better way to use the generics instead of using the static method.

libsquoosh/src/index.ts
ergunsh: It looks like the correct syntax is `[key in EncoderKey]?: any; // any is okay for now`. WDYT?

ergunsh:
Hey hey, I have also pushed my changes for the comments I added above to here my own fork (I don't have access rights for your fork 😅)

Here is the commit: https://github.com/ergunsh/squoosh/commit/fafcf97f0c76f34401a9f5bb8044a4c112dd224a

If you feel like it, I think you can merge the changes from there. It addresses all the comments I added :)

libsquoosh/src/worker_pool.ts
styfle: `npm run build` works fine, but VS Code is not happy

If you hit F12 (Go To Definition), where does it take you?

libsquoosh/src/codecs.ts
styfle: We can change this to `Uint8ClampedArray` (without the union) if we change this line:

https://github.com/GoogleChromeLabs/squoosh/blob/fafcf97f0c76f34401a9f5bb8044a4c112dd224a/libsquoosh/src/index.ts#L128

Basically from `bitmapIn.data.buffer` to `bitmapIn.data`

ergunsh: I think it makes sense to do that change, then it would be nice if we can test the library by hand to make sure it doesn't break anything (I don't expect it to so but... 😅)

libsquoosh/src/worker_pool.ts
ergunsh: It goes to here for me: `<squoosh-path>/libsquoosh/node_modules/web-streams-polyfill/dist/types/ts3.6/polyfill.d.ts`

ergunsh (approved):
LGTM with https://github.com/GoogleChromeLabs/squoosh/pull/1110#discussion_r684384164

@surma can you take a final look? :)

libsquoosh/src/worker_pool.ts
styfle: I tried several different ways to install and I never get a `./libsquoosh/node_modules` directory.

- `npm install`
- `npm ci`
- `yarn install`

Then I realized that workspaces aren't set up so I did `cd ./libsquoosh && npm ci` and that worked 😅

surma (approved):
There are some `as` declarations that I’d want to clean up, but no need to block this PR on this now. I can get to those together with @ergunsh.

Thanks a bunch, @styfle !

PR #1115, opened 2021-08-04
Create an upper limit for scaling

### LINKED ISSUE #1091: Cap maximum zoom %
Current app lets zoom in without max. 

sort of low priority, nice-to have.

rphsoftware:
This pull request is an attempt to create an upper limit to how far the user can zoom in, in order to prevent bugs related to floating point inaccuracy and reaching "Infinite" zoom when scrolling far enough

The reason this changes "setTransform" instead of the lower level "_updateTransform" function is to prevent incorrect x and y values from being applied.

Note that this current implementation seems to behave erratically when after zooming in to the max value the user pans the image and attempts to zoom in further. Additionally, my current implementation of correcting for the new zoom value seems to be slightly inaccurate and shift the image when reaching the max value (hence the return if the scale is already maximum).

Also, the number input box for zoom probably needs to have a `max` attribute set to the same value as what's decided to be the maximum.

This should fix #1091

### DIFF
--- a/src/client/lazy-app/Compress/Output/custom-els/PinchZoom/index.ts
+++ b/src/client/lazy-app/Compress/Output/custom-els/PinchZoom/index.ts
@@ -81,6 +81,7 @@ function createPoint(): SVGPoint {
 }
 
 const MIN_SCALE = 0.01;
+const MAX_SCALE = 655.36; // 65536%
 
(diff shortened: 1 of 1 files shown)


jakearchibald:
Fixed in https://github.com/GoogleChromeLabs/squoosh/pull/1128

Issue #1084, opened 2021-07-01
[CLI] Make it run on the web


kosamari:
Run command from Web interface 
TODO: needs more detail on implementation

styfle:
Are you looking to make `@squoosh/cli` work on the web, or just `@squoosh/lib`?

surma:
The CLI would remain a thing wapper for the command line around `@squoosh/lib`  :)

styfle:
Great! What are the steps to get it working on the web?

surma:
Sorry for taking so long to reply. 

cc @atjn as he might have opinions on this.
Also cc @ergunsh, who is working on TypeScriptifying libSquoosh, in case he has thoughts/concerns.

---

To get libSquoosh running on the web I was thinking something along these lines:

- _Remove_ the capabilities to pass file paths to `ImagePool`. Loading from the file system should be the responsibility of the CLI, not the lib.
- Remove the `web-streams-polyfill` from the library and generate a node-specific output instead.
- Expose a hook/async callback to load Wasm modules. Since Node, Deno and web will all rely on different APIs to load a .wasm file.
- Use `new URL(..., import.meta.url)` to reference .wasm files so that bundlers can handle the .wasm assets correctly.

Did I miss anything? What do y’all think?

atjn:
That seems good to me. One quick note:

> Remove the capabilities to pass file paths to `ImagePool`. Loading from the file system should be the responsibility of the CLI, not the lib.

Being able to pass a file path directly to `ImagePool` is a pretty big nice-to-have IMO. If we have to build separate versions anyway, I would consider using the `fs.readFile` function to load images in Node, which has a lot of Node-specific nice-to-haves, including loading files directly from a path :)

I also have two questions that I think could be good to know for us non-core contributors:

1. Creating separate builds for so many different environments requires a lot of resources, not just initially but also to maintain the different builds in the future. Do you expect to be able to dedicate those resources? If not, you could just turn `libSquoosh` into a web-only package and create wrappers for Node and Deno that use [`puppeteer`](https://github.com/puppeteer/puppeteer) to run it. It might not be as elegant, but if the alternative is an abandoned package, or one that constantly has breaking bugs, then I'd prefer the inelegant solution :)

2. Is the plan to eventually use `libSquoosh` on the website? Should we be thinking on ways to make that happen, or would you like to keep them separate, at least for now?

[...]

PR #1123, opened 2021-08-16
Reduce `@squoosh/lib` Node.js API usage (make it run on the web)

styfle:
- Remove `fs` since its Node.js only
- Remove `os` since its Node.js only
- Remove other types of inputs besides ArrayBuffer

This lays the foundation for making `@squoosh/lib` run on the web.

Related to #1084


libsquoosh/src/index.ts
```diff
@@ -251,22 +250,26 @@ class Image {
...
    * Ingest an image into the image pool.
-   * @param {FileLike} image - The image or path to the image that should be ingested and decoded.
+   * @param {URL} url - The URL path to the image that should be ingested and decoded.
    * @returns {Image} - A custom class reference to the decoded image.
    */
-  ingestImage(image: FileLike): Image {
-    return new Image(this.workerPool, image);
+  ingestImage(url: URL): Image {
+    const file = await this.loadFile(url);
+    return new Image(this.workerPool, file);
```
styfle, line 277: We can't use `await` here unless we make `ingestImage` async as well.

@surma Thoughts?

surma: I think that’s fine!

surma: But we should still allow buffers to be passed to `ingestImage` as well.

styfle: Oh in that case, let's just use `ArrayBuffer` as the parameter here.

No need for a special `loadFile()` function since the caller can do `ingestImage(loadFile())` 🙂  

I just pushed a change that reduced a lot of Node.js specific code that gets us closer to running on the web 👍

surma (approved):
This looks great, thanks @styfle (and apologies for the delay).

I want to give both @ergunsh and @atjn a chance to weigh in on this, but then we can go ahead and get this merged (and I should probably whip up some sort of release notes as this is a pretty major breaking change).

libsquoosh/README.md
```diff
@@ -16,7 +16,8 @@ You can start using the libSquoosh by adding these lines to the top of your JS p
 
 ```js
 import { ImagePool } from '@squoosh/lib';
-const imagePool = new ImagePool();
+import { cpus } from 'os';
+const imagePool = new ImagePool(cpus().length);
 ```
 
 This will create an image pool with an underlying processing pipeline that you can use to ingest and encode images. The ImagePool constructor takes one argument that defines how many parallel operations it is allowed to run at any given time. By default, this number is set to the amount of CPU cores available in the system it is running on.
```
atjn:
```suggestion
This will create an image pool with an underlying processing pipeline that you can use to ingest and encode images. The ImagePool constructor takes one argument that defines how many parallel operations it is allowed to run at any given time.
```
This is no longer the case.

libsquoosh/README.md
```diff
@@ -26,11 +27,12 @@ This will create an image pool with an underlying processing pipeline that you c
 You can ingest a new image like so:
 
 ```js
-const imagePath = 'path/to/image.png';
+import fs from 'fs/promises';
+const file = await fs.readFile('./path/to/image.png');
 const image = imagePool.ingestImage(imagePath);
```
atjn:
```suggestion
const image = imagePool.ingestImage(file);
```

libsquoosh/README.md
```diff
@@ -26,11 +27,12 @@ This will create an image pool with an underlying processing pipeline that you c
...
 ```
 
-The `ingestImage` function can take anything the node [`readFile`][readfile] function can take, including a buffer and `FileHandle`.
+The `ingestImage` function can accept any [ArrayBuffer](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer) whether that is from `readFile()` or `fetch()`.
```
atjn:
```suggestion
The `ingestImage` function can accept any [`ArrayBuffer`][arraybuffer] whether that is from `readFile()` or `fetch()`.
```
This is how links are referenced in the rest of the document.

atjn: You'll have to manually replace the `readfile` link at line `172` with:
```md
[arraybuffer]: (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer)
```
It doesn't seem like I can make that change through this review feature 🤷

atjn (changes requested):
If the plan is to make this code fully platform-agnostic, then it looks great, I just found some minor issues :)

However, I believe [we agreed](https://github.com/GoogleChromeLabs/squoosh/issues/1084#issuecomment-885605209) that it would be nice to have wrappers for specific platforms, which would (among other things) still allow passing file paths to `ingestImage`, and still allow the default `worker_threads` to be set with `cpus.length`?
If that is still the plan, I would suggest making that wrapper before publishing a new version. These changes are very breaking, and it is pretty annoying for all parties involved to be forced to implement these new changes, only to see them be reverted in the near future.

libsquoosh/README.md
styfle: Oops, nice catch!

libsquoosh/README.md
styfle: No worries, added it in c21d371 👍

libsquoosh/README.md
styfle: Good point! Removed it 👍

styfle:
> it would be nice to have wrappers for specific platforms

I don't think it's worth the maintenance burden because the consumer only needs to add one line of code to get CPU count.

If we wanted to do this, we would need to make sure CI tested each of those platforms (it looks like there isn't even any tests yet at all 😬 )

Let's keep `@squoosh/lib` platform agnostic.

styfle:
I just realized, the only remaining Node.js API is `worker_threads`.

Should I swap that out for [web-worker](https://www.npmjs.com/package/web-worker) in this PR or create a new PR after this one is merged?

surma:
Hm, I’m torn. I kind agree with @atjn that this is pretty breaking and it’d be nice to provide wrappers. That being said, we have removed paths as parameters from `ingestImage()`, so to avoid a breaking change the wrapper would have to wrap that call as well. In the end, if there’s a time to do breaking changes it’s now. We are still in v0 territory and @squoosh/lib is still in its early days of adoption. I’m leaning towards making the breaking change here and publishing a `CHANGELOG.md` when we make the next release.

To your question, @styfle: Yes, let’s pull in `web-worker`. If there are problems with that, we should work with @developit to fix them.

styfle:
@surma @developit I tried to convert from Node.js `worker_threads` to the `web-worker` package but I think this will require a bigger refactor. In particular, I don't think there's a way to check for `isMainThread` or `parentPort`.

Any thoughts on how to restructure this code to make `web-worker` fit here?

https://github.com/GoogleChromeLabs/squoosh/blob/816d1f92fdd60c4492cf55d75d729db40f64bb0e/libsquoosh/src/index.ts#L281-L283

libsquoosh/README.md
```diff
@@ -169,3 +170,4 @@ const encodeOptions: {
 [codecs.ts]: https://github.com/GoogleChromeLabs/squoosh/blob/dev/libsquoosh/src/codecs.ts
 [butteraugli]: https://github.com/google/butteraugli
 [readfile]: https://nodejs.org/api/fs.html#fs_fspromises_readfile_path_options
```
atjn:
```suggestion
```

atjn: This is no longer used

atjn (approved):

developit:
I'm doing some work locally based on this PR to try to get `@squoosh/lib` working in both Node + Browser. Thanks for the awesome work @styfle - here's hoping the WASM loading ends up being possible 🤞

styfle:
> I'm doing some work locally based on this PR to try to get `@squoosh/lib` working in both Node + Browser.

@developit Thats great to hear!

Are you going to submit it as a new PR with all the changes or should we merge this PR first and you create a follow up?

surma:
@styfle Good Q. I’d say lets merge this (do you mind doing a re-base to make sure everything is still working with @ergunsh’s work?). And then Jason and I can follow up on the worker/wasm issue separately.

styfle:
@surma Everything is up to date and the CLI works as expected 👍

surma (approved):
Thanks so much, @styfle!

developit:
Thanks @styfle

Issue #905, opened 2020-12-10
Scrollwheel zooming is uncapped and zooming in too far breaks the panning.


rphsoftware:
**Describe the bug**
After zooming in really quickly with a "fast scroll" feature or an unlockable scroll wheel on the mouse, the zoom percentage can become infinite and make the image unpannable.

**To Reproduce**
1. Load any image
2. Scroll up a lot

**Expected behavior**
The zooming should be limited

**Version:**
 - OS w/ version: Arch Linux rolling
 - Browser w/ version: Chrome 87

**Is your issue related to the quality of image compression?**
No

**Additional context, screenshots, screencasts**
https://www.youtube.com/watch?v=id20luk5_xs

jakearchibald:
Fixed in 14c3d190e9c7c0d324e0ce9c3b6c60ec904cdf3a

PR #1128, opened 2021-08-21
Max cap on scale #905

Vishalghyv:
Adds max cap scaling

Adding very large value for scroll or scrolling too much for Zoom can lead to unexpected behavior in many cases.

Behavior I can reproduce:
In Mozilla, the browser simply crashes with large value
In chrome, zoom leads to Infinity value and the website stops responding

Issue linked #905 

Solutions I could think of :-
1 - To add max cap on scaling
2 - There could be a notification to user that large value of scaling can lead to breakage in computation, but user can click okay and continue.
3 - Check at which point the zooming will lead to no change in result, like when we reach a zoomed single pixel, then just keep increasing the scale without actually zooming.

The first was the easiest to implement. PTAL,
Thanks

### DIFF
--- a/src/client/lazy-app/Compress/Output/custom-els/PinchZoom/index.ts
+++ b/src/client/lazy-app/Compress/Output/custom-els/PinchZoom/index.ts
@@ -81,6 +81,7 @@ function createPoint(): SVGPoint {
 }
 
 const MIN_SCALE = 0.01;
+const MAX_SCALE = 1000000;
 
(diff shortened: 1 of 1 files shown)


jakearchibald:
1 seems like the best option to me, and it's consistent with what we do with minimum scale. Thank you!

jakearchibald (approved):

jakearchibald:
Deploying to live

Vishalghyv:
I can test it on the site :) thanks

Issue #1087, opened 2021-07-01
Add `image-rendering: pixelated` toggle


kosamari:
TODO: catch up with @argyleink on UI placement

kosamari:
Please see use case rationale in #950

argyleink:
I'm thinking:
- only show this when zooming is above 100% (nice to have)
- showing an antialiasing toggle to the left of the zoom controls

i'll find an icon and make a design comp soon 👍🏻

argyleink:
![alias-icon](https://user-images.githubusercontent.com/1134620/130819831-ce75ba84-1330-4a76-b51a-24877abdcf5e.png)

Vishalghyv:
Fixed by #1132

PR #1132, opened 2021-08-23
Aliasing toggle button

Vishalghyv:
Feature request from #1087 

As mentioned in https://github.com/GoogleChromeLabs/squoosh/issues/1087#issuecomment-872351367 
 - Button only shown when zooming above 100%
 - Button shown left of zoom

Icon for anti aliasing can take some inspiration from this:https://www.iconfinder.com/icons/365919/antialiasing_digital_filter_quality_icon

PTAL, Thanks


Vishalghyv:
Screen record of the feature
Screencast from 08-24-2021 12 25 28 AM

Vishalghyv:
Devices with less than 300px ex galaxy fold. The button does cover more than required screen space.
Screenshot:
Screenshot(33)

Although, this can be fixed if button is kept with rotate and toggleBackground button

jakearchibald:
@argyleink I know you asked for the button to appear and disappear, but it seems a bit… crap to me. Did you envision it some other way? Also, are you happy with the button placement? I wonder if it's a bit weird it being on another side compared to the other buttons.

src/client/lazy-app/Compress/Output/style.css
```diff
@@ -161,3 +166,9 @@ input.zoom {
     pointer-events: auto;
   }
 }
+
+.pixelated {
+  image-rendering: -webkit-optimize-contrast;
+  image-rendering: -moz-crisp-edges;
+  image-rendering: pixelated;
```
jakearchibald:
```suggestion
  image-rendering: crisp-edges;
  image-rendering: pixelated;
```

Feels better to go with unprefixed versions since they're supported.

jakearchibald (commented):
It seems like Safari has bugs here (it kinda does pixelated, but then interpolates the texture), so we should avoid giving the feature to Safari. Unfortunately I don't know a better way of doing this other than a user-agent detect. https://github.com/jakearchibald/safari-14-idb-fix/blob/main/src/index.ts#L8 might be useful here (put it in the util file).

src/client/lazy-app/util/index.ts
```diff
@@ -295,3 +295,13 @@ export async function abortable<T>(
     }),
   ]);
 }
+
+/**
+ * Checks if render engine is Safari.
+ */
+export function isSafari(): boolean {
```
jakearchibald: This can just be a `const`, since it's the same result every time.

src/client/lazy-app/Compress/Output/index.tsx
```diff
@@ -145,6 +148,14 @@ export default class Output extends Component<Props, State> {
     return props.rightCompressed || (props.source && props.source.preprocessed);
   }
 
+  private toggleAliasing = () => {
+    if (!isSafari()) {
```
jakearchibald: I think it'd be better if the buttons was just hidden for Safari

jakearchibald (changes requested):

Vishalghyv:
Thanks for the review, I have added the change PTAL

jakearchibald:
Looks good! Just need to resolve the stuff in https://github.com/GoogleChromeLabs/squoosh/pull/1132#issuecomment-904626155 and then this is ready to go.

argyleink:
> @argyleink I know you asked for the button to appear and disappear, but it seems a bit… crap to me. Did you envision it some other way? 

I didn't have any alt visions. the original feature request was looking for the button when at 500+ zoom, and subsequent mentions (like for pixel art) still zoom in quite far to verify things. it felt very connected to zooming in, so the thought was tuck it away when not zoomed in so it's not UI noise, conversely make the button appear when it could be useful (zoomed). 

<br>

> Also, are you happy with the button placement? I wonder if it's a bit weird it being on another side compared to the other buttons.

i dont find it that weird. i felt it needed some separation (other side) because, while it's global like the other buttons, it doesnt change the image. but i see the backdrop image flip feature is similar to this effect, as a viewport change. i could see it joining the rotate and backdrop button groups. 

**summary**
not sure jake's design ideas, i think these current 2 items are good as is. buuut, i could see an alternative solution that's just as good tho (which perhaps jake prefers and doesnt get crap vibes from), where the button is with the backdrop button and is always visible. imo, it'll work out either way, the user's flows are unblocked. i was able to zoom and inspect pixel art or inspect artifacts, nice work!

**icon**
attached is an svg icon to use. i think we could flip it horizontally to show on/off. but maybe we want 2 icons, an aliased one and a smooth one, and js flips between them?

alias-icon

```svg
<svg xmlns="http://www.w3.org/2000/svg" width="76" height="102">
  <defs>
    <clipPath id="b">
      <path data-name="Rectangle 106" transform="translate(79 22)" fill="#fff" stroke="#707070" d="M0 0h45v91H0z"/>
    </clipPath>
    <clipPath id="a">
      <path d="M0 0h76v102H0z"/>
    </clipPath>
  </defs>
  <g clip-path="url(#a)">
    <path fill="#fff" d="M0 0h76v102H0z"/>
    <path fill="none" stroke="#000" stroke-linecap="round" stroke-width="5" d="M37.5 2.5v97"/>
    <g transform="translate(-42 -19)" clip-path="url(#b)">
      <g transform="translate(41 32)" fill="none" stroke="#000" stroke-width="5">
        <circle cx="38" cy="38" r="38" stroke="none"/>
        <circle cx="38" cy="38" r="35.5"/>
      </g>
    </g>
    <path d="M70 48h5v5h-5z"/>
    <path d="M29 13h11v5H29z"/>
    <path d="M29 83h11v5H29z"/>
    <path d="M19 14h10v5H19z"/>
    <path d="M19 82h10v5H19z"/>
    <path d="M12 18h9v5h-9z"/>
    <path d="M12 78h9v5h-9z"/>
    <path d="M8 23h7v5H8z"/>
    <path d="M8 73h7v5H8z"/>
    <path d="M5 28h6v5H5z"/>
    <path d="M5 68h6v5H5z"/>
    <path d="M2 33h6v5H2z"/>
    <path d="M2 63h6v5H2z"/>
    <path d="M1 38h5v5H1z"/>
    <path d="M1 58h5v5H1z"/>
    <path d="M0 43h5v15H0z"/>
  </g>
</svg>
```

jakearchibald:
It's mostly the disappearing/appearing I don't like, as it's distracting and causes a layout shift in an unexpected place. Would you be against the button being always there?

Button seems good to me, although we'll see how it looks in place.

argyleink:
> It's mostly the disappearing/appearing I don't like, as it's distracting and causes a layout shift in an unexpected place. Would you be against the button being always there?

not against it nope, mentioned it as a valid solution in my comment "i could see an alternative solution that's just as good tho (which perhaps jake prefers and doesnt get crap vibes from), where the button is with the backdrop button and is always visible"

jakearchibald:
@Vishalghyv did you mean to close this?

Vishalghyv:
Added the changes as mentioned in [comment](https://github.com/GoogleChromeLabs/squoosh/pull/1132#issuecomment-904626155). Currently, second button is the toggle alias one.

I have created two new svg for the icons
With aliasing
final2

Without aliasing
final

Vishalghyv:
@jakearchibald I was reverting a change and by mistake force pushed the commit to wrong head :( Is there a way to fix it and bring back the commits here? Although added the changes back using the commit code

Vishalghyv:
Okay, here is the latest button.
Any changes in the icon like flipping it or changing it have no effect because of the small size of the icon.
Screenshot from 2021-08-26 01-32-39

Please take a look.

argyleink:
can you try with this better icon:
- no bg
- better low res simulation (i hope)
- merged paths (less nodes)
- smaller

```svg
<svg xmlns="http://www.w3.org/2000/svg" width="100" height="100">
  <defs>
    <clipPath id="b">
      <path transform="translate(79 22)" fill="#fff" stroke="#707070" d="M0 0h45v91H0z"/>
    </clipPath>
    <clipPath id="a">
      <path d="M0 0h100v100H0z"/>
    </clipPath>
  </defs>
  <g clip-path="url(#a)">
    <path fill="none" stroke="#000" stroke-linecap="round" stroke-width="5" d="M50 4v92"/>
    <g transform="translate(-29 -22)" clip-path="url(#b)">
      <g transform="translate(41 32)" fill="none" stroke="#000" stroke-width="5">
        <circle cx="38" cy="38" r="38" stroke="none"/>
        <circle cx="38" cy="38" r="35.5"/>
      </g>
    </g>
    <path d="M37 86v-3H26v-8h-8V65h-5V55h9v10h7v8h8v8h11v5zM9 55V41h9v14zm4-14V31h9v10zm5-10V21h8v-8h11v-3h11v5H37v8h-8v8z"/>
  </g>
</svg>
```

Vishalghyv:
This is the latest icon
Screenshot(34)
It's usage in the app
Screenshot(35)

jakearchibald:
@googlebot I consent.

jakearchibald:
How about this? I used a circle the same width as the rotate circle, which toggles to a 9x9 pixel representation:

Screenshot 2021-08-26 at 11 13 34

Screenshot 2021-08-26 at 11 13 43

I also added tooltips for the buttons.

jakearchibald (approved):

jakearchibald:
@argyleink are you happy with this?

argyleink:
> @argyleink are you happy with this?

yep.

icon changes look good 👍🏻 
tooltips were super needed, i was going to comment on that but you got there first 🏆

PR #1136, opened 2021-08-26
JPEG XL: Add lossy Modular option

ziemek99:
This commit adds alternative lossy mode for JPEG XL.

Currently the UI should work, but isn't functional since I don't know how to recompile WASM version of JPEG XL and I'm afraid that doing so would mean me running out of RAM and/or disk space for new toolchain.


surma:
> Currently the UI should work, but isn't functional since I don't know how to recompile WASM version of JPEG XL and I'm afraid that doing so would mean me running out of RAM and/or disk space for new toolchain.

It’s all dockerized. If you have Docker setup, you can just run `npm run build` in `/codecs/jxl` and it should rebuild everything. Don’t quite follow  what you are worried about re disk space or memory.

ziemek99:
> > Currently the UI should work, but isn't functional since I don't know how to recompile WASM version of JPEG XL and I'm afraid that doing so would mean me running out of RAM and/or disk space for new toolchain.
> 
> It’s all dockerized. If you have Docker setup, you can just run `npm run build` in `/codecs/jxl` and it should rebuild everything. Don’t quite follow what you are worried about re disk space or memory.

I've never, ever used Docker in my life. I have to be cautious about what I install (*cries in 3.75 GiB of free disk space*) and what I run (JPEG XL's `./ci.sh release` failed way too many times due to running out of RAM, killing WSL2 window, making me use different build system). I'm trying to install it, but I'm confused because there are so many versions of Docker Engine.

ziemek99:
FINALLY WORKS! @surma PTAL.

(Configuring Docker was PITA though, compiling took an eternity and my free disk space went down to 1.9 GiB. WHYYYYY)

ziemek99:
By the way @jakearchibald, I've been greatly inspired by your PR #1039. I've got a question: why does your commit (8639ccb674b7e8786b70acc540e470b64ed6c6d6) involve just the `codecs/jxl/enc/` directory (as expected), but mine (b81307c2013be4612fb3f2abba0f61534aaee401) involves `codecs/jxl/dec/` as well? I only fiddled with the encoder...

jakearchibald:
Huh, I'm not sure. I didn't do anything special to avoid messing with the decoder. I'll try rebuilding this before we land it and see if I get a different result.

src/features/encoders/jxl/client/index.tsx
```diff
@@ -161,6 +165,13 @@ export class Options extends Component<Props, State> {
                   Quality:
                 </Range>
               </div>
+              <label class={style.optionToggle}>
+                Alternative lossy mode
```
jakearchibald, line 170: Should this be something like "Force modular mode" or similar? We already use modular mode automatically at some qualities.

jakearchibald: Alternatively, it could be a drop down "Mode", with options "auto", "???", "modular" (I'm not sure what the non-modular mode is called).

ziemek99: > Should this be something like "Force modular mode" or similar?

Wording has been intentional. I don't think that anyone outside JPEG XL world knows what "Modular" is, potential users would probably be afraid to click that option. But "Alternative lossy mode"? Sounds easy and encouraging to use, like "wow, there are two lossy modes? Current one isn't doing well, maybe this new one would help".

> We already use modular mode automatically at some qualities.

Yes, quality < 7 forces Modular, which might be confusing to the user ("why this switch doesn't work")? I know about this issue. Maybe resemble encoder's behavior in the UI by making the checkbox (in JS) checked and disabled at those qualities, and after choosing quality >= 7 re-enable the checkbox and set the setting that user picked previously? Is it possible at all? (Another option would be simply hiding the checkbox when the user chooses quality < 7 and reveal it back otherwise.)

> (I'm not sure what the non-modular mode is called).

It's VarDCT. Almost the same DCT like in JPEG, but "Var" because of variable block sizes, not just JPEG's 8x8.

jakearchibald: Is there any benefit to allowing VarDCT at qualities < 7?

jakearchibald: I'm thinking we could have a `<select>` with the options:

- Auto lossy
- Main lossy (VarDCT)
- Alternate lossy (Modular)
- Lossless

Where "auto lossy" is what we currently do for lossy, and "lossless" is what we currently do for lossless. This would replace the lossless checkbox.

ziemek99: > Another option would be simply hiding the checkbox when the user chooses quality < 7 and reveal it back otherwise.

I've just implemented this because it was easier. This solution:
- allows user-friendly caption to remain,
- eliminated the need for a complete redesign of lossy/lossless choice (that'd be hard to do _while remaining user-friendly_),
- preserves checkbox's state when going in and out of quality < 7 range (nice feature).

If you've got any feedback on my solution, please make a comment on the new commit since this code became outdated (or feel free to unresolve this conversation if you prefer to do so).

ziemek99:
I've just added another commit because I guess I'm not the only one that finds JS glue code diffs annoying, especially in Git CLI...
For those who *really* want to view such a diff, GitHub will still generate it in a browser.

By the way I kindly ask to use a "normal" merge strategy ("Create a merge commit"), without squashing, since those changes are a bit related, but still separate and I'd like it to be preserved in commit history. Of course if you prefer changing .gitattributes to be in a different PR, I'll do it and rebase that commit out of this PR.

jakearchibald:
> By the way I kindly ask to use a "normal" merge strategy ("Create a merge commit"), without squashing, since those changes are a bit related, but still separate and I'd like it to be preserved in commit history.

Would you be happy with a rebase + fastforward?

Also, what did you think about the `<select>` with the options:

- Auto lossy
- Main lossy (VarDCT)
- Alternate lossy (Modular)
- Lossless

Where "auto lossy" is what we currently do for lossy, and "lossless" is what we currently do for lossless. This would replace the lossless checkbox.

This is consistent with what we did for the "Tuning" option in AVIF, where 'auto' picks an option depending on the quality setting, but it can be overridden to force PSNR or SSIM.

Right now it seems difficult to explain why the modular checkbox appears and disappears at particular quality numbers.

ziemek99:
> > By the way I kindly ask to use a "normal" merge strategy ("Create a merge commit"), without squashing, since those changes are a bit related, but still separate and I'd like it to be preserved in commit history.
> 
> Would you be happy with a rebase + fastforward?

I'd like to keep GPG signatures, but if you prefer not to clutter the history with merge commits (why can't GitHub just support fast-forward merges?!), which I completely understand, feel free to rebase. (BTW: I've just checked in my toy repo ziemek99/test#5 and apparently you can do fast-forward merges by doing it in Git CLI and GitHub will understand this, closing the PR. It's quite clunky, though. Definitely not recommended if you've got plenty of PRs... Just saying.)

> Also, what did you think about the `<select>` with the options:
> 
> * Auto lossy
> * Main lossy (VarDCT)
> * Alternate lossy (Modular)
> * Lossless
> 
> Where "auto lossy" is what we currently do for lossy, and "lossless" is what we currently do for lossless. This would replace the lossless checkbox.
>
> This is consistent with what we did for the "Tuning" option in AVIF, where 'auto' picks an option depending on the quality setting, but it can be overridden to force PSNR or SSIM.

I don't think it's consistent with other codecs. Every other one has a checkbox for choosing lossless, not a dropdown. It'd unnecessarily complicate things and overwhelm users, especially non-JXL people. Another thing is that switching between lossy modes won't be as easy, because a dropdown requires more clicks, screen estate and attention compared to a checkbox. And is there any way to copy the quality slider value between those modes? What if user picks quality 1 in Modular, switches to VarDCT (where 7 is the lowest possible value) and then back to Modular? What's supposed to happen?

> Right now it seems difficult to explain why the modular checkbox appears and disappears at particular quality numbers.

I understand your concerns, indeed it might be a bit confusing to users (maybe a bit less than a `<select>` with 3 lossy modes, but still). But remember my other proposal for solving this problem:

> Maybe resemble encoder's behavior in the UI by making the checkbox (in JS) checked and disabled at those qualities, and after choosing quality >= 7 re-enable the checkbox and set the setting that user picked previously?

What do you think about that? It might be harder to code, but way more suggestive to the user about what happens behind the scenes.

ziemek99:
OK, you should probably get the idea. Now I only need to make the checkbox look disabled under quality < 7, but that's gonna be tricky. (UPDATE: it was easier than I expected, it works now!) Another thing: maybe should I hide the noise synth slider since it doesn't work with Modular yet (but it might in the future since the JXL bistream doesn't forbid it)?

ziemek99:
I'm reworking this thing a bit, hang on 🙂

jakearchibald:
Give me a shout when this is ready for review again.

ziemek99:
Merging libjxl/libjxl#868 fixed the blocking issue with noise not working in Modular, but now there are weird things happening when I tried to re-build `codecs/jxl/` based on the latest git: libjxl/libjxl@44135538027732cb5824fea28ab29ede79862d09. It's complaining about something related to skcms...
```
Building CXX object lib/CMakeFiles/jxl_dec-obj.dir/jxl/toc.cc.o
/emsdk/upstream/emscripten/em++ \pm:load:cleanup
        -O3 -flto -std=c++17 -Wno-deprecated-declarations -pthread \
        -O3 -flto -s FILESYSTEM=0 -s PTHREAD_POOL_SIZE=navigator.hardwareConcurrency -s ALLOW_MEMORY_GROWTH=1 -s TEXTDECODER=2 -s NODEJS_CATCH_EXIT=0 -s NODEJS_CATCH_REJECTION=0  \
        -I node_modules/jxl \
        -I node_modules/jxl/lib \
        -I node_modules/jxl/lib/include \
        -I node_modules/jxl/build/mt-simd/lib/include \
        -I node_modules/jxl/third_party/highway \
        -I node_modules/jxl/third_party/skcms \
        --bind \
        -s ENVIRONMENT=worker \
        -s EXPORT_ES6=1 \
        -o enc/jxl_enc_mt_simd.js \
        enc/jxl_enc.cpp node_modules/jxl/build/mt-simd/lib/libjxl.a node_modules/jxl/build/mt-simd/lib/libjxl_threads.a \
        node_modules/jxl/build/mt-simd/third_party/brotli/libbrotlidec-static.a \
        node_modules/jxl/build/mt-simd/third_party/brotli/libbrotlienc-static.a \
        node_modules/jxl/build/mt-simd/third_party/brotli/libbrotlicommon-static.a \
        node_modules/jxl/build/mt-simd/third_party/libskcms.a \
        node_modules/jxl/build/mt-simd/third_party/highway/libhwy.a
em++: error: node_modules/jxl/build/mt-simd/third_party/libskcms.a: No such file or directory ("node_modules/jxl/build/mt-simd/third_party/libskcms.a" was expected to be an input file, based on the commandline arguments provided)
make: *** [Makefile:39: enc/jxl_enc_mt_simd.js] Error 1
make: *** Waiting for unfinished jobs....
```
That was a clean compile, with the whole Squoosh repo removed completely from the drive and re-cloned from the remote, hence the recent push (in order not to loose my work). Surprisingly enough libjxl/libjxl@a205468bc5d3a353fb15dae2398a101dff52f2d3 (v0.6.1) worked...

ziemek99:
@CanadianBaconBoi THAAAAANK YOUUUUUUU!!! 😃

ziemek99:
> Give me a shout when this is ready for review again.

@jakearchibald It's time. 🙂

jakearchibald (approved):
Brilliant! Thanks for all the work on this

jakearchibald:
Deploying this change to live

Issue #1065, opened 2021-06-19
Node.js - abort(RangeError: WebAssembly.instantiate(): Out of memory: wasm memory)


Tchoupinax:
Hello 👋 

I met this error when I try to process serveral image in parallel : 

```js
ERROR [15:10:28.879]: abort(RangeError: WebAssembly.instantiate(): Out of memory: wasm memory). Build with -s ASSERTIONS=1 for more info.
    RuntimeError: abort(RangeError: WebAssembly.instantiate(): Out of memory: wasm memory). Build with -s ASSERTIONS=1 for more info.
        at abort (/app/node_modules/@squoosh/lib/build/index.js:24:9428)
        at /app/node_modules/@squoosh/lib/build/index.js:24:11293
failed to asynchronously prepare wasm: RangeError: WebAssembly.instantiate(): Out of memory: wasm memory
```

I used this code

```js
const imagePool = new ImagePool();

const image = imagePool.ingestImage(`./public/${req.params.s3Name}`);
await image.decoded;

const preprocessOptions = {
  resize: {
    // method:'lanczos3',
    enabled: true,
    height: 300,
  },
};
// eslint-disable-next-line no-console
console.time(`preprocess ./public/${req.params.s3Name}`);
await image.preprocess(preprocessOptions);
// eslint-disable-next-line no-console
console.timeEnd(`preprocess ./public/${req.params.s3Name}`);

const encodeOptions = {
  mozjpeg: {
    quality: 50,
  }, //an empty object means 'use default settings'
};

// eslint-disable-next-line no-console
console.time(`encode ./public/${req.params.s3Name}`);
await image.encode(encodeOptions);
// eslint-disable-next-line no-console
console.timeEnd(`encode ./public/${req.params.s3Name}`);
imagePool.close();

const extension = req.params.s3Name.split('.').pop();
res.set('Content-Type', `image/${extension}`);

const rawEncodedImage = (await image.encodedWith.mozjpeg).binary;

fs.writeFileSync(`./public-light/${req.params.s3Name}`, rawEncodedImage);
```

Several information about this issue : 
- although this issue is thrown, once I have catch it, the image seems to be processed anyway.
- although this code was under a classic `try/catch`, the error crashed Node.js process. I had to use this code to really catch the error : 
```js
process.on('uncaughtException', (err) => {
  logger.error(err);
});
```
Then, with this code, the process does not crash anymore and seems to continue its operation.

[...]

PR #1171, opened 2021-10-20
Add warning for using multiple image pools

TimvdLippe:
Instantiating and using multiple image pools is not intended and can
lead to memory problems. Instead, users should use a single pool and
reuse that pool across all image processing. Therefore, let's add
a warning in the README to specify call this out and avoid users
of running into issues.

Relates to #1065


surma (approved):

ergunsh:
Hey hey, I saw this and wanted to suggest that maybe we can enforce this on the API level or add a warning log for the users that used it (with the option of suppressing the log if they really know what they are doing). Does it make sense? :)

PR #1172, opened 2021-10-20
Tidy up CLI implementation

TimvdLippe:
A couple of issues while working on it:
1. The `processFiles` function is `async`. If it would throw an error,
then `commander` wouldn't properly catch the error, which results in
an uncaught promise rejection. Instead, we should be using `parseAsync`,
which can handle an async action
2. If the user would specify the encoder, but without encoding configuration,
then the CLI would crash. Now, double check that we the user passes in
a string and, if not, throw a nicer error.
3. For some reason, creating the `ora` spinner in the `progressTracker`
function results in leaked listeners when listening on `stdout`. Presumably
that's because `commander` already listens to it, but I don't know for sure.
If we instead call `ora` in the module scope, it can attach to the proper
outputstream and ensure that the leak is prevented.

### DIFF
--- a/cli/src/index.js
+++ b/cli/src/index.js
@@ -10,6 +10,8 @@ import kleur from 'kleur';
 
 import { ImagePool, preprocessors, encoders } from '@squoosh/lib';
 
+const spinner = ora();
+
 function clamp(v, min, max) {
   if (v < min) return min;
   if (v > max) return max;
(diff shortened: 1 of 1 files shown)


cli/src/index.js
```diff
@@ -233,4 +237,4 @@ for (const [key, value] of Object.entries(encoders)) {
   );
 }
 
-program.parse(process.argv);
+await program.parseAsync(process.argv);
```
TimvdLippe, line 240: Note that we are using top level await here, which is required to prevent uncaught promise rejection errors. Currently, the CLI specifies `^12.20.2` which I don't think ships it, but the other two listed versions do support it.

Issue #1152, opened 2021-09-11
libSquoosh declares a global `navigator` variable


aduh95:
**Before you start**
Please take a look at the [FAQ](https://github.com/GoogleChromeLabs/squoosh/wiki/FAQ) as well as the already opened issues! If nothing fits your problem, go ahead and fill out the following template:

**Describe the bug**

`import "@squoosh/lib"` add a global `navigator` variable in `codecs.ts`. The issue is that some libraries assume the code is running is a web browser environment if there's a `navigator` object available on the global object.

**To Reproduce**

```sh
mkdir repro
cd repro
npm i sass @squoosh/lib
echo 'import "@squoosh/lib"; import "sass";' > test.mjs
node ./test.mjs
```

**Expected behavior**

`test.mjs` should exit with code 0.

**Actual behavior**

```
…/repro/node_modules/sass/sass.dart.js:92529
    if (ua.indexOf("DumpRenderTree") >= 0) return hooks;
           ^

TypeError: Cannot read properties of undefined (reading 'indexOf')
    at …/repro/node_modules/sass/sass.dart.js:92529:12
    at Object.applyHooksTransformer (…/repro/node_modules/sass/sass.dart.js:7236:14)
    at Object.initHooks (…/repro/node_modules/sass/sass.dart.js:7216:251)
    at Object.initNativeDispatchContinue (…/repro/node_modules/sass/sass.dart.js:7182:9)
    at Object.initNativeDispatch (…/repro/node_modules/sass/sass.dart.js:7176:9)
    at Object.getNativeInterceptor (…/repro/node_modules/sass/sass.dart.js:9070:13)
    at Object.getInterceptor$x (…/repro/node_modules/sass/sass.dart.js:9326:16)
    at Object.set$render$x (…/repro/node_modules/sass/sass.dart.js:9363:16)
    at main0 (…/repro/node_modules/sass/sass.dart.js:17757:9)
    at …/repro/node_modules/sass/sass.dart.js:93924:7
```

**Version:**
 - OS w/ version: [e.g. iOS 12] N/A
 - Browser w/ version [e.g. Chrome 70] N/A
 - Node version: [e.g. 10.11.0] 16.9.1
 - npm version: [e.g. 6.4.1] 7.21.1

**Additional context, screenshots, screencasts**

[...]

PR #1174, opened 2021-10-21
Add script to regenerate all codecs

TimvdLippe:
The `build-all.sh` script runs `npm run build` for all codecs.
Additionally, recompile all codecs using the script with the current
state of the repository. This is in preparation of making some build
changes for #1152


codecs/build-all.sh
```diff
@@ -0,0 +1,7 @@
+CURRENT_DIRECTORY=$(dirname "$(readlink -f "$0")")
+
+for codec in "avif" "hqx" "imagequant" "jxl" "mozjpeg" "oxipng" "png" "resize" "rotate" "visdif" "webp" "wp2"
```
surma: Any chance we can just go into each directory? If someone adds a codec, they’ll 100% forget to add it to this list.

surma (commented):
Bit surprised that the Wasm binaries changed. They should have been identical as far as I can tell. Any idea why?

also cc @RReverser

codecs/build-all.sh
TimvdLippe: Yup, turns out we can do that :smile:

TimvdLippe:
Unfortunately, I don't know why they have changed. There is also the extra `.d.ts` file which was missing before maybe?

I did build it locally and ran through all codecs to check if they were still working, which they were. So I suppose it is fine? :sweat_smile:

TimvdLippe:
It seems like the MakeFiles don't recompile sometimes. This means that you manually have to delete files to force a rebuild, which then shows the new result. Maybe we should force clean the output files whenever we build, to ensure we always build fresh?

surma:
The whole point of Makefiles is to only rebuild files that have changed. We do have `make clean` in all Makefiles to force a rebuild tho. I am not sure we should do that _by default_ from the script, but a parameter/env to force a rebuild seems sensible.

TimvdLippe:
I did attempt running `make clean`, but that in itself was throwing errors complaining about missing files or something. I would have expected Make to figure out that the environment variables had changed, but maybe these environment variables are missing? E.g. make detects that the scripts nor the files have changed, so it bails out. But it accidentally overlooks the fact that these environment variables can affect the build?

PR #1176, opened 2021-10-21
Use automatic detection of environment for hardware concurrency

### LINKED ISSUE #1152: libSquoosh declares a global `navigator` variable

TimvdLippe:
We can pass in any string into `PTHREAD_POOL_SIZE`, which means we can
inject a bit of code to implement the switch-case for both Node and web.
On the web, it will use `navigator.hardwareConcurrency` as before, while
on Node it will use `cpus().length`. This allows us to remove the
assignment to `navigator` on the global scope in `codecs.ts`, which
avoids throwing off tools like `sass` which assume a web environment
when that global is set.

Fixes #1152
Builds upon #1174


codecs/cpp.Dockerfile
```diff
@@ -4,7 +4,7 @@ ENV CFLAGS "-O3 -flto"
 ENV CXXFLAGS "${CFLAGS} -std=c++17"
 ENV LDFLAGS "${CFLAGS} \
 -s FILESYSTEM=0 \
--s PTHREAD_POOL_SIZE=navigator.hardwareConcurrency \
+-s PTHREAD_POOL_SIZE=\"ENVIRONMENT_IS_NODE?require('os').cpus().length:navigator.hardwareConcurrency\" \
```
TimvdLippe, line 7: Note that I tried to keep this with spaces, but unfortunately appeasing both Bash and Docker didn't seem to work out. Instead, we can remove the spaces (which are only visual anyways) and avoid these spacing issues in the environment argument.

RReverser: Huh weird - I'd expect it to handle spaces with escaped quotes like any other - but okay.

codecs/build-all.sh
```diff
@@ -0,0 +1,8 @@
+CURRENT_DIRECTORY=$(dirname "$(readlink -f "$0")")
+SUB_DIRECTORIES=$(find "$CURRENT_DIRECTORY" -mindepth 1 -maxdepth 1 -type d)
```
RReverser, line 2: Nit: might be simpler to use `ls -d $CURRENT_DIRECTORY/*/`.

TimvdLippe: Changed in #1174

RReverser:
Hm not sure why we didn't merge this. @surma Thoughts?

donmccurdy:
This would help me as well – I currently can't use `@squoosh/lib` alongside another package that does something like:

```js
if (typeof navigator === 'undefined') return;
// proceeds to assume Web environment ...
```

donmccurdy:
@RReverser @jakearchibald is there anything I could do to help get this PR merged? E.g. rebasing it or testing some specific use cases? I'd like to use `@squoosh/lib` in a project but the modification of the global namespace (below) is breaking other dependencies that use `navigator` to detect Web vs. Browser contexts.

```js
// breaks other project dependencies
(globalThis as any).navigator = {
  hardwareConcurrency: cpus().length,
};
```

burner03:
I just run 

`navigator = '';` 

at the top of my automation. Not a great solution. Would love to see this merged.

donmccurdy:
For anyone else blocked – I had another package breaking on the navigator shim. My workaround was similar to what @burner03 describes, but I had to replace async 'import' with synchronous 'require' to ensure the fix was in place at the right time. This got a little trickier with TypeScript, so here's the full workaround:

before:

```js
import other from 'other-package';
```

after:

```typescript
import type Other from 'other-package';

const other = createOther();

/** Creates and returns a 'other' instance. */
function createOther(): typeof Other {
	// Workaround for https://github.com/GoogleChromeLabs/squoosh/pull/1176.
	const navigator = global.navigator;
	delete (global as Record<string, unknown>).navigator;
	try {
		return require('other-package');
	} finally {
		global.navigator = navigator;
	}
}
```

Issue #1184, opened 2021-11-03
Add support for Node 17


joshunger:
**Describe the bug**
Add support for Node 17

**To Reproduce**
```
nvm use 17
yarn global add @squoosh/cli
yarn global v1.22.10
[1/4] 🔍  Resolving packages...
[2/4] 🚚  Fetching packages...
error @squoosh/lib@0.4.0: The engine "node" is incompatible with this module. Expected version " ^12.5.0 || ^14.0.0 || ^16.0.0 ". Got "17.0.1"
error Found incompatible module.
info Visit https://yarnpkg.com/en/docs/cli/global for documentation about this command.
```

**Expected behavior**
Should be able to install with node 17.

jakearchibald:
Unfortunately, due to a few people leaving the team, and staffing issues resulting from the current economic climate (ugh), the CLI and libsquoosh packages are no longer actively maintained. I know that sucks, but there simply isn't the time & people to work on this. If anyone from the community wants to fork it, you have my blessing. The [squoosh.app](https://squoosh.app/) web app will continue to be supported and improved.

PR #1194, opened 2021-12-01
fix: support Node.js v17

alexander-akait:


alexander-akait:
@surma I don't link pinging, sorry, but we really need that

atjn:
Closes #1184

ghost:
I believe this requires a change of package-lock.json as well.

PR #1252, opened 2022-06-07
Add option for easy self hosting with Docker

firstdorsal:
Check out the example hosted at: https://missing-surma-at-http203.y.gy/

Image is ~25MB in size.

Is the name **squoosh** subject to trademarks mentioned in the [license](https://github.com/GoogleChromeLabs/squoosh/blob/dev/LICENSE)?
Could someone host it at their subdomain like `squoosh.example.com`?

Someone could also create an official docker image at dockerhub or similar.


NaturalDevCR:
It would be awesome to have the option to self host it, sadly this never made it.

Issue #1064, opened 2021-06-19
Center Image after typing in the zoom value


nathnolt:
**Is your feature request related to a problem? Please describe.**
When you change the zoom value by typing something in (using the keyboard), the image moves (pans) off of the screen (out of view). (right bottom)

**Describe the solution you'd like**
If the image would get centered (panned to the center of the screen) when the zoom value is changed, by typing, or even better: if the mouse is over the image when the typing happens, zoom in like scroll zoom, otherwise, pan image to middle.

**Does other service/app have this feature?**
I don't know. but gimp doesn't even allow the image to be panned out of view more than 50% of the image.

**Additional context**
I feel like a static image of Squoosh where the image is not visible is kinda useless.

surma:
Before I start: The image is always clamped to the border of the viewport, so while it might be invisible, it is _at worst_ just outside so panning into all directions a bit should get the image back into view.

That being said, it is bad UX what’s happening right now. I think the main mistake here is updating the view _while_ the user is typing. So when a user wants to type in “200”, after pressing “2” the view updated to a 2% zoom of the image. This (for some reason) puts the image off-center so that the subsequent “0”s move the image out of frame. 

Maybe it makes sense that for keyboard input on the zoom field we switch to `change` event (or Enter)?

nathnolt:
I feel like a debounced ```input``` would be more responsive, but either ```change``` and listening to enter would also work.

NoamRa:
Started digging into this and it doesn't look like the problem is with how `setTransform`. 
* [This matrix](https://github.com/GoogleChromeLabs/squoosh/blob/d526877147990cf71e4dc4c0ff02797e034f586e/src/client/lazy-app/Compress/Output/custom-els/PinchZoom/index.ts#L208) contains very low negative numbers for x and y values ( matrix's content example: `a: 9.99999985098839, b: 0 ,
c: 0 , d: 9.99999985098839 , e: -3177.0001610219483 , f: -6489.000045150519` ) and when transforming `topLeft` and `bottomRight`, throws the image out of the viewport.
* The part that's responsible to making sure the image is not out of bounds does not work as expected.

Rajkaran-122:
@kosamari @surma sir please review this pr #1498 .

PR #1256, opened 2022-07-02
Solves: center image after typing zoom value

eduebernal:
Solves: #1064

Quick demo: 
https://user-images.githubusercontent.com/92958867/177018349-8ecad6a0-0739-4aeb-9d7a-97acce684ae5.mov

Changed the trigger for the scaleTo function so that it is called onChange, rather than onInput. Doing this, the scaling is applied if you press enter, if you use the arrow keys or if you use the UI buttons. I think this option is good, but debouncing could also be implemented as suggested in the issue. Let me know what you think.

Also added a part to recenter the image after scaling it. Before, if you changed the scaling manually by typing and the difference in scaling was big, the image would go out of bounds. Now it recenters every time you scale it by typing or using the UI buttons.

Tested it with different dimensions for the viewport.

Let me know if you have any suggestions or concerns!

### DIFF
--- a/src/client/lazy-app/Compress/Output/custom-els/PinchZoom/index.ts
+++ b/src/client/lazy-app/Compress/Output/custom-els/PinchZoom/index.ts
@@ -173,11 +173,27 @@ export default class PinchZoom extends HTMLElement {
       originY -= currentRect.top;
     }
 
+    //Apply scaling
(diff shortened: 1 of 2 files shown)


IOLOII (commented):
Success test localhost

PR #1278, opened 2022-08-19
build(libsquoosh): add `del` to devDependencies

Asuka109:
`libsquoosh/lib/move-d-ts.js` will try to import `del`, but it doesn't exist. So just install it


ziemek99:

PR #1321, opened 2023-01-03
Remove CLI / libsquoosh

jakearchibald:
Unfortunately, due to a few people leaving the team, and staffing issues resulting from the current economic climate (ugh), I'm deprecating the CLI and libsquoosh parts of Squoosh. The web app will continue to be supported and improved. I know that sucks, but there simply isn't the time & people to work on this. If anyone from the community wants to fork it, you have my blessing.


fullheart:
Here an active fork of the CLI: https://github.com/frostoven/Squoosh-with-CLI

PR #1335, opened 2023-03-01
Rotate copy-over-buttons correctly

aryanpingle:
# Minor UI improvement

This PR rotates the `copy-over-button` elements in the correct direction to point towards the other `options` element. <br>It's especially confusing on mobile, where there's nothing on either side of the `options` element so you don't know what the button is supposed to do.

It's only a subtle difference on desktop but the issue still remains (**top right of each `options` element**):

### Comparison (Desktop)

| Before | After |
|---|---|
| localhost_8800_ (1) | localhost_8800_editor |

### Comparison (Mobile)

| Before | After |
|---|---|
| localhost_8800_(Pixel 5) | localhost_8800_(Pixel 5) (1) |

### DIFF
--- a/src/client/lazy-app/Compress/Options/style.css
+++ b/src/client/lazy-app/Compress/Options/style.css
@@ -112,6 +112,9 @@
 .copy-over-button {
   composes: title-button;
 
+  /* Make the filled arrow point towards the other options element */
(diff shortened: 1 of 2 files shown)


jakearchibald (approved):
Hah, this is a nice tweak, thank you!

PR #1348, opened 2023-03-22
Better Keyboard A11y + Navigation

aryanpingle:
Keyboard accessibility is super important, and not just for the visually impaired. It's common practice to `TAB` through a website when your mouse is faulty and/or you can't be bothered to reach for it. The options components  aren't very keyboard-accessible; this PR aims to change that.

Changes include:
* Making all input types react to keyboard focus
* Minor design improvements in this scope

## Visual Comparison

In both before and after I `TAB` to the `Advanced Settings` expander, and then `TAB` to the end. If it feels like I'm not doing anything in the _before changes_ video... well, that's the issue. I also try to make the checkboxes feel like Material UI, but I can only get so far without a complete overhaul.

[Before changes - Little to no focus effects](https://imgur.com/a/eXLFNsR)
[After changes - 100% support for keyboard navigation](https://imgur.com/a/6TFwupl)

### DIFF
--- a/src/client/lazy-app/Compress/Options/style.css
+++ b/src/client/lazy-app/Compress/Options/style.css
@@ -83,11 +83,11 @@
 }
 
 .text-field {
-  background: var(--white);
-  color: var(--black);
+  background-color: var(--black);
+  color: var(--white);
   font: inherit;
   border: none;
-  padding: 6px 0 6px 10px;
+  padding: 6px 6px 6px 10px;
   width: 100%;
   box-sizing: border-box;
   border-radius: 4px;
--- a/src/client/lazy-app/Compress/Options/style.css
+++ b/src/client/lazy-app/Compress/Options/style.css
@@ -53,6 +53,16 @@
   composes: option-toggle;
   grid-template-columns: auto 1fr;
   gap: 1em;
+
+  border-top: 1px solid #fff4;
+
+  transition-property: background-color;
+  transition-duration: 250ms;
+}
+
+.option-reveal:focus-within,
+.option-reveal:hover {
+  background-color: #fff2;
 }
 
 .option-one-cell {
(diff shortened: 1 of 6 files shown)


aryanpingle:
Hi @argyleink, do the changes look good to you? Before/After comparisons are in the description, and I've made the branch up-to-date with `dev` :)

argyleink:
sorry for the delay here, I got super sick last week and was out for a few days. I'm ramping back up this week and have this on my todo list 👍🏻

aryanpingle:
@argyleink Ahh no worries at all, hope you're feeling better now 😄

src/client/lazy-app/Compress/Options/style.css
```diff
@@ -118,6 +128,11 @@
   svg {
     fill: var(--header-text-color);
   }
+
+  &:focus {
```
argyleink, line 132: can you also add this new style to the new buttons (merged last week) `.save-button, .import-button`? a little akward atm that one has it and the other doesnt.

Screenshot 2023-04-11 at 12 23 12 PM

argyleink (changes requested):
this is a significant improvement to the keyboard experience, thanks so  much! 

I've got 2 small requests 🙂 
1. I left one request inline with some code, there's 2 new buttons that could use the same focus treatment.
2. Could you also add a focus state to the close button? maybe it just goes opaque? or an outline works great too.

src/client/lazy-app/Compress/Options/style.css
aryanpingle: Ahh nice catch, didn't see the new icons. I'll apply the same style to them.

aryanpingle:
@argyleink I added some new changes, would love to hear your thoughts :)

1. I added the focus outline to the save/import buttons, and made them conditionally disabled (instead of just reducing the opacity)
2. The back button now has a simple blue pulse outline animation on focus. Nothing too fancy, just something to make it stand out against the light/dark backgrounds (image below)

argyleink (approved):
looks great!
so much better, ty for this 🙏🏻

PR #1358, opened 2023-04-04
Fix : Sample File Size Unit

### LINKED ISSUE #1192: Wrong units on the main website
**Describe the bug**
image

I doubt You wanted to show the file sizes in milibits (mb). After checking the size of these assets I confirmed the proper unit would be Megabyte (MB)
Also, the last file's size does not even have a unit - it's just 13k (k what?) - and the size is actually 10.7 kB, not 13.

Taking this a step further I would recommend using [IEC prefixes](https://en.wikipedia.org/wiki/Binary_prefix#mebi) - eg. Mebibytes - instead of the SI ones.

harsh26shah03:
Sample data had size label showing wrong memory units

13k instead of 13KB
2.8mb instead of 2.8MB

small b corresponds to bits and this would change entire meaning of file so fixed it

Closes #1192

### DIFF
--- a/src/shared/prerendered-app/Intro/index.tsx
+++ b/src/shared/prerendered-app/Intro/index.tsx
@@ -24,28 +24,28 @@ import SlideOnScroll from './SlideOnScroll';
 const demos = [
   {
     description: 'Large photo',
-    size: '2.8mb',
+    size: '2.8MB',
     filename: 'photo.jpg',
     url: largePhoto,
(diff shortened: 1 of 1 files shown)


harsh26shah03:
Closes #1192

harsh26shah03:
@argyleink ready for review

argyleink (approved):
ty for this! lgtm 👍🏻

PR #1361, opened 2023-04-08
implemented scrolling shadows

### LINKED ISSUE #218: It isn't clear when options can be scrolled
I want to experiment with one of those scrolling highlight/shadows. But it feels like other things have priority right now.

tsukinoko-kun:
As @argyleink suggested in issue #218 I tried to implement the scrolling shadows using a gradient background.

Like the one explained here: https://lea.verou.me/2012/04/background-attachment-local

A dark shadow is hardly noticeable so I went for something brighter.

I removed the `options-section` background to make the scrolling shadows visible. It doesn't look great, since the other components are obviously displayed on top of the scrolling shadows.

Bildschirmfoto 2023-04-08 um 17 05 42

Do you have any ideas on how to fix that? Or is this acceptable?

closes #218

### DIFF
--- a/src/client/lazy-app/Compress/Options/style.css
+++ b/src/client/lazy-app/Compress/Options/style.css
@@ -3,6 +3,32 @@
   border-radius: var(--scroller-radius);
   overflow: hidden;
 
+  /* Visualize that the menu is scrollable */
+  background: linear-gradient(var(--off-black) 30%, rgba(255, 255, 255, 0)),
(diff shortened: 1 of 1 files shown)


argyleink:
i tested out your PR and also tried 2 other techniques for providing a "scroll hint" and none of them felt viable. i'm currently thinking we can close #218, it's obviously not been a huge blocker and there's no clean solutions. the best solution would be if we knew the div was currently overflowing and could apply conditional styles, but that's a style query that's not specced or available yet.

PR #1366, opened 2023-06-01
Progressively add descriptions for codec settings

aryanpingle:
Every codec has a ton of settings to mess around with, but the average user won't know what half of them mean. I'm adding small, slightly technical descriptions for each parameter in their respective component's `title` attribute so that users can hover over it and read about its purpose.

All changes are progressive, so descriptions can be added on an *as-needed* basis.

- [ ] AVIF
- [X] ~Browser JPEG~
- [X] ~Browser PNG~
- [X] MozJPEG
- [ ] OxiPNG
- [ ] WebP

| Example |
|:-:|
| image |

### DIFF
--- a/src/features/encoders/mozJPEG/client/index.tsx
+++ b/src/features/encoders/mozJPEG/client/index.tsx
@@ -1,4 +1,8 @@
-import { EncodeOptions, MozJpegColorSpace } from '../shared/meta';
+import {
+  EncodeOptions,
+  MozJpegColorSpace,
+  getParameterDescription,
+} from '../shared/meta';
 import type WorkerBridge from 'client/lazy-app/worker-bridge';
 import { h, Component } from 'preact';
 import {
@@ -106,7 +110,10 @@ export class Options extends Component<Props, State> {
     // gathering the data.
     return (
       <form class={style.optionsSection} onSubmit={preventDefault}>
-        <div class={style.optionOneCell}>
+        <div
+          class={style.optionOneCell}
+          title={getParameterDescription('quality')}
+        >
           <Range
             name="quality"
             min="0"
@@ -127,7 +134,10 @@ export class Options extends Component<Props, State> {
         <Expander>
           {showAdvanced ? (
             <div>
-              <label class={style.optionTextFirst}>
+              <label
+                class={style.optionTextFirst}
+                title={getParameterDescription('color_space')}
+              >
                 Channels:
                 <Select
                   name="color_space"
@@ -142,7 +152,10 @@ export class Options extends Component<Props, State> {
               <Expander>
                 {options.color_space === MozJpegColorSpace.YCbCr ? (
                   <div>
-                    <label class={style.optionToggle}>
+                    <label
+                      class={style.optionToggle}
+                      title={getParameterDescription('auto_subsample')}
+                    >
                       Auto subsample chroma
                       <Checkbox
                         name="auto_subsample"
@@ -152,7 +165,10 @@ export class Options extends Component<Props, State> {
                     </label>
                     <Expander>
                       {options.auto_subsample ? null : (
-                        <div class={style.optionOneCell}>
+                        <div
+                          class={style.optionOneCell}
+                          title={getParameterDescription('chroma_subsample')}
+                        >
                           <Range
                             name="chroma_subsample"
                             min="1"
@@ -165,7 +181,10 @@ export class Options extends Component<Props, State> {
                         </div>
                       )}
                     </Expander>
-                    <label class={style.optionToggle}>
+                    <label
+                      class={style.optionToggle}
+                      title={getParameterDescription('separate_chroma_quality')}
+                    >
                       Separate chroma quality
                       <Checkbox
                         name="separate_chroma_quality"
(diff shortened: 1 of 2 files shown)


src/features/encoders/mozJPEG/shared/meta.ts
```diff
@@ -37,3 +37,67 @@ export const defaultOptions: EncodeOptions = {
...
+        balancedText += '\n' + word + ' ';
+        runningLength = word.length;
+      } else {
+        runningLength += word.length + 1;
+        balancedText += word + ' ';
+      }
+    }
+  }
+  return balancedText.trim();
+}
```
surma, line 62: This function is for the UI and should probably in `src/client/lazy-app/util` or something. Rather than applying the text transform here, the UI should probably do it on render.

surma: I also think we can simplify this. The algorithm here should probably be something like

- Split text into arrays of lines
- For each line:
  - While current output line is shorter than `limit`
    - Add the next word from the current line to the current output line 
  - Emit current output line
  - Set output line to empty string.

(This could be a great use-case for a generator)

src/features/encoders/mozJPEG/shared/meta.ts
```diff
@@ -37,3 +37,67 @@ export const defaultOptions: EncodeOptions = {
...
+
+// Until we make a better way of displaying information (like with modals), displaying a simple 'title' attribute should suffice for each codec parameter
+type mozjpeg_param = keyof typeof defaultOptions;
+const paramDescriptions: Record<mozjpeg_param, string> = {
+  quality:
+    'Trade-off between compressed file size and image quality\nHigher quality means larger file sizes',
```
surma, line 68: If you use template literals and `dedent`, we can make this more readable:

```suggestion
  quality: dedent(`
    Trade-off between compressed file size and image quality.
    Higher quality means larger file sizes
  `),
```

src/features/encoders/mozJPEG/shared/meta.ts
```diff
@@ -37,3 +37,67 @@ export const defaultOptions: EncodeOptions = {
...
+        balancedText += word + ' ';
+      }
+    }
+  }
+  return balancedText.trim();
+}
+
+// Until we make a better way of displaying information (like with modals), displaying a simple 'title' attribute should suffice for each codec parameter
+type mozjpeg_param = keyof typeof defaultOptions;
+const paramDescriptions: Record<mozjpeg_param, string> = {
```
surma, line 66: With the logic above removed, this file can be a singular default export of this dictionary.

src/features/encoders/mozJPEG/client/index.tsx
```diff
@@ -234,25 +268,87 @@ export class Options extends Component<Props, State> {
...
-                  <option value="3">ImageMagick</option>
-                  <option value="4">PSNR-HVS-M-tuned Kodak</option>
-                  <option value="5">Klein et al</option>
-                  <option value="6">Watson et al</option>
-                  <option value="7">Ahumada et al</option>
-                  <option value="8">Peterson et al</option>
+                  <option
+                    value="0"
+                    title={
+                      'Standard Quant Tables given in Annex K of the CCITT paper'
```
surma, line 284: Any reason why you put some descriptions in the meta file and some are inline? I think it’d be nice to move these into the meta file as well, so all descriptions for a codec are in one place.

surma (changes requested):
This is an amazing quality-of-life improvement for users. Really nice work. This is a really good start.

One thing we should talk about that the text in `title` is effectively inaccessible to mobile users, but this is already better than having no description at all.

Maybe as a follow-up we can move to custom tooltip so that we can control when and how it appears on mobile and maybe even do rich-text formatting and links.

I also have _not_ review the descriptions themselves just yet.

src/features/encoders/mozJPEG/shared/meta.ts
aryanpingle: Ahh you're right, it would be better suited for the utilities file. I'll make a generator version as well.

src/features/encoders/mozJPEG/shared/meta.ts
aryanpingle: How didn't I know about `dedent` till now, it's exactly what I wanted LOL. I'll use it now.

src/features/encoders/mozJPEG/client/index.tsx
aryanpingle: I wanna give descriptions for params _and_ their options (like `ImageMagick` for `quant_table`). Problem is, for intellisense to do some nice autocompletion for the `getParameterDescription` arguments, I'll need an interface containing all params and their options, which ends up looking very verbose IMO.

So I went with using `getParameterDescription` for the params, and inlining the descriptions for the options.

<hr>

I'll be refactoring all of this once I finish making a custom modal component (better explanations with better UI). For now I'll try to put all descriptions in the meta file; it'll definitely look cleaner once the modal is ready.

aryanpingle:
With the Modal PR this becomes pretty much obsolete

PR #1369, opened 2023-06-12
Adding A Modal Component

aryanpingle:
This PR adds a custom Modal component so that Squoosh can communicate better with users (_snackbars are SO last year_).

While the snackbar popup will convey small and concise messages to users, the Modal can be used for displaying verbose  information like:

1. **INFO** mode - Explaining codecs and their parameters
2. **UPDATE** mode - Informing users about new features
3. **ERROR** mode - Expanding error messages on mobile, where users can't view the console error messages

It's keyboard accessible, has pre-formatted styles for different HTML elements, is responsive across the board and can be easily accessed by other components/elements.

I've tested the feature on the following devices:

* Safari - iOS / Mac
* Chrome - iOS / Android / Windows
* Firefox - Android / Windows

---

### DEMO

In the right-side options under the quality slider, click on any of the modal checkboxes. If you're using a keyboard, focus will be restored to that checkbox once the modal is closed.

---

**Unnecessary PR Lore**: This is complementary to PR #1366 where Surma rightly pointed out that the HTML hover _title_ is inaccessible to mobile users. I initially planned to put small one-line descriptions for each codec param in the _title_ attribute, then later add modal support with more detailed descriptions. But I figure if those one-liners will be overwritten anyway, might as well do them soon.


aryanpingle:
@argyleink @surma could y'all take a look at this whenever you're free?

src/client/lazy-app/Compress/Output/index.tsx
```diff
@@ -331,6 +331,18 @@ export default class Output extends Component<Props, State> {
...
+            class={
+              style.previewSize +
+              ' ' +
+              (altBackground ? style.previewSizeDark : style.previewSizeLight)
+            }
+          >
+            {Math.round(scale * originalImage.width)} ×{' '}
+            {Math.round(scale * originalImage.height)}
+          </div>
+        )}
```
surma: Is this supposed to be part of the PR? I think this can be really useful but pretty independent of the feature :D

src/client/lazy-app/Compress/Modal/index.tsx
```diff
@@ -0,0 +1,148 @@
...
+    // Set element to return focus to after hiding
+    this.returnFocusElement = document.activeElement as HTMLElement;
+
+    this.modal.style.display = '';
+    this.setState({
+      message: message,
+      shown: true,
+    });
+    // Wait for the 'display' reset to take place, then focus
+    setTimeout(() => {
```
surma:
```suggestion
    requestAnimationFrame(() => {
```

surma: But on second thought, it might actually be more idiomatic to use the `ref` prop to focus the button once a DOM element been created for it. Something like

```js
  <button ref={(el) => el?.focus()} /*...*/>
```

src/client/lazy-app/Compress/Modal/index.tsx
```diff
@@ -0,0 +1,148 @@
...
+      shown: false,
+    });
+    setTimeout(() => {
+      this.modal && (this.modal.style.display = 'none');
+      this.returnFocusElement?.focus();
+    }, 250);
+  }
+
+  private _getCloseButton() {
+    return this.modal!.querySelector('button')!;
```
surma: We should have a dedicated class or id for the close button. Feels brittle otherwise with user-provided content for the modal.

src/client/lazy-app/Compress/Modal/index.tsx
```diff
@@ -0,0 +1,148 @@
...
+    } else {
+      // If TAB was pressed on the last focusable element
+      // Move focus to the first focusable element
+      if (document.activeElement === this._getLastFocusable()) {
+        this._getCloseButton().focus();
+        e.preventDefault();
+        e.stopImmediatePropagation();
+      }
+    }
+  }
```
surma: On the one hand, I am impressed that you managed to successfully trap focus manually :D

However, I think there are some platform features that we can use to achieve the same thing with less code. The first options is the `inert` attribute, that can make all other HTML in the tree unfocusable. Read [more here](https://developer.chrome.com/articles/inert/). The other option, and the even more natural fit I think, is using the `<dialog>` element, which is now [well supported](https://caniuse.com/dialog).

src/client/lazy-app/Compress/Modal/style.css
```diff
@@ -0,0 +1,174 @@
...
+  display: grid;
+  grid-template-columns: auto 1fr auto;
+  align-items: center;
+  gap: 0.5em;
+  border-bottom: 1px solid rgba(0, 0, 0, 8%);
+
+  & > * {
+    height: 100%;
+    font-size: 1.5em;
+  }
```
surma: This seems a bit dangerous. I think instead of setting a height on each child, setting `align-content: stretch;` on the header should do the trick. Same with the font size, which is automatically inherited.

src/features/encoders/mozJPEG/client/index.tsx
```diff
@@ -32,6 +36,46 @@ interface State {
...
+        &lt;code&gt;.
+      </p>
+      <figure>
+        <img src={SampleImage.default} alt="Alt Text" />
+        <figcaption>
+          Source: Nintendo EDP / Nintendo (and this is a figcaption)
+        </figcaption>
+      </figure>
+      <p>
+        That's a link.{' '}
```
surma, line 65: lol

src/client/lazy-app/Compress/Modal/modal-context.ts
surma, line 1: Considering that there can only ever be one modal shown at time (it’s a singleton), I wonder if we need to do the Context/Provider dance. We could just have a ES module that exports a `showModal()` function and forwards that to a `ref` of a single, global `<Modal>` instance. Not sure what others think of that, tho

src/client/lazy-app/Compress/Modal/index.tsx
```diff
@@ -0,0 +1,148 @@
...
+  DiamondStarIcon,
+  ModalErrorIcon,
+} from 'client/lazy-app/icons';
+
+interface Props {}
+
+export interface ModalMessage {
+  type: 'info' | 'error' | 'update';
+  title: string;
+  content: string;
```
surma: I don’t think `content` should be a string, but rather a `VNode`. With string we lose the ability to have more complex interactivity in the modal (event listeners would get lost etc). Even `title` should probably be a `VNode`.

The next logical step is to make the Modal itself actually unaware of whether it’s an error or info or update. It’s just a component for modal dialogs, and since we can pass arbitrary VNodes for title and content, the user can style it depending on what the use-case is.

src/client/lazy-app/Compress/Output/index.tsx
surma: Screenshot 2023-06-20 at 10 52 17
The text can also get quite easily lost due to contrast.

surma (changes requested):
Great work! This is gonna be a super useful primitive to have in Squoosh going forward.

Have a couple of remarks, and a bigger rework request/idea around the `<dialog>` element. LMK what you think!

src/client/lazy-app/Compress/Modal/index.tsx
aryanpingle: I'm confused, will that work in this case? I'm not adding/removing the Modal component to the DOM tree on show/hide, just changing some CSS properties to "hide" it from the user.

src/client/lazy-app/Compress/Modal/index.tsx
aryanpingle: That was just lazy coding on my part LOL, I'll make a dedicated id for it

src/client/lazy-app/Compress/Modal/index.tsx
aryanpingle: Ahh you're right, `VNode` would be better suited. Will do!

src/client/lazy-app/Compress/Modal/index.tsx
aryanpingle: I **did** want to use the `<dialog>` element, but isn't it fairly new on iOS? I personally know lots of people using iOS 14 and who stubbornly refuse to update. But yeah, most user devices should be able to support it, and I'm all for using a cool new web feature. Lemme know what you think!

aryanpingle: I was iffy about using `inert` because I'd have to apply it to all element siblings of the Modal components. Wouldn't that initiate a huge reflow every time the Modal is shown/hidden?

src/client/lazy-app/Compress/Modal/index.tsx
```diff
@@ -0,0 +1,148 @@
+import { h, Component } from 'preact';
```
jakearchibald: This file is in `/Compress/` - is its usage intended to be scoped to this part of the app, or is it for more general use?

src/client/lazy-app/Compress/Modal/index.tsx
jakearchibald: Wouldn't it be more natural for the content to be the node's children? (like how `<dialog>` works)

src/client/lazy-app/Compress/Modal/index.tsx
```diff
@@ -0,0 +1,148 @@
...
+import {
+  InfoIcon,
+  DiamondStarIcon,
+  ModalErrorIcon,
+} from 'client/lazy-app/icons';
+
+interface Props {}
+
+export interface ModalMessage {
+  type: 'info' | 'error' | 'update';
```
jakearchibald: Is the intent to use the 'type' for other things? Right now it's just impacting the icon. So, maybe accept a vnode for the icon?

src/client/lazy-app/Compress/Modal/index.tsx
```diff
@@ -0,0 +1,148 @@
...
+    message: {
+      type: 'info',
+      title: 'default',
+      content: 'default',
+    },
+    shown: false,
+  };
+
+  private modal?: HTMLElement;
+  private returnFocusElement?: HTMLElement | null;
```
jakearchibald: Not required, but we can start using 'real' private properties. If you want to do that, you might need to adjust the config.

The benefit in real private vars is their names compress.

src/client/lazy-app/Compress/Modal/index.tsx
jakearchibald: I think the correct Preact way to do this would be in `componentDidUpdate`. But, you get all this stuff for free with `<dialog>`, so we should just use that.

src/client/lazy-app/Compress/Modal/index.tsx
jakearchibald: Also seems odd to have a method like this, but then other code above getting the button manually. But I don't think we'll need this at all with `<dialog>`.

src/client/lazy-app/Compress/Modal/modal-context.ts
jakearchibald: I think this is another problem solved by `<dialog>`

jakearchibald (changes requested):
I didn't do a full review, as I think this will change pretty significantly when it moves to use `<dialog>`, which will hopefully reduce the complexity.

src/client/lazy-app/Compress/Modal/index.tsx
aryanpingle: General use. Maybe I could put it in the shared directory like the snack bar element?

src/client/lazy-app/Compress/Modal/index.tsx
aryanpingle: I can't really think of reasons other than these 3 for displaying a modal - so I thought it might be better to abstract away the Icon behind the `type` parameter

src/client/lazy-app/Compress/Modal/index.tsx
aryanpingle: Ahh got it, will do

src/client/lazy-app/Compress/Modal/index.tsx
jakearchibald: The benefit of providing the icon is unused icons won't be bundled

src/client/lazy-app/Compress/Modal/index.tsx
jakearchibald: `lazy-app` is a good place for it if it isn't needed for the first interaction.

src/client/lazy-app/Compress/Modal/style.css
```diff
@@ -0,0 +1,174 @@
...
+  code {
+    background-color: rgba(0, 0, 0, 10%);
+    padding: 2px 4px;
+    border-radius: 3px;
+  }
+
+  pre {
+    overflow-x: auto;
+  }
+}
```
argyleink:
```suggestion
  hr {
    border: none;
    border-bottom: 1px solid var(--dim-text);
  }
}
```

src/client/lazy-app/Compress/Modal/style.css
```diff
@@ -0,0 +1,174 @@
...
+      color: hsl(220, 100%, 50%);
+    }
+    &:visited {
+      color: rgb(255, 0, 103);
+    }
+  }
+
+  pre,
+  code {
+    background-color: rgba(0, 0, 0, 10%);
```
argyleink:
```suggestion
    background-color: var(--black);
```

src/client/lazy-app/Compress/Modal/style.css
```diff
@@ -0,0 +1,174 @@
...
+    padding: 2px 4px;
+    border-radius: 3px;
+  }
+
+  pre {
+    overflow-x: auto;
```
argyleink:
```suggestion
    overflow-x: auto;
    padding: 1rem 2rem;
```

src/client/lazy-app/Compress/Modal/style.css
```diff
@@ -0,0 +1,174 @@
...
+  max-height: 60vh;
+  overflow-y: auto;
+}
+
+.content {
+  font-size: 1em;
+  line-height: 1.5;
+
+  padding: 0 1em;
+
```
argyleink:
```suggestion
  h1 {
    color: var(--white);
  }
```

src/client/lazy-app/Compress/Modal/style.css
```diff
@@ -0,0 +1,174 @@
...
+    font-size: 1em;
+  }
+}
+
+.header,
+.footer {
+  max-height: 100%;
+  padding: 1em;
+
+  background-color: rgba(0, 0, 0, 6%);
```
argyleink:
```suggestion
  background-color: var(--dark-gray);
```

src/client/lazy-app/Compress/Modal/style.css
```diff
@@ -0,0 +1,174 @@
...
+
+  opacity: 1;
+}
+
+.modal {
+  width: 70vw;
+  max-width: 60ch;
+  max-height: 70vh;
+
+  border: 4px solid black;
```
argyleink:
```suggestion
```

src/client/lazy-app/Compress/Modal/style.css
```diff
@@ -0,0 +1,174 @@
...
+
+  display: grid;
+  grid-template-rows: auto 1fr auto;
+
+  font-size: 1.5em;
+
+  border-radius: 10px;
+  overflow: hidden;
+
+  background-color: hsl(250deg, 100%, 90%);
```
argyleink:
```suggestion
  background-color: var(--off-black);
  color: var(--light-gray);
```

src/client/lazy-app/Compress/Modal/style.css
```diff
@@ -0,0 +1,174 @@
...
+
+  figcaption {
+    font-size: 0.8em;
+    text-align: center;
+    font-style: italic;
+  }
+
+  a {
+    &:link {
+      color: hsl(220, 100%, 50%);
```
argyleink:
```suggestion
      color: var(--blue);
```

src/client/lazy-app/Compress/Modal/style.css
```diff
@@ -0,0 +1,174 @@
...
+    }
+    &:visited {
+      color: rgb(255, 0, 103);
```
argyleink:
```suggestion
      color: var(--pink);
```

argyleink (changes requested):
Overall pretty good! Spacing works, sizes work, but the colors were a bit off the rails, so I've made a few suggestions to bring it back into the theme.

Screenshot 2023-06-26 at 1 46 57 PM

A couple other notes!
- I agree with others that dialog is a good choice here, it'll save a bunch of maintenance burden (like trying to trap focus, etc) and I dont think a majority of the users are on iOS making these transformations
- The labels should probably also be keyboard focusable. It's a bit strange that focus is "restored" but not really to where I came from, rather the nearest checkbox to where I came from. Also not sure how to invoke this from keyboard, tho it's clear how to dismiss it.

Cool idea here. Hope this review is helpful!

aryanpingle:
@argyleink That's a wayyy better design, thank you for the review!

And yep, my plan is to give an optional prop to the `Range` / `Checkbox` / etc components that turns their labels into modal-activating keyboard-navigable buttons. So far this was just a test to see if focus can be returned to the calling checkbox, which it can.

Working on a refactor using the `<dialog>` element as we speak 👍

aryanpingle:
- [x] Placed the Modal component directory in `lazy-app` instead of in `Compress`
- [x] Refactored using the `<dialog>` element
- [x] Used the improved design by @argyleink 
- [x] Switched to using `VNode`s for modal content and icon
- [x] Switched from using Context/Provider to shared functions 

Caveats:

* ~~Still uses Context/Provider to share `showModal()`. Can't use props like with `showSnack` because deeply nested children should be able to access it. Gonna try @surma 's suggestion of a single global `showModal()` function that can use the modal as soon as it's mounted.~~
* The CSS for fading the modal in and out is a bit ugly, but I couldn't find an easier way considering the dialog element's nuances.

aryanpingle:
Switched from using Context/Provider to shared functions

src/client/lazy-app/Modal/index.tsx
```diff
@@ -0,0 +1,128 @@
...
+  private _hideModal() {
+    if (!this.dialogElement || !this.dialogElement.open) return;
+
+    // Make the modal fade out
+    this.dialogElement.classList.add(style.modalClosing);
+
+    this.setState({
+      message: { ...this.state.message },
+      shown: false,
+    });
```
surma:
```suggestion
  this.setState(cleanSet(this.state, `shown`, false));
```

src/client/lazy-app/Modal/index.tsx
```diff
@@ -0,0 +1,128 @@
...
+  }
+
+  private _onKeyDown(e: KeyboardEvent) {
+    // Default behaviour of <dialog> closes it instantly when you press Esc
+    // So we hijack it to smoothly hide the modal
+    if (e.key === 'Escape' || e.keyCode == 27) {
```
surma: I [don’t think](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent#:~:text=Toggle%20history-,key,-51) we need both.

```suggestion
    if (e.key === 'Escape') {
```

src/client/lazy-app/Modal/modal-context.ts
surma, line 1: I think this file is unused?

src/client/lazy-app/Modal/index.tsx
```diff
@@ -0,0 +1,128 @@
...
+  static showModal(message: ModalMessage) {
+    Modal.modalInstance?._showModal(message);
+  }
+
+  static hideModal() {
+    Modal.modalInstance?._hideModal();
+  }
+
+  private _showModal(message: ModalMessage) {
+    if (!this.dialogElement) return;
```
surma: It might be worth throwing here (something like “called `Modal.showModal()` without putting the `<Modal />` component being present”).

surma (approved):
This looks really good!

Some minor nits, and of course you should remove the dummy checkboxes and content before merging, but otherwise LGTM! 🎉

src/client/lazy-app/Modal/index.tsx
aryanpingle: TIL about `cleanSet`. On it!

src/client/lazy-app/Modal/modal-context.ts
aryanpingle: Nice catch, I'll remove it from the commits and do some cleanup

src/client/lazy-app/Modal/index.tsx
aryanpingle: Sure thing!

argyleink (approved):
approved but with 1 change request!
- [x] the mobile layout has the dialog at `translateY(0)`, doesnt look right: pop that to `20px` or so it looks more intentional 👍🏻

src/client/lazy-app/Modal/style.css
```diff
@@ -0,0 +1,200 @@
...
+
+  &.modal--closing::backdrop {
+    transition: opacity 250ms linear;
+    opacity: 0;
+  }
+}
+
+dialog[open] {
+  opacity: 1;
+  transform: translateY(0px);
```
aryanpingle: @argyleink I'm a bit confused, you suggested 20px to show intent - but wouldn't it be better to add a comment here instead of changing the translation amount? 20px isn't making much of a visual difference on mobile IMO

argyleink: sorry for the confusion!

this
Screenshot 2023-07-10 at 12 20 45 PM

is better than this
Screenshot 2023-07-10 at 12 20 33 PM

is what I'm getting at. the intent is to have a modal, and it's weird if it's stuck to the top flush, better if it's got some breathing room from the top edge.

aryanpingle: Huh, that's weird - it's supposed to be centered on all devices/sizes, not stuck to the top. Could you tell me which browser/os you're using? I've tested it on most Android browsers and remotely tested on a couple iOS versions too, seemed to work fine on those.

argyleink: Chrome stable 114

argyleink: confirm devtools issue

src/client/lazy-app/Modal/index.tsx
```diff
@@ -0,0 +1,127 @@
...
+  shown: boolean;
+}
+
+export default class Modal extends Component<Props, State> {
+  state: State = {
+    message: {
+      icon: <svg></svg>,
+      title: 'default',
+      content: <Fragment></Fragment>,
+    },
```
jakearchibald: What's the benefit of having this as one object rather than individual properties.

jakearchibald: Why are these state rather than props?

src/client/lazy-app/Modal/index.tsx
```diff
@@ -0,0 +1,127 @@
...
+    shown: false,
+  };
+
+  private dialogElement!: HTMLDialogElement;
+  static modalInstance?: Modal | undefined;
```
jakearchibald: Does it need to be static? One of the benefits of `<dialog>` is that it can be deep in the DOM but still pop out to the top-level. I wonder if you're trying to recreate that feature here, but given it already exists in the platform, we can just use it.

src/client/lazy-app/Modal/index.tsx
```diff
@@ -0,0 +1,127 @@
...
+
+  componentDidMount() {
+    // Once a transition ends, check if the modal should be closed (not just hidden)
+    // dialog.close() instantly hides the modal, so we call it AFTER fading it out i.e. on transition end
+    this.dialogElement.addEventListener(
+      'transitionend',
+      this._closeOnTransitionEnd.bind(this),
+    );
```
jakearchibald: I don't think this is the right way to do this. The `transitionend` even bubbles, so the wrong transition may end up closing the dialog. Also, if it somehow ends up trying to transition to the same value, this event won't trigger. No requirement to watch it, but there's a 203 that covers these gotchas https://www.youtube.com/watch?v=9-6CKCz58A8

If you want to do something after an animation, the safest way is to do the animation with the web animation API, and then use its `finished` promise.

src/client/lazy-app/Modal/index.tsx
```diff
@@ -0,0 +1,127 @@
...
+    this.dialogElement.setAttribute('inert', 'enabled');
```
jakearchibald: I think this stuff will be refactored anyway, but for future reference, you can just do:

```js
el.inert = true;
```

src/client/lazy-app/Modal/index.tsx
```diff
@@ -0,0 +1,127 @@
...
+
+    Modal.modalInstance = this;
+  }
+
+  private _closeOnTransitionEnd() {
```
jakearchibald: fwiw, I'm happy for new code to use real JS privacy. As in `#closeOnTransitionEnd`

jakearchibald (changes requested):
This is heading in the right direction, but I think trying to make the modal a singleton is unnecessary. 

One of the benefits of `<dialog>` is that it doesn't need to be placed out-of-context in the top of the DOM, as it pops into the top-layer when it opens. It feels like the implementation will become simpler if we lean into that ("use the platform" etc etc).

So, the usage of this component would be:

```jsx
<Modal shown={bool} icon={iconJSX} title="Hello!">
  Modal content goes here.
</Modal>
```

We'd have many of those in our JSX, in context.

And when `shown` changes, the component handles the animated show and hide.

I guess at some point we'd also have a tooltip component that outputs an icon and a modal, and causes the modal to show when the icon is clicked.

jakearchibald:
Thinking about the helper component:

```jsx
<ModalHelpIcon title="Hello!">
  Modal content goes here.
</ModalHelpIcon>
```

So the above would output a ℹ️ icon (or whatever), and a `<Modal />`. When the icon is clicked, it shows the modal, by toggling the `shown` prop passed to `<Modal />`.

That means we can just add `<ModalHelpIcon />` where we want to details codecs etc, and that component takes care of the show/hide state.

aryanpingle:
Added a `ModalHint` component that contains a Modal within itself and shows it when activated. Went with a subtle green accent color so that hints like in the image below stand out. @argyleink, would love your thoughts on this!

image

argyleink:
fun additions!

I've got a 2 overall comments 🙂 

1. replace the green with `var(--blue)` (maybe) or `var(--less-light-gray)` (prolly), like you've used in the dialog header here. info icons shouldn't try and steal attention, unless hovered, you can spice them up then.  
Screenshot 2023-07-26 at 7 38 22 AM

2. there's alignment issues, here's some screenshots 
Screenshot 2023-07-26 at 7 38 06 AM Screenshot 2023-07-26 at 7 38 33 AM

aryanpingle:
* Fixed `ModalHint` alignment issues
* Switched from transitionEnd event handler to the Web Animations API

src/features/encoders/mozJPEG/client/index.tsx
```diff
@@ -114,7 +163,9 @@ export class Options extends Component<Props, State> {
             value={options.quality}
             onInput={this.onChange}
           >
-            Quality:
+            <ModalHint modalTitle="mozJPEG - quality" content={sampleContent()}>
+              Quality:
+            </ModalHint>
```
jakearchibald: I think it makes more sense for the content of the modal to be the child of `<ModalHint>`, and the 'label' to be a prop. The label could probably just be a string too.

src/client/lazy-app/icons/index.tsx
```diff
@@ -11,6 +11,36 @@ const Icon = (props: preact.JSX.HTMLAttributes) => (
...
+    />
+  </svg>
```
jakearchibald: Passed it through SVGOMG.

```suggestion
  <svg viewBox="0 0 512 512" fill="currentColor">
    <path fill-rule="evenodd" d="M64 0C29 0 0 29 0 64v384c0 35 29 64 64 64h384c35 0 64-29 64-64V64c0-35-29-64-64-64H64Zm234 377 3-3c4-5 6-9 4-16-2-8-11-11-19-9-10 4-27 15-27 15l13-64 4-15c6-25 14-53-4-72-15-16-34-12-51 4l-24 25-3 4c-4 5-6 9-4 16s11 11 18 8c11-3 28-14 28-14l-12 56-1 1c-3 12-20 73-1 94 15 16 35 12 51-4l25-26Zm-72-240a43 43 0 1 0 87-8 43 43 0 0 0-87 8Z"/>
  </svg>
```

src/client/lazy-app/icons/index.tsx
```diff
@@ -11,6 +11,36 @@ const Icon = (props: preact.JSX.HTMLAttributes) => (
...
+);
+
+export const ModalErrorIcon = (props: preact.JSX.HTMLAttributes) => (
+  <svg viewBox="0 0 24 24" fill="currentColor">
+    <path
+      fill-rule="evenodd"
+      clip-rule="evenodd"
+      d="M12 22c5.523 0 10-4.477 10-10S17.523 2 12 2 2 6.477 2 12s4.477 10 10 10zm-1.5-5.009c0-.867.659-1.491 1.491-1.491.85 0 1.509.624 1.509 1.491 0 .867-.659 1.509-1.509 1.509-.832 0-1.491-.642-1.491-1.509zM11.172 6a.5.5 0 0 0-.499.522l.306 7a.5.5 0 0 0 .5.478h1.043a.5.5 0 0 0 .5-.478l.305-7a.5.5 0 0 0-.5-.522h-1.655z"
+    />
+  </svg>
```
jakearchibald:
```suggestion
  <svg viewBox="0 0 24 24" fill="currentColor">
    <path
      fill-rule="evenodd"
      d="M12 22c5.523 0 10-4.477 10-10S17.523 2 12 2 2 6.477 2 12s4.477 10 10 10zm-1.5-5.009c0-.867.659-1.491 1.491-1.491.85 0 1.509.624 1.509 1.491 0 .867-.659 1.509-1.509 1.509-.832 0-1.491-.642-1.491-1.509zM11.172 6a.5.5 0 0 0-.499.522l.306 7a.5.5 0 0 0 .5.478h1.043a.5.5 0 0 0 .5-.478l.305-7a.5.5 0 0 0-.5-.522h-1.655z"
    />
  </svg>
```

src/client/lazy-app/icons/index.tsx
```diff
@@ -11,6 +11,36 @@ const Icon = (props: preact.JSX.HTMLAttributes) => (
...
+);
+
+export const DiamondStarIcon = (props: preact.JSX.HTMLAttributes) => (
+  <svg
+    viewBox="0 0 480 480"
+    fill="currentColor"
+    xmlns="http://www.w3.org/2000/svg"
+  >
+    <path d="M240 480C240 360 120 240 0 240C120 240 240 120 240 0C240 120 360 240 480 240C360 240 240 360 240 480Z" />
+  </svg>
```
jakearchibald:
```suggestion
  <svg viewBox="0 0 48 48" fill="currentColor">
    <path d="M24 48c0-12-12-24-24-24 12 0 24-12 24-24 0 12 12 24 24 24a25.8 25.8 0 0 0-24 24Z"/>
  </svg>
```

src/client/lazy-app/Modal/style.css
```diff
@@ -0,0 +1,171 @@
+dialog {
```
jakearchibald: Instead of styling all dialogs, use a class name. That avoids clashes with other dialogs we may have in future.

src/client/lazy-app/Modal/style.css
```diff
@@ -0,0 +1,171 @@
...
+  composes: unbutton from global;
+  flex: 0 0 auto;
+  height: 1em;
+  width: 1em;
+  background-color: var(--pink);
+  border-radius: 5px;
+  display: grid;
+  outline-color: var(
+    --medium-light-gray
+  ) !important; /* Overwrite default black outline */
```
jakearchibald: Where's this default coming from? As in, why does it need `!important`?

src/client/lazy-app/Modal/style.css
```diff
@@ -0,0 +1,171 @@
...
+  }
+
+  /* Don't show focus ring on mobile */
+  @media (min-width: 720px) {
+    &:focus {
+      outline-width: 2px;
+      outline-style: solid;
+      outline-offset: 2px;
+    }
+  }
```
jakearchibald: Would it be better to use https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible ?

src/client/lazy-app/Modal/index.tsx
```diff
@@ -0,0 +1,136 @@
...
+                onClick={() => this._hideModal()}
+              >
+                <svg viewBox="0 0 480 480" fill="currentColor">
+                  <path
+                    d="M119.356 120L361 361M360.644 120L119 361"
+                    stroke="#fff"
+                    stroke-width="37"
+                    stroke-linecap="round"
+                  />
+                </svg>
```
jakearchibald: Move this to icons?

src/client/lazy-app/Modal/index.tsx
```diff
@@ -0,0 +1,136 @@
...
+      e.preventDefault();
+      e.stopImmediatePropagation();
+    }
+  }
+
+  render({ title, icon, content }: Props, { shown }: State) {
+    return (
+      <dialog
+        ref={linkRef(this, 'dialogElement')}
+        onKeyDown={(e) => this._onKeyDown(e)}
```
jakearchibald: You can avoid creating a function on each render by using class fields. See the pattern here https://github.com/GoogleChromeLabs/squoosh/blob/dev/src/shared/prerendered-app/Intro/index.tsx#L282

src/client/lazy-app/Modal/index.tsx
```diff
@@ -0,0 +1,136 @@
...
+      >
+        {shown && (
+          <Fragment>
+            <header class={style.header}>
```
jakearchibald: If you're going to use `<header>` and `<footer>`, then wrap both in an `<article>`. This semantically links the header and footer.

src/client/lazy-app/Modal/index.tsx
```diff
@@ -0,0 +1,136 @@
...
+      animateTo(this.dialogElement, [{ opacity: 0 }], {
+        duration: 250,
+        easing: 'linear',
+        pseudoElement: '::backdrop',
+      });
+    } catch (e) {}
+    anim.onfinish = this._closeOnTransitionEnd;
+  }
+
+  private _onKeyDown(e: KeyboardEvent) {
```
jakearchibald: Nit: Rename the arg to `event`. Leave minification to the minifier 😄

src/client/lazy-app/Modal/index.tsx
```diff
@@ -0,0 +1,136 @@
...
+    if (!this.dialogElement) return;
+
+    this.dialogElement.close();
+    this.dialogElement.inert = true;
+    this.setState({ shown: false });
+  };
+
+  showModal() {
+    if (!this.dialogElement || this.dialogElement.open)
+      throw Error('Modal missing / already shown');
```
jakearchibald: My preference is that, if the thing is already in the desired state, then just return rather than throwing. Also, try to avoid using both component state & dialog element state. Use one as the source of truth (the component state?)

src/client/lazy-app/Modal/index.tsx
```diff
@@ -0,0 +1,136 @@
...
+}
+
+export default class Modal extends Component<Props, State> {
+  private dialogElement?: HTMLDialogElement;
+
+  componentDidMount() {
+    if (!this.dialogElement) throw new Error('Modal missing');
+
+    // Set inert by default
+    this.dialogElement.inert = true;
```
jakearchibald: Can you talk me through the usage of `inert` in this component?

src/client/lazy-app/Modal/index.tsx
```diff
@@ -0,0 +1,136 @@
...
+
+    // Prevent events from leaking through the dialog
+    this.dialogElement.onclick = (event) => {
+      event.preventDefault();
+      event.stopImmediatePropagation();
+    };
```
jakearchibald: Can you talk me through the reason for this, and why the event is attached this way, rather than the usual Preact way?

src/client/lazy-app/Modal/index.tsx
```diff
@@ -0,0 +1,136 @@
...
+      { duration: 250, easing: 'ease' },
+    );
+    // animate modal::backdrop
+    // some browsers don't support ::backdrop, catch those errors
+    try {
+      animateTo(this.dialogElement, [{ opacity: 0 }, { opacity: 1 }], {
+        duration: 250,
+        easing: 'linear',
+        pseudoElement: '::backdrop',
+      });
```
jakearchibald: Would it make more sense to use `animateFrom` in cases like this?

src/client/lazy-app/Modal/index.tsx
```diff
@@ -0,0 +1,136 @@
...
+    // animate modal closing
+    const anim = animateTo(
+      this.dialogElement,
+      { opacity: 0, transform: 'translateY(50px)' },
+      { duration: 250, easing: 'ease' },
+    );
+    // animate modal::backdrop
+    // some browsers don't support ::backdrop, catch those errors
+    try {
+      animateTo(this.dialogElement, [{ opacity: 0 }], {
```
jakearchibald, line 73:
```suggestion
      animateTo(this.dialogElement, { opacity: 0 }, {
```

src/client/lazy-app/Modal/index.tsx
```diff
@@ -0,0 +1,136 @@
...
+        duration: 250,
+        easing: 'linear',
```
jakearchibald:
```suggestion
        easing: 'ease',
```

src/client/lazy-app/Modal/index.tsx
```diff
@@ -0,0 +1,136 @@
...
+        { opacity: 1, transform: 'translateY(0px)' },
+      ],
+      { duration: 250, easing: 'ease' },
+    );
+    // animate modal::backdrop
+    // some browsers don't support ::backdrop, catch those errors
+    try {
+      animateTo(this.dialogElement, [{ opacity: 0 }, { opacity: 1 }], {
+        duration: 250,
+        easing: 'linear',
```
jakearchibald:
```suggestion
        easing: 'ease',
```

It looks better 😄

src/client/lazy-app/Modal/ModalHint/style.css
```diff
@@ -0,0 +1,31 @@
+span.modal-hint {
```
jakearchibald: Why require a particular kind of element here? (`span`)

src/client/lazy-app/Modal/ModalHint/index.tsx
```diff
@@ -0,0 +1,56 @@
...
+  };
+
+  render({ modalTitle, content }: Props) {
+    return (
+      <span
+        class={style.modalHint}
+        onClick={(event) => {
+          event.preventDefault();
+          event.stopImmediatePropagation();
+        }}
```
jakearchibald: Why is this needed?

jakearchibald (changes requested):
This is coming along nicely!

Where did the icons come from? Do we have the rights to use them?

src/client/lazy-app/Modal/ModalHint/index.tsx
aryanpingle: Might be obsolete because I prevented click events from _leaking_ out of the dialog, I'll look into it

src/client/lazy-app/Modal/ModalHint/style.css
aryanpingle: No reason in particular, just seemed to make it easier to understand

src/client/lazy-app/Modal/index.tsx
aryanpingle: I have a feeling it's another instance of obsolete code now that we're using the native `dialog` element, but here's the logic: when hidden, the dialog is made inert to prevent its contents from being `TAB`bed to. Might be a remnant of the time when I didn't use the dialog element and had to ensure the modal wasn't focusable.

src/client/lazy-app/Modal/index.tsx
aryanpingle: No good reason, I'll refactor it

src/client/lazy-app/Modal/index.tsx
aryanpingle: (For browsers that support backdrop) We need fill-mode `forwards` or `both` to keep the backdrop opaque after the animation - but animateFrom uses `back`.
If I edit that part of animateFrom then it essentially becomes animateTo without committing styles, and in that case a try-catch is much more readable IMO.

src/client/lazy-app/Modal/index.tsx
aryanpingle: Didn't cross my mind that I could change it to an arrow function to preserve the `this` context, will do!

src/client/lazy-app/Modal/style.css
aryanpingle: The global `unbutton` sets the outline-color with the focus selector, so that overrides the default

...and now that I think of it, I should've just done the same and set it on focus - eliminating `!important`. Will do.

src/features/encoders/mozJPEG/client/index.tsx
aryanpingle: Hmm I suppose that makes more sense. My thought process was we can just wrap existing text nodes (like for the custom input elements) in the `ModalHint` and get the modal functionality without much refactoring - but in hindsight that's a cop-out.

src/client/lazy-app/Modal/index.tsx
jakearchibald: I think I get it, but wouldn't the backdrop's default from-CSS opacity be 1, so it wouldn't need a fill-forward?

src/features/encoders/mozJPEG/client/index.tsx
jakearchibald: It works both ways, and in this case it isn't clear-cut which thing is naturally the "children". But "content" sounds like just another name for "children", whereas the actual child is the label. That, and I usually lean towards having the larger, more tree-like content as the child.

But yeah, it isn't clear-cut.

src/client/lazy-app/Modal/index.tsx
jakearchibald: I think so. `<dialog>` content isn't tabbable when it's not-shown.

src/client/lazy-app/Modal/ModalHint/style.css
jakearchibald: The good thing about using class names (and CSS modules) is that you can decouple style and semantics. So, if a particular `modal-hint` made sense as an `<aside>`, then that's fine, because the class name can be applied to any element. However, in this case it can only be applied to `<span>`, and I don't think that's a useful restriction.

In general I'd just use class names, and only use tag names in cases where it's justified. Cases like `.icon > svg` are fiiiine, and your use of tag names for dialog content is great, but I think just sticking with a class name is better here.

jakearchibald:
@argyleink @kosamari @surma UX question: Right now the whole label becomes an activator for the modal (see "Auto sample chroma" in the preview). My hunch is that only the icon should activate the modal, and the rest of the label should remain as an activator for the checkbox. Can I get a second opinion?

src/client/lazy-app/icons/index.tsx
```diff
@@ -11,6 +11,46 @@ const Icon = (props: preact.JSX.HTMLAttributes) => (
...
+    <defs>
+      <clipPath id="a">
+        <rect width="520" height="520" rx="70" fill="currentColor" />
+      </clipPath>
+    </defs>
+  </svg>
+);
+
+export const DiamondStarIcon = (props: preact.JSX.HTMLAttributes) => (
+  <svg viewBox="0 0 48 48" fill="currentColor">
```
jakearchibald:
```suggestion
  <svg viewBox="0 0 48 48" fill="currentColor" {...props}>
```

Otherwise the parameter is unused.

src/client/lazy-app/icons/index.tsx
```diff
@@ -11,6 +11,46 @@ const Icon = (props: preact.JSX.HTMLAttributes) => (
...
+        d="M520 0H0v520h520V0ZM230.122 77.186c-8.096 7.7587-12.144 17.3727-12.144 28.842 0 11.469 4.048 21.083 12.144 28.842s17.71 11.638 28.842 11.638c11.132 0 20.578-3.879 28.336-11.638 8.096-7.759 12.144-17.373 12.144-28.842 0-11.4693-4.048-21.0833-12.144-28.842-7.758-8.096-17.204-12.144-28.336-12.144-11.132 0-20.746 4.048-28.842 12.144Zm1.012 106.766h-5.566v199.364c0 10.795 1.856 19.903 5.566 27.324 4.048 7.084 9.108 12.987 15.18 17.71 6.41 4.385 13.325 7.59 20.746 9.614 7.759 2.024 15.518 3.036 23.276 3.036h3.036V243.154c0-20.577-6.072-35.589-18.216-45.034-11.806-9.445-26.48-14.168-44.022-14.168Z"
+        fill="currentColor"
+      />
+    </g>
+    <defs>
+      <clipPath id="a">
+        <rect width="520" height="520" rx="70" fill="currentColor" />
+      </clipPath>
+    </defs>
+  </svg>
```
jakearchibald:
```suggestion
  <svg viewBox="0 0 520 520" fill="currentColor" {...props}>
    <path d="M70 0C31 0 0 31 0 70v380c0 39 31 70 70 70h380c39 0 70-31 70-70V70c0-39-31-70-70-70H70zm189 65c11 0 21 4 28 12 8 8 12 18 12 29s-4 21-12 29c-7 8-17 12-28 12s-21-4-29-12-12-18-12-29 4-21 12-29 18-12 29-12zm-33 119h5c18 0 32 5 44 14 12 10 18 25 18 45v198h-3l-23-3c-7-2-14-5-21-10-6-4-11-10-15-17-4-8-5-17-5-28V184z"/>
  </svg>
```

- Optimised the path
- Added the `{...props}`, otherwise it's an unused argument
- Removed the `xmlns`, it isn't needed in HTML
- Removed the clip-path - it's ok to use clip-paths, but remember that `id="a"` would be going into the document, and IDs have document scope. As it was, the icon would likely weirdly break if there was ever something else with `id="a"` (probably another icon that uses the same pattern).

src/client/lazy-app/Modal/style.css
```diff
@@ -0,0 +1,171 @@
...
+  display: grid;
+  outline-color: var(--medium-light-gray);
+
+  color: black; /* Influences currentColor of the icon */
+
+  &:focus {
+    outline: var(--medium-light-gray); /* Overwrite default black outline */
+  }
+
+  /* Don't show focus ring on mobile */
```
jakearchibald, line 112:
```suggestion
  /* Only show the focus ring on keyboard interaction */
```

src/client/lazy-app/Modal/index.tsx
```diff
@@ -0,0 +1,120 @@
...
+import 'add-css:./style.css';
+import { linkRef } from 'shared/prerendered-app/util';
+import { cleanSet } from '../util/clean-modify';
+import { animateFrom, animateTo } from '../util';
+import { RoundedCrossIcon } from '../icons';
+
+interface Props {
+  icon: VNode;
+  title: string;
+  content: VNode;
```
jakearchibald, line 12: I think we decided that the content of the model should be the children of the modal?

src/client/lazy-app/Modal/index.tsx
```diff
@@ -0,0 +1,120 @@
...
+  render({ title, icon, content }: Props, {}: State) {
+    return (
+      <dialog
+        class={style['modalDialog']}
+        ref={linkRef(this, 'dialogElement')}
+        onKeyDown={this._onKeyDown}
+        onClick={(event) => {
+          // Prevent clicks from leaking out of the dialog
+          event.preventDefault();
+          event.stopImmediatePropagation();
```
jakearchibald, line 101: I don't think it needs to stop immediate propogation?

src/client/lazy-app/Modal/index.tsx
```diff
@@ -0,0 +1,120 @@
...
+        }}
+      >
+        <article class={style.article}>
+          <header class={style.header}>
+            <span class={style.modalIcon}>{icon}</span>
+            <span class={style.modalTitle}>{title}</span>
```
jakearchibald, line 107: This should be a heading element. `<h1>` seems fine.

src/client/lazy-app/Modal/index.tsx
```diff
@@ -0,0 +1,120 @@
...
+            <button class={style.closeButton} onClick={() => this._hideModal()}>
+              <RoundedCrossIcon></RoundedCrossIcon>
```
jakearchibald, line 109:
```suggestion
              <RoundedCrossIcon/>
```

src/client/lazy-app/Modal/index.tsx
```diff
@@ -0,0 +1,120 @@
...
+      this._hideModal();
+      event.preventDefault();
+      event.stopImmediatePropagation();
+    }
+  };
+
+  render({ title, icon, content }: Props, {}: State) {
+    return (
+      <dialog
+        class={style['modalDialog']}
```
jakearchibald, line 95:
```suggestion
        class={style.modalDialog}
```

src/client/lazy-app/Modal/index.tsx
```diff
@@ -0,0 +1,120 @@
...
+    anim.onfinish = this._closeOnTransitionEnd;
+  }
+
+  private _onKeyDown = (event: KeyboardEvent) => {
+    // Default behaviour of <dialog> closes it instantly when you press Esc
+    // So we hijack it to smoothly hide the modal
+    if (event.key === 'Escape') {
+      this._hideModal();
+      event.preventDefault();
+      event.stopImmediatePropagation();
```
jakearchibald, line 88: I don't think `stopImmediatePropagation` is needed

src/client/lazy-app/Modal/index.tsx
```diff
@@ -0,0 +1,120 @@
...
+    // animate modal::backdrop
+    // some browsers don't support ::backdrop, catch those errors
+    try {
+      animateTo(this.dialogElement, [{ opacity: 0 }], {
+        duration: 250,
+        easing: 'ease',
+        pseudoElement: '::backdrop',
+      });
+    } catch (e) {}
+    anim.onfinish = this._closeOnTransitionEnd;
```
jakearchibald, line 79:
```suggestion
    anim.finished.then(() => this._closeOnTransitionEnd());
```

Using the promise makes it clearer that we only expect this to happen once. Also, it avoids passing an event into a function that doesn't expect it. https://jakearchibald.com/2021/function-callback-risks/

src/client/lazy-app/Modal/index.tsx
```diff
@@ -0,0 +1,120 @@
...
+          duration: 250,
+          easing: 'ease',
+          pseudoElement: '::backdrop',
+        },
+      );
+    } catch (e) {}
+  }
+
+  private _hideModal() {
+    if (!this.dialogElement) throw Error('Modal missing / hidden');
```
jakearchibald, line 61: I think it's ok to assume `dialogElement` is there, since this component never presents a state where it isn't. Other instances of this can be skipped too.

src/client/lazy-app/Modal/ModalHint/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+import Modal from '..';
+import { InfoIcon } from 'client/lazy-app/icons';
+import { linkRef } from 'shared/prerendered-app/util';
+
+import * as style from './style.css';
+import 'add-css:./style.css';
+
+interface Props {
+  modalTitle: string;
+  text?: string;
```
jakearchibald, line 11: Let's change this to "label" or something that better describes its usage.

src/client/lazy-app/Modal/ModalHint/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+          event.preventDefault();
+          event.stopImmediatePropagation();
+        }}
+      >
+        <button
+          class={style.modalButton}
+          onClick={this.onclick}
+          title="Learn more"
+        >
+          <InfoIcon></InfoIcon>
```
jakearchibald, line 45:
```suggestion
          <InfoIcon/>
```

src/client/lazy-app/Modal/ModalHint/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+  render({ modalTitle, text }: Props) {
+    return (
+      <span
+        class={style.modalHint}
+        onClick={(event) => {
+          // When the button is clicked, the event starts bubbling up
+          // which might cause unexpected behaviour
+          event.preventDefault();
+          event.stopImmediatePropagation();
```
jakearchibald, line 37: I don't think `stopImmediatePropagation` is needed.

src/client/lazy-app/Modal/ModalHint/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+}
+
+interface State {}
+
+export default class ModalHint extends Component<Props, State> {
+  private modalComponent?: Modal;
+
+  private onclick = (event: Event) => {
+    if (!this.modalComponent)
+      throw new Error('ModalHint is missing a modalComponent');
```
jakearchibald, line 21: This can probably be skipped, as it's guaranteed to be there.

jakearchibald (changes requested):
Coming along nicely!

argyleink:
> UX question: Right now the whole label becomes an activator for the modal (see "Auto sample chroma" in the preview). My hunch is that only the icon should activate the modal, and the rest of the label should remain as an activator for the checkbox. Can I get a second opinion?

Confirm, label clicks are typically activators for an associated checkbox, plus this is also how Squoosh works today, using labels as activators.

I agree with Jake, the icon should isolate the interaction to learn more from a modal, both because users expect labels to activate checkboxes on the web, but also because that's the UX that users currently expect from the label interactions.

Same goes for the nested link in the mozJPEG quality modal, a link should be a link and a button to invoke a modal should be a button to invoke a modal.

Button was a good element choice to wrap the icon, as this keeps it in keyboard tab flow 👍🏻

src/client/lazy-app/icons/index.tsx
aryanpingle: Nice catch with the `id`! I guess I've never really considered when the downloaded SVG has attributes that might interfere with the code, I'll keep that in mind

PR #1374, opened 2023-07-16
[DRAFT] progressive decode

### LINKED ISSUE #1373: Visualize progressive decoding by feeding partial data to decompressor

It would be interesting to play with progressive decoding by adding a slider to control the percentage, or absolute number of kB of an encoded file to send to the decoder. The user could see how much of the image could be rendered with a partial file.

A slider at the bottom of the viewer controls how much data is sent to the image decoder. The decoder is instructed to return what image data it can even if the stream is incomplete.

dholth:
WIP fix #1373 

TODO

- [x] User interface to choose how much data to send to decoder.


jakearchibald:
It seems reasonable for this to be a JXL-only feature to begin with. Ideally it'd be something that worked across all codecs, where you're given a single `<input type="range">` where the left side is zero bytes, and the right is the number of bytes in the larger of the two output images, but that's a much harder problem to solve.

dholth:
It's far enough along to display partially decoded images and get an idea of how they work in jpegxl, but the partial decoding is just attached to the percent on-change function and is not well integrated with the post-compression decoding system. It looks like the browser's built-in decode doesn't display partial images at all, e.g. for JPEG or for Firefox Nightly's built-in jpeg-xl; the wasm encoder should always be used even if the browser supports that codec.

The images I've been testing seem to display after a minimum of 20-40% or the first 100-200kB of image data is available. First you will see a lower resolution version of the whole image. Then higher resolution blocks are filled in from left-to-right, top-to-bottom, and the effect is subtle. Squoosh app doesn't expose the compressor's block order parameters like "spiral out from the center or from a point of interest". Some level of progressive loading works even when progressive is not checked. Progressive images often compress better than non-progressive. Interesting experiment over all.

PR #1381, opened 2023-09-15
Update libavif

aryanpingle:
* Update libavif for improved compression and speed
* v1.0.0 deprecates usage of min and max-quantizers; we use `quality` and `qualityAlpha` instead
* Renamed `maxSpeed` to `MAX_EFFORT` for better readability


src/features/encoders/avif/client/index.tsx
```diff
@@ -120,16 +127,16 @@ export class Options extends Component<Props, State> {
         };
 
         const newOptions: EncodeOptions = {
-          cqLevel: optionState.lossless ? 0 : maxQuant - optionState.quality,
+          cqLevel: optionState.lossless ? 100 : optionState.quality,
```
surma: Can you move this back into a global variable `maxQuality` or something?

src/features/encoders/avif/client/index.tsx
```diff
@@ -183,7 +190,7 @@ export class Options extends Component<Props, State> {
             <div class={style.optionOneCell}>
               <Range
                 min="0"
-                max="63"
+                max="99"
```
surma: I know this wasn’t the case previously, but it really should have been 😅 

```suggestion
                max={maxQuality - 1}
```

Same below!

surma (approved):
Minor nit to avoid magic numbers, otherwise LGTM!

src/features/encoders/avif/client/index.tsx
aryanpingle: Yep that'd be easier to read, on it!

wantehchang (commented):
Hi Aryan,

I will review this pull request tomorrow. Please wait for my review.

codecs/avif/Makefile
```diff
@@ -1,6 +1,6 @@
 # libavif and libaom versions are from
 # google3/third_party/libavif/METADATA
-CODEC_URL = https://github.com/AOMediaCodec/libavif/archive/647c3c208cf152395d777c1bf7240d2ecf7df5a9.tar.gz
+CODEC_URL = https://github.com/AOMediaCodec/libavif/archive/d1c26facaf5a8a97919ceee06814d05d10e25622.tar.gz
```
wantehchang: 1. Remove or update the comments at lines 1-2. You are now using official libavif and libaom releases, rather than the stable snapshots in google3/third_party/libavif/METADATA and google3/third_party/libaom/METADATA

2. Consider adding a comment to note that this hash is v1.0.1.

3. The latest libaom release is v3.7.0. So you can also update libaom to v3.7.0 (line 6) in a separate CL.

codecs/avif/enc/avif_enc.cpp
```diff
@@ -6,9 +6,9 @@
 using namespace emscripten;
 
 struct AvifOptions {
-  // [0 - 63]
-  // 0 = lossless
-  // 63 = worst quality
+  // [0 - 100]
+  // 0 = worst quality
+  // 100 = lossless
```
wantehchang, line 23: Please rename the `cqLevel` and `cqAlphaLevel` members to `quality` and `qualityAlpha`, respectively.

In the comment at line 13, change "cqLevel" to "quality".

wantehchang, line 23: Change lines 61-63 from:

```
  bool lossless = options.cqLevel == AVIF_QUANTIZER_LOSSLESS &&
                  options.cqAlphaLevel <= AVIF_QUANTIZER_LOSSLESS &&
                  format == AVIF_PIXEL_FORMAT_YUV444;
```

to

```
  bool lossless = options.quality == AVIF_QUALITY_LOSSLESS &&
                  (options.qualityAlpha == -1 || options.qualityAlpha == AVIF_QUALITY_LOSSLESS) &&
                  format == AVIF_PIXEL_FORMAT_YUV444;
```

wantehchang, line 23: At line 65, check if the `avifImageCreate()` call returns a null pointer?

wantehchang, line 23: At line 79, check the return value of `avifImageRGBToYUV()`.

codecs/avif/enc/avif_enc.cpp
```diff
@@ -81,23 +81,16 @@ val encode(std::string buffer, int width, int height, AvifOptions options) {
...
-    encoder->minQuantizer = AVIF_QUANTIZER_BEST_QUALITY;
-    encoder->maxQuantizer = AVIF_QUANTIZER_WORST_QUALITY;
-    encoder->minQuantizerAlpha = AVIF_QUANTIZER_BEST_QUALITY;
-    encoder->maxQuantizerAlpha = AVIF_QUANTIZER_WORST_QUALITY;
     avifEncoderSetCodecSpecificOption(encoder, "end-usage", "q");
-    avifEncoderSetCodecSpecificOption(encoder, "cq-level", std::to_string(options.cqLevel).c_str());
     avifEncoderSetCodecSpecificOption(encoder, "sharpness",
                                       std::to_string(options.sharpness).c_str());
 
+    encoder->quality = options.cqLevel;
```
wantehchang: Delete line 87: `avifEncoderSetCodecSpecificOption(encoder, "end-usage", "q");`

This is the default now.

wantehchang: At line 88, check the return value of `avifEncoderSetCodecSpecificOption()`.

wantehchang: Change lines 91-94 from

```
    encoder->quality = options.cqLevel;
    if (options.cqAlphaLevel != -1) {
      encoder->qualityAlpha = options.cqAlphaLevel;
    }
```

to

```
    encoder->quality = options.quality;
    if (options.qualityAlpha == -1) {
      encoder->qualityAlpha = options.quality;
    } else {
      encoder->qualityAlpha = options.qualityAlpha;
    }
```

Note that this code can also be used in the `if (lossless)` case (lines 84-85).

codecs/avif/enc/avif_enc.cpp
```diff
@@ -81,23 +81,16 @@ val encode(std::string buffer, int width, int height, AvifOptions options) {
...
     if (options.cqAlphaLevel != -1) {
-      avifEncoderSetCodecSpecificOption(encoder, "alpha:cq-level",
-                                        std::to_string(options.cqAlphaLevel).c_str());
+      encoder->qualityAlpha = options.cqAlphaLevel;
```
wantehchang: Change line 96 from

```
    if (options.tune == 2 || (options.tune == 0 && options.cqLevel <= 32)) {
```

to

```
    if (options.tune == 2 || (options.tune == 0 && options.quality >= 50)) {
```

wantehchang: Change lines 127-128 from

```
      .field("cqLevel", &AvifOptions::cqLevel)
      .field("cqAlphaLevel", &AvifOptions::cqAlphaLevel)
```
to

```
      .field("quality", &AvifOptions::quality)
      .field("qualityAlpha", &AvifOptions::qualityAlpha)
```

src/features/encoders/avif/shared/meta.ts
```diff
@@ -18,7 +18,7 @@ export const label = 'AVIF';
 export const mimeType = 'image/avif';
 export const extension = 'avif';
 export const defaultOptions: EncodeOptions = {
-  cqLevel: 33,
+  cqLevel: 50,
```
wantehchang: Change lines 21-22 from

```
  cqLevel: 50,
  cqAlphaLevel: -1,
```

to

```
  quality: 50,
  qualityAlpha: -1,
```

src/features/encoders/avif/client/index.tsx
```diff
@@ -228,7 +236,7 @@ export class Options extends Component<Props, State> {
                         <div class={style.optionOneCell}>
                           <Range
                             min="0"
-                            max="63"
+                            max="99"
```
wantehchang: Change `"99"` to `{MAX_QUALITY - 1} // MAX_QUALITY would mean lossless`?

See line 194.

src/features/encoders/avif/client/index.tsx
```diff
@@ -67,18 +75,18 @@ export class Options extends Component<Props, State> {
     return {
       options,
       lossless,
-      quality: maxQuant - cqLevel,
+      quality: cqLevel,
```
wantehchang: Change `cqLevel` to `quality`

src/features/encoders/avif/client/index.tsx
```diff
@@ -67,18 +75,18 @@ export class Options extends Component<Props, State> {
...
       separateAlpha,
-      alphaQuality:
-        maxQuant -
-        (separateAlpha ? options.cqAlphaLevel : defaultOptions.cqLevel),
+      alphaQuality: separateAlpha
+        ? options.cqAlphaLevel
+        : defaultOptions.cqLevel,
```
wantehchang: Change lines 81-82 from

```
        ? options.cqAlphaLevel
        : defaultOptions.cqLevel,
```

to


```
        ? options.qualityAlpha
        : defaultOptions.quality,
```

Note: I don't know the distinction between `options` and `defaultOptions`.

wantehchang: In line 84:

```
        options.subsample === 0 || lossless
```

Is the `options.subsample === 0` check correct? `options.subsample === 0` means 4:0:0 (monochrome). I wonder if the check should be `options.subsample === -1`.

Also, if `lossless` is true, it seems that `subsample` should be set to 3 (4:4:4).

src/features/encoders/avif/client/index.tsx
```diff
@@ -120,16 +128,16 @@ export class Options extends Component<Props, State> {
         };
 
         const newOptions: EncodeOptions = {
-          cqLevel: optionState.lossless ? 0 : maxQuant - optionState.quality,
+          cqLevel: optionState.lossless ? MAX_QUALITY : optionState.quality,
```
wantehchang: Change `cqLevel` to `quality`.

wantehchang: In line 132, change `cqAlphaLevel` to `qualityAlpha`.

src/features/encoders/avif/client/index.tsx
```diff
@@ -120,16 +128,16 @@ export class Options extends Component<Props, State> {
...
           cqAlphaLevel:
             optionState.lossless || !optionState.separateAlpha
-              ? -1
-              : maxQuant - optionState.alphaQuality,
+              ? -1 // default AVIF alphaLevel
```
wantehchang: Change `alphaLevel` to `alphaQuality` or `qualityAlpha`?

wantehchang (commented):
Hi Aryan,

I reviewed all the files that look human readable. For avif_enc.cpp, I read the entire file.

I suggest some changes. Note that where I suggest the name `qualityAlpha`, you may replace it with `alphaQuality`. (`qualityAlpha` is the name used in avif/avif.h.)

aryanpingle:
@wantehchang I'll go through your reviews now, thanks for replying so quickly 😄

codecs/avif/Makefile
aryanpingle: Sure thing. I saw [this line](https://github.com/AOMediaCodec/libavif/blob/fc592f01f88acdf44bc7019cc78c850485120f38/CHANGELOG.md?plain=1#L128) in the changelogs that mentioned libaom version 3.6.1, so how tightly coupled are libavif and libaom releases? (just for reference)

codecs/avif/enc/avif_enc.cpp
aryanpingle: Ooh nice catch, I'm on it 👍

src/features/encoders/avif/client/index.tsx
aryanpingle: `options` is the user-selected codec parameters, `defaultOptions` is just the default parameters for the codec when it is selected for the first time. I'll change the field names and make sure they match correctly.

aryanpingle: Hmm, it turns out we aren't even exposing options 0 and 2 in the form. My guess is it handles the default value (0) before the user selects anything? I'll look into it

codecs/avif/Makefile
wantehchang: libaom 3.7.0 was released three days after libavif 1.0.0 was released. This is why you saw "Update aom.cmd: v3.6.1" in the changelog of libavif 1.0.0.

libavif and libaom releases are not coupled. libaom 3.7.0 is compatible with earlier libaom 3.x releases, so you can safely update to libaom 3.7.0. (And this is true for future libaom 3.x releases.)

src/features/encoders/avif/client/index.tsx
wantehchang: Thanks for the explanation. If `separateAlpha` is false, why do we use `defaultOptions.cqLevel` instead of `options.cqLevel`? The latter seems more reasonable to me.

src/features/encoders/avif/client/index.tsx
```diff
@@ -40,8 +40,16 @@ interface State {
...
+/**
+ * AVIF quality ranges from 0 (worst) to 100 (lossless).
+ * Since lossless is a separate checkbox, we cap user-inputted quality at 99
+ *
+ * AVIF speed ranges from 0 (slowest) to 10 (fastest).
+ * We display it as 'effort' to the user since it conveys the speed-size tradeoff
+ * much better: speed = 10 - effort
+ */
+const MAX_QUALITY = 100;
+const MAX_EFFORT = 10;
```
wantehchang, line 52: Change lines 65-68 from

```
    const lossless =
      options.cqLevel === 0 &&
      options.cqAlphaLevel <= 0 &&
      options.subsample == 3;
```

to

```
    const lossless =
      options.quality === 100 &&
      (options.qualityAlpha == -1 || options.qualityAlpha == 100) &&
      options.subsample == 3;
```

src/features/encoders/avif/client/index.tsx
wantehchang: I wonder if the default value of `options.subsample` should be something other than 0 (e.g., 1 = 4:2:0), or an invalid value such as -1 to indicate that it is not initialized.

aryanpingle: Yep sounds good to me too, I'll make the changes

codecs/avif/enc/avif_enc.cpp
aryanpingle: Sure, and on a related note is there any documentation for the possible key-value pairs for `avifEncoderSetCodecSpecificOption`? The inline documentation says:

> See the codec documentation to know if a setting is persistent or applied only to the next frame.

but I couldn't find any references.

wantehchang: Here "codec" refers to the AV1 codec that libavif is configured to use. In Squoosh, libavif is configured to use only libaom as the AV1 codec. So "See the codec documentation" means "See the libaom documentation".

Now comes the hard part. The libaom documntation for the possible key-value pairs for `avifEncoderSetCodecSpecificOption` is best obtained in one of two ways:
- Read the source code in the libaom source file av1/arg_defs.c. This is the definitive way.
- Run the libaom command-line program `aomenc --help`. This is more "user friendly", but has the drawback that `aomenc` has some additional command-line options that are not implemented using libaom's key-value API, so not every option in the `aomenc --help` page can be passed to `avifEncoderSetCodecSpecificOption`.

My apologies for the current state of libaom's documentation for key-value pairs. Do you have any specific questions about them? I will be happy to find the answers for you (and tell you how I find the answers so you will be able to do it yourself in the future).

aryanpingle: No specific questions so far, though I'm sure I'll have some as I dive deeper into the codec. [The libaom file has pretty good documentation](https://aomedia.googlesource.com/aom/+/refs/heads/main/av1/arg_defs.c#145) so that answers the question!

wantehchang:
Hi Aryan,

I am receiving notifications of changes to this pull request. I am afraid that I may miss something, so please ping me when you'd like me to review this pull request again. Thanks.

aryanpingle:
@wantehchang Yep I'm just resolving conversations as and when I push changes, maybe that's why you're getting so many notifications. My mid-terms just got over so I finally have some time to work on the PR - I'll send you a review request when I finish!

aryanpingle:
@wantehchang thanks for your reviews, I've fixed and refactored the code you suggested. Also, you mentioned on the AV1 Discussion forum:

> Off the top of my head, I think you can expose the AVIF_CHROMA_DOWNSAMPLING_SHARP_YUV option for avifImageRGBToYUV()

I went through the [avif.h documentation](https://github.com/AOMediaCodec/libavif/blob/a5e63209fcf1c43c74b1d4218d4c6dd7fce8e8f2/include/avif/avif.h#L819) for the `avifChromaDownsampling` options, I didn't understand why the automatic / fastest / best quality downsampling options are marked as **(same as AVERAGE)**. Does this mean they all serve the same purpose as of now?

If so, then maybe a checkbox called "Enable sharp YUV chroma downsampling" can be used to toggle between `AVIF_CHROMA_DOWNSAMPLING_AUTOMATIC` and `AVIF_CHROMA_DOWNSAMPLING_SHARP_YUV`? Let me know what you think.

aryanpingle:
@wantehchang I think we got `AVIF_CHROMA_DOWNSAMPLING_SHARP_YUV` working now! [Here's the deploy preview](https://deploy-preview-1381--squoosh.netlify.app/).

codecs/avif/Makefile
```diff
@@ -68,24 +74,52 @@ $(OUT_NODE_DEC_JS) $(OUT_DEC_JS): $(OUT_DEC_CPP) $(CODEC_DIR)/CMakeLists.txt $(L
...
+	mkdir -p $(LIBWEBP_DIR)
+	curl -sL $(LIBWEBP_URL_WITH_SHARPYUV) \
+		| tar xzm -C $(LIBWEBP_DIR)
+
+# Make libsharpyuv.a
+$(LIBSHARPYUV): $(LIBWEBP_DIR)/CMakeLists.txt
+	mkdir -p $(@D)
+	emcmake cmake \
+		-DBUILD_SHARED_LIBS=OFF \
+		-DCMAKE_BUILD_TYPE=Release \
```
wantehchang, line 124: Optional: My colleague Vincent Rabaud recently added more cmake flags to speed up the build (by not generating the unneeded targets):

```
-DWEBP_BUILD_ANIM_UTILS=OFF -DWEBP_BUILD_CWEBP=OFF -DWEBP_BUILD_DWEBP=OFF -DWEBP_BUILD_GIF2WEBP=OFF -DWEBP_BUILD_IMG2WEBP=OFF -DWEBP_BUILD_VWEBP=OFF -DWEBP_BUILD_WEBPINFO=OFF -DWEBP_BUILD_LIBWEBPMUX=OFF -DWEBP_BUILD_WEBPMUX=OFF -DWEBP_BUILD_EXTRAS=OFF
```

codecs/avif/Makefile
```diff
@@ -68,24 +74,52 @@ $(OUT_NODE_DEC_JS) $(OUT_DEC_JS): $(OUT_DEC_CPP) $(CODEC_DIR)/CMakeLists.txt $(L
 			-DCONFIG_MULTITHREAD=0 \
 		" \
 		ENVIRONMENT=$(ENVIRONMENT) \
-		LIBAVIF_FLAGS="-DAVIF_CODEC_AOM_ENCODE=0"
+		LIBAVIF_FLAGS="-DAVIF_CODEC_AOM_ENCODE=0 -DAVIF_LOCAL_LIBSHARPYUV=ON"
 
-$(CODEC_PACKAGE):
-	mkdir -p $(@D)
-	curl -sL $(CODEC_URL) -o $@
+# \
```
wantehchang: Nit: Delete the backslash `\` ?

Also in line 93.

codecs/avif/Makefile
```diff
@@ -28,10 +33,10 @@ HELPER_MAKEFLAGS := -f helper.Makefile
 
 .PHONY: all clean
 
-all: $(OUT_ENC_JS) $(OUT_DEC_JS) $(OUT_ENC_MT_JS) $(OUT_NODE_ENC_JS) $(OUT_NODE_ENC_MT_JS) $(OUT_NODE_DEC_JS)
+all: $(OUT_ENC_JS) $(OUT_DEC_JS) $(OUT_ENC_MT_JS)
 
-$(OUT_NODE_ENC_JS) $(OUT_NODE_ENC_MT_JS): ENVIRONMENT=node
-$(OUT_NODE_ENC_JS) $(OUT_ENC_JS): $(OUT_ENC_CPP) $(CODEC_DIR)/CMakeLists.txt $(LIBAOM_DIR)/CMakeLists.txt
+# ST-Encoding
+$(OUT_ENC_JS): $(OUT_ENC_CPP) $(CODEC_DIR)/CMakeLists.txt $(LIBAOM_DIR)/CMakeLists.txt $(LIBSHARPYUV)
```
wantehchang, line 39: I noticed you pass `-DCONFIG_AV1_HIGHBITDEPTH=0` to libaom (e.g., in line 47). Do you only support bit depth 8 when encoding AVIF?

codecs/avif/enc/avif_enc.cpp
```diff
@@ -3,15 +3,22 @@
 #include <emscripten/val.h>
 #include "avif/avif.h"
 
+#define RETURN_NULL_IF_NOT_EQUALS(val1, val2) \
+  if (val1 != val2)                           \
+    return val::null();
+#define RETURN_NULL_IF_EQUALS(val1, val2) \
+  if (val1 == val2)                       \
+    return val::null();
```
wantehchang: Remove the semicolon (;) from these two macros (line 8 and line 11), otherwise you have an extra semicolon when you use these macros like this:

```
    RETURN_NULL_IF_NOT_EQUALS(status, AVIF_RESULT_OK);
```

Note: The best way to define this kind of macro is to put them in a do {} while (false) block:

```
#define RETURN_NULL_IF_NOT_EQUALS(val1, val2) \
  do { \
    if (val1 != val2)                           \
      return val::null(); \
  } while (false)
```

codecs/avif/enc/avif_enc.cpp
```diff
@@ -58,11 +69,12 @@ val encode(std::string buffer, int width, int height, AvifOptions options) {
...
   }
 
-  bool lossless = options.cqLevel == AVIF_QUANTIZER_LOSSLESS &&
-                  options.cqAlphaLevel <= AVIF_QUANTIZER_LOSSLESS &&
+  bool lossless = options.quality == AVIF_QUALITY_LOSSLESS &&
+                  (options.qualityAlpha == -1 || options.qualityAlpha == AVIF_QUALITY_LOSSLESS) &&
                   format == AVIF_PIXEL_FORMAT_YUV444;
 
   avifImage* image = avifImageCreate(width, height, depth, format);
+  RETURN_NULL_IF_EQUALS(image, NULL);
```
wantehchang: Nit: Use `nullptr` instead of `NULL`? Also in line 101.

codecs/avif/enc/avif_enc.cpp
```diff
@@ -35,11 +42,15 @@ struct AvifOptions {
   int tune;
   // 0-50
   int denoiseLevel;
+  // toggles AVIF_CHROMA_DOWNSAMPLING_SHARP_YUV
+  bool enableSharpDownsampling;
```
wantehchang: enableSharpDownsampling => enableSharpYUVDownsampling

codecs/avif/enc/avif_enc.cpp
```diff
@@ -76,40 +88,48 @@ val encode(std::string buffer, int width, int height, AvifOptions options) {
...
   srcRGB.pixels = rgba;
   srcRGB.rowBytes = width * 4;
-  avifImageRGBToYUV(image, &srcRGB);
+  if (options.enableSharpDownsampling) {
+    srcRGB.chromaDownsampling = AVIF_CHROMA_DOWNSAMPLING_SHARP_YUV;
+  }
+  status = avifImageRGBToYUV(image, &srcRGB);
+  if (status == AVIF_RESULT_NOT_IMPLEMENTED) {
+    printf("libsharpyuv not implemented methinks\n");
+  }
```
wantehchang: Delete lines 95-97? This looks like a debugging printf.

codecs/avif/enc/avif_enc.cpp
```diff
@@ -76,40 +88,48 @@ val encode(std::string buffer, int width, int height, AvifOptions options) {
...
-    avifEncoderSetCodecSpecificOption(encoder, "end-usage", "q");
-    avifEncoderSetCodecSpecificOption(encoder, "cq-level", std::to_string(options.cqLevel).c_str());
-    avifEncoderSetCodecSpecificOption(encoder, "sharpness",
-                                      std::to_string(options.sharpness).c_str());
-
-    if (options.cqAlphaLevel != -1) {
-      avifEncoderSetCodecSpecificOption(encoder, "alpha:cq-level",
-                                        std::to_string(options.cqAlphaLevel).c_str());
+    status = avifEncoderSetCodecSpecificOption(encoder, "sharpness",
+                                               std::to_string(options.sharpness).c_str());
```
wantehchang, line 111: Nit: Add `#include <string>` at the beginning of this file for `std::to_string`.

codecs/avif/enc/avif_enc.cpp
```diff
@@ -76,40 +88,48 @@ val encode(std::string buffer, int width, int height, AvifOptions options) {
...
-    if (options.tune == 2 || (options.tune == 0 && options.cqLevel <= 32)) {
-      avifEncoderSetCodecSpecificOption(encoder, "tune", "ssim");
+    if (options.tune == 2 || (options.tune == 0 && options.quality >= 50)) {
+      status = avifEncoderSetCodecSpecificOption(encoder, "tune", "ssim");
+      RETURN_NULL_IF_NOT_EQUALS(status, AVIF_RESULT_OK);
     }
 
     if (options.chromaDeltaQ) {
-      avifEncoderSetCodecSpecificOption(encoder, "enable-chroma-deltaq", "1");
+      status = avifEncoderSetCodecSpecificOption(encoder, "enable-chroma-deltaq", "1");
```
wantehchang: Nit: We should also prefix this with `color:` as you did in line 130. The alpha channel in AVIF is represented by a separate monochrome image, which doesn't have chroma planes.

src/features/encoders/avif/client/index.tsx
```diff
@@ -38,10 +38,19 @@ interface State {
   denoiseLevel: number;
   aqMode: number;
   tune: AVIFTune;
+  enableSharpDownsampling: boolean;
```
wantehchang: `aqMode` (line 39) doesn't seem to be used in codecs/avif/enc/avif_enc.cpp. Please double check.

wantehchang: Nit: Add "YUV" to the name of this boolean: enableSharpYUVDownsampling

src/features/encoders/avif/client/index.tsx
```diff
@@ -55,34 +64,30 @@ export class Options extends Component<Props, State> {
...
-      options.cqAlphaLevel <= 0 &&
+      options.quality === MAX_QUALITY &&
+      (options.qualityAlpha == -1 || options.qualityAlpha == MAX_QUALITY) &&
       options.subsample == 3;
 
-    const separateAlpha = options.cqAlphaLevel !== -1;
+    const separateAlpha = options.qualityAlpha !== -1;
 
-    const cqLevel = lossless ? defaultOptions.cqLevel : options.cqLevel;
+    const quality = lossless ? defaultOptions.quality : options.quality;
```
wantehchang: I may have asked you about this line before, but I still don't understand this line (specifically, why do we use `defaultOptions.quality` if `lossless` is true). It would be good to add a comment.

Also, `defaultOptions.quality` seems to be 50 (see src/features/encoders/avif/shared/meta.ts). The quality for lossless should be 100, not 50.

src/features/encoders/avif/client/index.tsx
```diff
@@ -120,20 +125,21 @@ export class Options extends Component<Props, State> {
...
 
         const newOptions: EncodeOptions = {
-          cqLevel: optionState.lossless ? 0 : maxQuant - optionState.quality,
-          cqAlphaLevel:
+          quality: optionState.lossless ? MAX_QUALITY : optionState.quality,
+          qualityAlpha:
             optionState.lossless || !optionState.separateAlpha
-              ? -1
-              : maxQuant - optionState.alphaQuality,
+              ? -1 // default AVIF alphaLevel
```
wantehchang: "alphaLevel" seems like a typo. I think this comment should point out that -1 means "set qualityAlpha to quality".

src/features/encoders/avif/client/index.tsx
```diff
@@ -55,34 +64,30 @@ export class Options extends Component<Props, State> {
...
       separateAlpha,
-      alphaQuality:
-        maxQuant -
-        (separateAlpha ? options.cqAlphaLevel : defaultOptions.cqLevel),
-      subsample:
-        options.subsample === 0 || lossless
-          ? defaultOptions.subsample
-          : options.subsample,
+      alphaQuality: separateAlpha ? options.qualityAlpha : options.quality,
+      subsample: defaultOptions.subsample,
```
wantehchang: Why do we use `defaultOptions` here? We use `options` everywhere else.

wantehchang (commented):
Hi Aryan,

I read avif_enc.cpp carefully. I tried to read index.tsx and meta.ts as much as I could, but I only skimmed through the parts in index.tsx that you didn't modify.

codecs/avif/Makefile
aryanpingle: Sure thing. I'll update this comment with the speed improvements as well.

codecs/avif/Makefile
aryanpingle: Yep, for all codecs on Squoosh.

codecs/avif/enc/avif_enc.cpp
aryanpingle: Ahh gotcha. I suppose without that it would've failed for compound `if` statements.

src/features/encoders/avif/client/index.tsx
aryanpingle: Yep it's an obsolete parameter, removing it.

src/features/encoders/avif/client/index.tsx
aryanpingle: By default, we don't use the lossless mode so we keep the quality at a _reasonable_ 50. Agreed, using `defaultOptions.quality` seems wrong here.

<hr>

@jakearchibald correct me if I'm wrong: There are are two situations where we want to generate the encoding state:

1. When the component is mounted for the first time - we use `getDerivedStateFromProps` to generate the state as a function of the props (which will be `defaultOptions`)
2. When any input changes - we use the `_inputChange` function to set the new encoding state.

So yeah I think we can completely remove `defaultOptions` from this file because it'll be equivalent to using `props.options`. Also, 

```typescript
const quality = lossless ? defaultOptions.quality : options.quality;
```

can be removed? Because `lossless` will be true iff `options.quality` is 100 anyway.

src/features/encoders/avif/client/index.tsx
aryanpingle: Yep it is, I'm on it

codecs/avif/Makefile
aryanpingle: Confirmed, it's around 2x faster with these flags. Adding them now!

codecs/avif/enc/avif_enc.cpp
aryanpingle: I feel the variable name is pretty wordy even now. Changing it to `enableSharpYUV`, that should make it more readable IMO.

src/features/encoders/avif/client/index.tsx
jakearchibald: Remember that this is UI state, not encoder state.

What if:

1. The user changes the quality slider to 30.
2. The user checks the lossless checkbox (removing the quality slider).
3. The user unchecks the lossless checkbox (brining back the quality slider).

What should the quality slider be? 100 or 30?

I feel strongly that it should be 30. The quality slider in the UI is only used for lossy quality control, so it should be returned to the last state the user set it to. They checked "lossless", sure, but I don't think that means the quality slider is set to 100 when the user returns to lossy encoding.

So, if we're going to return it to a value of 30, we need to store that state somewhere, and it's stored in the `quality` UI state.

When deriving the `EncodeOptions`, the `quality` UI state is ignored when the intended encode is lossless (see the code in `_inputChange`).

`getDerivedStateFromProps` performs the initial conversion from encode state to UI state. If the incoming encode options are lossless, then we need to figure out what to set the quality slider to if it's shown, hence:

```ts
const quality = lossless ? defaultOptions.quality : options.quality;
```

This code sets the quality slider to `defaultOptions.quality` if the incoming encode options are lossless. This avoids the quality slider being set to 100, which doesn't make sense as a default for a slider related to lossy quality.

aryanpingle: @jakearchibald  Ah right, if it's lossless then the quality should just be whatever it was before and the encoder should be given `100` without changing the state.

On a related note, I think

```typescript
const quality = lossless ? defaultOptions.quality : options.quality;
```

is redundant anyway because we're using `getDerivedStateFromProps()` only when the AVIF codec is selected, so `options` will be equal to `defaultOptions`. I've removed this line in the latest commit and it still (correctly) preserves the original `quality` before toggling lossless mode.

wantehchang:
Aryan wrote:
> I went through the [avif.h documentation](https://github.com/AOMediaCodec/libavif/blob/a5e63209fcf1c43c74b1d4218d4c6dd7fce8e8f2/include/avif/avif.h#L819) for the `avifChromaDownsampling` options, I didn't understand why the automatic / fastest / best quality downsampling options are marked as **(same as AVERAGE)**. Does this mean they all serve the same purpose as of now?

Some of those "(same as AVERAGE)" should be changed to "(same as AVERAGE in the current implementation)". The reason is that the average filter is the only downsampling filter that libavif uses automatically. The Sharp YUV downsampling filter was added later and we are not sure if it should be one of the automatically-used downsampling filters.

NOTE: AVIF_CHROMA_DOWNSAMPLING_FASTEST will always be the same as AVERAGE.

> If so, then maybe a checkbox called "Enable sharp YUV chroma downsampling" can be used to toggle between `AVIF_CHROMA_DOWNSAMPLING_AUTOMATIC` and `AVIF_CHROMA_DOWNSAMPLING_SHARP_YUV`? Let me know what you think.

Agreed.

codecs/avif/enc/avif_enc.cpp
```diff
@@ -3,15 +3,31 @@
...
+#define RETURN_NULL_IF_NOT_EQUALS(val1, val2) \
+  do {                                        \
+    if (val1 != val2)                         \
+      return val::null();                     \
+  } while (false)
+#define RETURN_NULL_IF_EQUALS(val1, val2) \
+  do {                                    \
+    if (val1 == val2)                     \
+      return val::null();                 \
+  } while (false)
```
wantehchang, line 13: Optional: You can merge these two macros into `RETURN_NULL_IF_FALSE` that takes a conditional expression as the argument.

codecs/avif/Makefile
```diff
@@ -68,24 +74,58 @@ $(OUT_NODE_DEC_JS) $(OUT_DEC_JS): $(OUT_DEC_CPP) $(CODEC_DIR)/CMakeLists.txt $(L
 			-DCONFIG_MULTITHREAD=0 \
 		" \
 		ENVIRONMENT=$(ENVIRONMENT) \
-		LIBAVIF_FLAGS="-DAVIF_CODEC_AOM_ENCODE=0"
+		LIBAVIF_FLAGS="-DAVIF_CODEC_AOM_ENCODE=0 -DAVIF_LOCAL_LIBSHARPYUV=ON"
```
wantehchang: I believe we can omit -DAVIF_LOCAL_LIBSHARPYUV=ON in the decoding-only build. Sharp YUV is only used in the RGB-to-YUV conversion, which is only used when preparing the YUV input to the AVIF encoder.

codecs/avif/enc/avif_enc.cpp
```diff
@@ -73,74 +95,79 @@ val encode(std::string buffer, int width, int height, AvifOptions options) {
...
+    RETURN_NULL_IF_NOT_EQUALS(status, AVIF_RESULT_OK);
   }
 
   encoder->maxThreads = emscripten_num_logical_cores();
   encoder->tileRowsLog2 = options.tileRowsLog2;
   encoder->tileColsLog2 = options.tileColsLog2;
   encoder->speed = options.speed;
 
-  avifResult encodeResult = avifEncoderWrite(encoder, image, &output);
+  avifResult encodeResult = avifEncoderWrite(encoder.get(), image.get(), &output);
```
wantehchang, line 144: Nit: I suggest moving the declaration of `output` (currently at line 63) right before this line.

codecs/avif/enc/avif_enc.cpp
```diff
@@ -3,15 +3,31 @@
 #include <emscripten/val.h>
 #include "avif/avif.h"
 
+#include <string>
```
wantehchang, line 7: Add `#include <memory>` (because we're using `std::unique_ptr` now).

codecs/avif/enc/avif_enc.cpp
```diff
@@ -3,15 +3,31 @@
...
-  // 63 = worst quality
-  int cqLevel;
-  // As above, but -1 means 'use cqLevel'
-  int cqAlphaLevel;
+  // [0 - 100]
+  // 0 = worst quality
+  // 100 = lossless
+  int quality;
+  // As above, but -1 means 'use quality'
+  int qualityAlpha;
```
wantehchang, line 26: In a follow-up pull request, we should look into using the new `autoTiling` option of the AVIF encoder.

I suggest Squoosh set `autoTiling` to `AVIF_TRUE` by default. (The default of `autoTiling` in libavif is `AVIF_FALSE` for backward compatibility.) The Squoosh UI can have a "manual tiling" box, which when checked,  shows the `tileRowsLog2` and `tileColsLog2` fields (both default to 0, meaning no tiling).

codecs/avif/enc/avif_enc.cpp
```diff
@@ -58,11 +78,13 @@ val encode(std::string buffer, int width, int height, AvifOptions options) {
...
   }
 
-  bool lossless = options.cqLevel == AVIF_QUANTIZER_LOSSLESS &&
-                  options.cqAlphaLevel <= AVIF_QUANTIZER_LOSSLESS &&
+  bool lossless = options.quality == AVIF_QUALITY_LOSSLESS &&
+                  (options.qualityAlpha == -1 || options.qualityAlpha == AVIF_QUALITY_LOSSLESS) &&
                   format == AVIF_PIXEL_FORMAT_YUV444;
 
-  avifImage* image = avifImageCreate(width, height, depth, format);
+  // Smart pointer for the generated image
```
wantehchang: Nit: "generated image" isn't quite accurate. This is the input image, in the YUV format.

wantehchang (approved):
LGTM! Just a few final minor issues.

src/features/encoders/avif/client/index.tsx
jakearchibald: Are you sure that still works when copying settings from one side to the other?

jakearchibald: Compare these steps on this deploy vs live:

1. Open demo image
2. On the right-hand side, select AVIF
3. On the right-hand side, select lossless
4. Copy settings to the left side (using the [< >] icon)
5. On the left-hand side, deselect lossless

On squoosh.app this gives you a default quality setting. In this PR it gives you a quality of 100.

I'm pretty sure this is because you removed that line, which would use `defaultOptions.quality` if the initial settings are lossless.

aryanpingle: Ahh I see what you mean. Reverting the change.

codecs/avif/enc/avif_enc.cpp
aryanpingle: That's so much better than the current macro LOL, thank you!

codecs/avif/enc/avif_enc.cpp
aryanpingle: Gotcha, I'll try to add it to this PR or at least make raise an issue for it. I assume `autoTiling` automatically chooses the _best_ `tileRowsLog2` and `tileColsLog2`, is there any more documentation about it that I can link to?

codecs/avif/enc/avif_enc.cpp
wantehchang: Note: I probably would not use a macro that saves only one line. But some people like to use this kind of macro so that the main flow of the code is not cluttered with error handling code.

codecs/avif/enc/avif_enc.cpp
wantehchang: Unless this requires just simple changes, I suggest filing an issue for this and implementing this in a follow-up pull request. The review of this pull request has been going on for a while. We need to watch out for "code review fatigue."

You understand `autoTiling` correctly. The `autoTiling` algorithm may be tweaked in the future, so we didn't document the current algorithm in the public header avif/avif.h. You can find the current algorithm in libavif/src/write.c:

```
    if (encoder->autoTiling) {
        // Use as many tiles as allowed by the minimum tile area requirement and impose a maximum
        // of 8 tiles.
        const int threads = 8;
        avifSetTileConfiguration(threads, firstCell->width, firstCell->height, &encoder->data->tileRowsLog2, &encoder->data->tileColsLog2);
    }
```

I do think all the automatic tiling algorithms have some common elements, such as imposing a minimum tile area or tile width and height, and potentially imposing a maximum number of tiles. But I am not sure if you need to mention this. You can, however, say that the use of tiles makes it easier for the decoder to decode an AVIF image with multiple threads in parallel.

codecs/avif/enc/avif_enc.cpp
aryanpingle: Yep that was my line of thinking. I'm pretty sure we'll be repeating that line again for other optimizations, so it's better to just have a macro for it.

aryanpingle:
@wantehchang implemented the changes, please give it a once-over if you find the time.
Totally agree about the code review fatigue so I've made a new issue referencing that comment thread.

aryanpingle:
Enabling the `ERROR_ON_UNDEFINED_SYMBOLS=0` flag in the Makefile was a hacky solution to get rid of this error:

```console
error: undefined symbol: setjmp (referenced by top-level compiled C/C++ code)
warning: Link with `-s LLD_REPORT_UNDEFINED` to get more information on undefined symbols
warning: To disable errors for undefined symbols use `-s ERROR_ON_UNDEFINED_SYMBOLS=0`
warning: _setjmp may need to be added to EXPORTED_FUNCTIONS if it arrives from a system library
```

It might be a compiler bug.
Regardless, @surma found that bumping the Emscripten version to `3.1.47` fixes this issue. Might be worth updating the version for all codec binaries.

wantehchang (approved):
LGTM!

z11h:
any idea about when this will get merged and pushed to squoosh.app? thanks!

grandpa1946:
> any idea about when this will get merged and pushed to squoosh.app? thanks!

you can always check at the bottom of the page where it says "changes approved" & you should click on "Details" to the right of deploy/netlify

jakearchibald:
@wantehchang it won't block this PR, but is the experimental progressive stuff in `avifenc` worth exposing? Also, is there any compression benefit to premultiplying alpha, or is it just an input option?

wantehchang:
@jakearchibald Hi Jake,

You can expose the encoding of progressive (layered) images if it is clearly marked as experimental in the UI.

There could be compression benefit to premultiplying alpha (see Note below). But the main benefit that after the image is decoded, we don't need to multiply it with alpha. We just need to multiply the backgroud with 1 - alpha. So the decoder has less work to do.

Note: If the alpha channel has a lot of zero or near-zero values, the alpha-premultiplied image will have a lot of zero or near-zero values, so it should result in a smaller compressed file.

jamsinclair:
Seems like single threaded AVIF encoding may have broken with this change. From my limited debugging it looks like a memory issue. The output handle seems to be no longer available when passing the value back to JS? 🤔 

```
TypeError: emval_handle_array[handle] is undefined
```

jakearchibald:
@jamsinclair thank you. We'll look into it.

jakearchibald:
@wantehchang Thanks for the details! Do any browsers support rendering of progressive AVIF? (as in, rendering before the file has fully downloaded)

wantehchang:
@jakearchibald Jake: Chrome and Microsoft Edge (Edge 121 supports AVIF) support progressive rendering of AVIF layered images.

Firefox doesn't support progressive rendering of AVIF layered images, based on inspection of its source code (by searching for "layer" and "progressive" in image/decoders/nsAVIFDecoder.{h,cpp}).

I will get back to you on Safari's support.

leo-barnes:
> @jakearchibald Jake: Chrome and Microsoft Edge (Edge 121 supports AVIF) support progressive rendering of AVIF layered images.
> 
> Firefox doesn't support progressive rendering of AVIF layered images, based on inspection of its source code (by searching for "layer" and "progressive" in image/decoders/nsAVIFDecoder.{h,cpp}).
> 
> I will get back to you on Safari's support.

It should be supported in Safari to the best of my knowledge.

jakearchibald:
Chrome support at least makes it interesting 😀

PR #1384, opened 2023-10-16
Adding the Quite OK Image (QOI) Codec

aryanpingle:
Adds the [QOI (Quite OK Image Format) codec](https://github.com/phoboslab/qoi) to Squoosh.

QOI is a lossless, super-fast codec that is stupidly simple to understand and implement. It's written by a single individual, yet manages to hold its own against giants like PNG in terms of compression ratios.
This PR adds the original (and final) version of QOI, but there are more performant variants of QOI out there - I highly encourage future contributors to try adding them to Squoosh!

IMO this is a great way to showcase how easy it is to add codecs to the codebase, without having to worry about platforms and compatibility.


codecs/qoi/enc/qoi_enc.cpp
```diff
@@ -0,0 +1,34 @@
...
+struct QoiOptions {};
+
+thread_local const val Uint8Array = val::global("Uint8Array");
+
+val encode(std::string buffer, int width, int height, QoiOptions options) {
+  int compressedSizeInBytes;
+  qoi_desc desc;
+  desc.width = width;
+  desc.height = height;
+  desc.channels = 4;
```
surma, line 18: We should make `channels` configurable via `options` (`3` and `4` are valid values)! If there is no transparency (i.e. 3 channels), the resulting image file should get smaller. 

While we are at it, let’s make `colorspace` configurable as well.

codecs/qoi/enc/qoi_enc.cpp
```diff
@@ -0,0 +1,34 @@
...
+  desc.colorspace = QOI_SRGB;
+
+  void* encodedData = qoi_encode(buffer.c_str(), &desc, &compressedSizeInBytes);
+  if (encodedData == NULL)
+    return val::null();
+
+  auto js_result =
+      Uint8Array.new_(typed_memory_view(compressedSizeInBytes, (const uint8_t*)encodedData));
```
surma, line 26: This is leaking memory. From the `qoi.h::qoi_encode()` docs:

> The returned qoi data should be free()d after use.

It’s not a big deal as we are killing the Wasm modules quite regularly, but if someone is encoding really big images in fast succession, it could lead to an OOM error.

```suggestion
      Uint8Array.new_(typed_memory_view(compressedSizeInBytes, (const uint8_t*)encodedData));
      free(encodedData);
```

codecs/qoi/dec/qoi_dec.cpp
```diff
@@ -0,0 +1,29 @@
...
+  qoi_desc desc;
+  uint8_t* rgba = (uint8_t*)qoi_decode(qoiimage.c_str(), qoiimage.length(), &desc, 4);
+
+  // Resultant width and height stored in descriptor
+  int decodedWidth = desc.width;
+  int decodedHeight = desc.height;
+
+  val result = ImageData.new_(
+      Uint8ClampedArray.new_(typed_memory_view(4 * decodedWidth * decodedHeight, rgba)),
+      decodedWidth, decodedHeight);
```
surma, line 22: Memory leak here! See comment in encoder for explanation.
```suggestion
      decodedWidth, decodedHeight);
   free(rgba);
```

surma (commented):
Really nicely done!

I think we should expose the two options that the codec offers, and then this is good to go!

codecs/qoi/enc/qoi_enc.cpp
surma: As @jakearchibald said on chat: It’d be idea if the code would automatically use 3-channel mode if there are no transparent pixels.

codecs/qoi/enc/qoi_enc.cpp
aryanpingle: Ohh, nice catch! Committing your suggestion directly 👍

codecs/qoi/dec/qoi_dec.cpp
aryanpingle: Nice catch again, committing this directly

codecs/qoi/enc/qoi_enc.cpp
aryanpingle: Turns out a side-effect of QOI's simplicity is that it can't encode RGB data differently from RGBA. Pretty much just sets the alpha to `255` and the source image's channels matter only during decoding. So yeah, using only 3 channels ends up with the same compressed size as if you use all 4 channels (for opaque images).

surma: You are right. I read trhe source wrong. Ignore me ;D

surma (approved):

PR #1396, opened 2024-01-07
Update oxipng to v9 [updated]

andrews05:
I've attempted to update to oxipng to v9, but the parallel build doesn't seem to work - the worker seems to hang when encoding. When I changed the encoder to always `initST()` it all worked fine.
Not sure how to debug it from here. I've spent hours just getting the thing to build correctly 😩. (Not currently possible on Apple Silicon, it seems).

There's also the new raw API in v9 which may be of use here, but I'll probably leave that for someone else 😆


andrews05:
Fixed the parallel build by updating wasm-pack. It's all working now!

andrews05:
Another update: I've reworked it to use the raw api, so it doesn't need to convert to png first.

jakearchibald:
@surma this looks could. Could you give the rust bits a review?

codecs/oxipng/src/lib.rs
```diff
@@ -1,20 +1,27 @@
...
-    options.alphas.insert(AlphaOptim::Right);
-    options.interlace = Some(if interlace { 1 } else { 0 });
+    options.optimize_alpha = true;
+    options.interlace = Some(if interlace {
+        Interlacing::Adam7
+    } else {
+        Interlacing::None
+    });
 
-    options.deflate = oxipng::Deflaters::Libdeflater;
```
surma, line 18: Oh I just saw that Oxipng now also offers Zopfli. Coo. We should expose that at some point.

surma (approved):
Confirm working. Great stuff!!

codecs/oxipng/src/lib.rs
jakearchibald: We absolutely should

jakearchibald:
The demo Squoosh logo, with 256 colors, no dithering, goes from 17.7kb to 17.2kb with this change. Nice.

andrews05:
Awesome, thanks guys!

jakearchibald:
Deployed! Thanks for the contribution. Excellent improvement.

PR #1409, opened 2024-04-05
Fix single threaded avif encoding 

### LINKED ISSUE #1397: AVIF encoding is broken on browsers without multi-threaded support
**Describe the bug**
When the browser does not support SharedArrayBuffer/Multi-threading, Squoosh fails to encode AVIF images.

**To Reproduce**
1. Clone the repository
1. Remove the `Cross-Origin-Opener-Policy` and `Cross-Origin-Embedder-Policy` to force the dev environment to use the single threaded logic
1. Run `npm run dev`
1. Open the dev page at http://localhost:5000
1. Upload an image. Choose the `AVIF` compress option.
1. Observe the following error is thrown in the console and the image is not encoded to AVIF
```
Uncaught (in promise) RuntimeError: indirect call to null
    invoke_iiiiiiiiii http://localhost:5000/c/avif_enc-8f7d01d5.js:10
    invoke_iii http://localhost:5000/c/avif_enc-8f7d01d5.js:10
    encode http://localhost:5000/c/avif_enc-8f7d01d5.js line 10 > Function:11
    encode http://localhost:5000/c/features-worker-2348e279.js:276
```

**Expected behavior**
The image would still be encoded to AVIF with single threaded browser.

**Version:**
 - OS w/ version: MacOS Sonoma 14.1.1
 - Browser w/ version: Chrome 120.0.6099.234
 - Node version: 16 LTS
 - npm version: 8.14.0

**Additional context, screenshots, screencasts**
This issue was introduced in the recent update of the AVIF version #1381

jamsinclair:
Fixes #1397 

As discussed in the issue above, we need to compile a copy of `libsharpyuv` with threads disabled for the single threaded codec.

I am not that familiar with makefiles or C/C++ compilation. Please feel free to make relevant changes to improve this 🙏 

As always, big thanks to the creators/maintainers of Squoosh 🙇‍♂️


codecs/avif/Makefile
```diff
@@ -106,8 +110,29 @@ $(LIBWEBP_DIR)/CMakeLists.txt: $(CODEC_DIR)/CMakeLists.txt
 	curl -sL $(LIBWEBP_URL_WITH_SHARPYUV) \
 		| tar xzm -C $(LIBWEBP_DIR)
 
-# Make libsharpyuv.a
-$(LIBSHARPYUV): $(LIBWEBP_DIR)/CMakeLists.txt
+# Make libsharpyuv.a for ST-Encoding
+$(LIBSHARPYUV_ST): $(LIBWEBP_DIR)/CMakeLists.txt
+	mkdir -p $(@D)
+	emcmake cmake \
+		-DCMAKE_DISABLE_FIND_PACKAGE_Threads=1 \
```
jamsinclair: I tried using the suggested `-DWEBP_USE_THREAD=OFF`, however, it still resulted in the same error seen in #1397

Any ideas @wantehchang? 🤔

codecs/avif/Makefile
```diff
@@ -36,7 +38,8 @@ HELPER_MAKEFLAGS := -f helper.Makefile
 all: $(OUT_ENC_JS) $(OUT_DEC_JS) $(OUT_ENC_MT_JS)
 
 # ST-Encoding
-$(OUT_ENC_JS): $(OUT_ENC_CPP) $(CODEC_DIR)/CMakeLists.txt $(LIBAOM_DIR)/CMakeLists.txt $(LIBSHARPYUV)
+$(OUT_ENC_JS): $(OUT_ENC_CPP) $(CODEC_DIR)/CMakeLists.txt $(LIBAOM_DIR)/CMakeLists.txt $(LIBSHARPYUV_ST)
+	mkdir -p $(LIBWEBP_DIR)/build && cp $(LIBSHARPYUV_ST) $(LIBSHARPYUV)
```
jamsinclair, line 42: Due to the CMakesList.txt in `libavif` when `AVIF_LOCAL_LIBSHARPYUV` is turned on it has to be located at `/ext/libwebp/build/libsharpyuv.a`

The only workaround I could think of was to copy the relevant file over for each build. Does anyone have a cleaner idea? Or would it be worth requesting an improvement upstream to provide a custom location?

codecs/avif/Makefile
surma: @wantehchang Any downsides in using the flag used here instead of the one you recommended? I had a quick sift through the build files and I suspect that flag isn’t forwarded correctly to the yuv build, but not 100% sure.

codecs/avif/Makefile
surma: @RReverser No idea if you have time for this, but if you do: Any thoughts on this?

codecs/avif/Makefile
wantehchang: Jamie, Surma: I am sorry I forgot to reply. This seems like a bug in `-DWEBP_USE_THREAD=OFF`. I am not familiar with the build system of libwebp (of which libsharpyuv is a component). @jzern James, is `-DWEBP_USE_THREAD=OFF` the right cmake option to disable all threading code (including the dependency on the pthread library) in libwebp?

jzern: > Jamie, Surma: I am sorry I forgot to reply. This seems like a bug in `-DWEBP_USE_THREAD=OFF`. I am not familiar with the build system of libwebp (of which libsharpyuv is a component). @jzern James, is `-DWEBP_USE_THREAD=OFF` the right cmake option to disable all threading code (including the dependency on the pthread library) in libwebp?

Yes. Looking at an x86 build, there's no detection or reference to pthread when `-DWEBP_USE_THREAD=OFF` is set. sharpyuv doesn't create threads, but uses a mutex to guard function pointer initialization.

[CMakeLists.txt](https://chromium.googlesource.com/webm/libwebp/+/refs/tags/v1.4.0-rc1/CMakeLists.txt#84) should set this option to OFF when `-DWEBP_BUILD_WEBP_JS=1` is used. `WEBP_USE_THREAD` is used to control the use of find_package in [deps.cmake](https://chromium.googlesource.com/webm/libwebp/+/refs/tags/v1.4.0-rc1/cmake/deps.cmake#39). I'd be curious to see the `src/webp/config.h` produced and the cmake log.

jzern:
```
$ emcmake cmake .. -DWEBP_BUILD_WEBP_JS=ON
$ emmake make -j
$ grep WEBP_USE_THREAD src/webp/config.h 
/* #undef WEBP_USE_THREAD */
```

So maybe I'm missing something with the build setup in Squoosh.

jzern: > So maybe I'm missing something with the build setup in Squoosh.

I wasn't looking at the diff.

```
$ emcmake cmake .. \
  -DWEBP_BUILD_ANIM_UTILS=OFF \
  -DWEBP_BUILD_CWEBP=OFF \
  -DWEBP_BUILD_DWEBP=OFF \
  -DWEBP_BUILD_GIF2WEBP=OFF \
  -DWEBP_BUILD_IMG2WEBP=OFF \
  -DWEBP_BUILD_VWEBP=OFF \
  -DWEBP_BUILD_WEBPINFO=OFF \
  -DWEBP_BUILD_LIBWEBPMUX=OFF \
  -DWEBP_BUILD_WEBPMUX=OFF \
  -DWEBP_BUILD_EXTRAS=OFF \
  -DBUILD_SHARED_LIBS=OFF \
  -DCMAKE_BUILD_TYPE=Release
...
-- Found Threads: TRUE  
...
$ grep WEBP_USE_THREAD src/webp/config.h
#define WEBP_USE_THREAD 1
```

That's expected.

```
$ emcmake cmake .. \
  -DWEBP_USE_THREAD=OFF \
  -DWEBP_BUILD_ANIM_UTILS=OFF \
  -DWEBP_BUILD_CWEBP=OFF \
  -DWEBP_BUILD_DWEBP=OFF \
  -DWEBP_BUILD_GIF2WEBP=OFF \
  -DWEBP_BUILD_IMG2WEBP=OFF \
  -DWEBP_BUILD_VWEBP=OFF \
  -DWEBP_BUILD_WEBPINFO=OFF \
  -DWEBP_BUILD_LIBWEBPMUX=OFF \
  -DWEBP_BUILD_WEBPMUX=OFF \
  -DWEBP_BUILD_EXTRAS=OFF \
  -DBUILD_SHARED_LIBS=OFF \
  -DCMAKE_BUILD_TYPE=Release
...
# no mention of threads
...
$ grep WEBP_USE_THREAD src/webp/config.h 
/* #undef WEBP_USE_THREAD */
```

codecs/avif/Makefile
```diff
@@ -50,7 +53,8 @@ $(OUT_ENC_JS): $(OUT_ENC_CPP) $(CODEC_DIR)/CMakeLists.txt $(LIBAOM_DIR)/CMakeLis
 		LIBAVIF_FLAGS="-DAVIF_CODEC_AOM_DECODE=0 -DAVIF_LOCAL_LIBSHARPYUV=ON"
 
 # MT-Encoding
-$(OUT_ENC_MT_JS): $(OUT_ENC_CPP) $(CODEC_DIR)/CMakeLists.txt $(LIBAOM_DIR)/CMakeLists.txt $(LIBSHARPYUV)
+$(OUT_ENC_MT_JS): $(OUT_ENC_CPP) $(CODEC_DIR)/CMakeLists.txt $(LIBAOM_DIR)/CMakeLists.txt $(LIBSHARPYUV_MT)
+	mkdir -p $(LIBWEBP_DIR)/build && cp $(LIBSHARPYUV_MT) $(LIBSHARPYUV)
```
jzern, line 58: This looks like it could race with `$(OUT_ENC_JS)` in a multi-job build, unless they're otherwise serialized.

jzern:
Maybe it's unrelated, but even with this change I see AVIF encode errors in a stock Linux Chrome (123.0.6312.105 (Official Build) (64-bit)). WebP and others encode all right.

I rebuilt the avif wasm and the main app with:

```
$ cd codecs/avif
$ npm run build
$ cd ../../
$ npm run build
```

```
Compress-e2686f9a.js:4210 Uncaught (in promise) Error: Encoding error
    at encode (features-worker-83baeae9.js:278:19)
deserialize @ util-06ce6ead.js:59
fromWireValue @ util-06ce6ead.js:265
await in fromWireValue (async)
updateImage @ Compress-e2686f9a.js:4106
(anonymous) @ Compress-e2686f9a.js:3931
setTimeout (async)
queueUpdateImage @ Compress-e2686f9a.js:3931
componentDidUpdate @ Compress-e2686f9a.js:3916
(anonymous) @ VM4:78
(anonymous) @ VM4:78
(anonymous) @ VM4:78
j @ VM4:78
(anonymous) @ VM4:78
g @ VM4:78
handleMouseUp_ @ unknown
Promise.then (async)
k @ VM4:78
d.setState @ VM4:78
Compress.onEncoderTypeChange @ Compress-e2686f9a.js:3730
Options$9.onEncoderTypeChange @ Compress-e2686f9a.js:2973
z @ VM4:78
handleMouseUp_ @ unknown
features-worker-83baeae9.js:92 jxlEncode: 10146.291015625 ms
features-worker-83baeae9.js:92 jxlDecode: 3007.024169921875 ms
```

This matches the behavior with the checked in wasm files.

codecs/avif/Makefile
jamsinclair: Good call. I've added a dependency on `$(OUT_ENC_JS)` to ensure there's no conflicts there.

It's the best we can do for now, unless there's a friendlier way to pass the custom `libsharpyuv` path to the build.

jzern: > Good call. I've added a dependency on $(OUT_ENC_JS) to ensure there's no conflicts there.

It's a minor detail, but an [order-only dependency](https://www.gnu.org/software/make/manual/make.html#Prerequisite-Types) might be enough here `(: ... | $(OUT_ENC_JS)`.

> It's the best we can do for now, unless there's a friendlier way to pass the custom `libsharpyuv` path to the build.

Not without modifying the [LocalLibsharpyuv.cmake](https://github.com/AOMediaCodec/libavif/blob/98ce107ee731a80894e389b26f972e48e82a259d/cmake/Modules/LocalLibsharpyuv.cmake#L3) file.

jamsinclair: Thank you! I wasn't aware of that. I believe that's a nicer solution, I'll update.

codecs/avif/Makefile
jamsinclair: ~Thanks. From what I can gather would it be best to use the flag `-DWEBP_BUILD_WEBP_JS=ON`? That seems to disable the use of the mutex, among other things~

Edit: Oh, perhaps I made a mistake earlier. Sorry for pulling you in, setting the flag `-DWEBP_USE_THREAD=OFF` seems to solve our problems.

jamsinclair:
> Maybe it's unrelated, but even with this change I see AVIF encode errors in a stock Linux Chrome.

@jzern I can't replicate this behaviour. Were you building from the forked branch?

There's a deploy preview generated by the CLI that you can test with
- https://deploy-preview-1409--squoosh.netlify.app/

And I've also hosted a version that will always use the single-threaded variants
- https://deploy-preview-1409--squoosh--single-thread.surge.sh

If you get a chance to try with those, do you get the same result?

codecs/avif/Makefile
RReverser: Hm looking at their CMake, not really, copy or symlink seems like the best option. Often build configs allow to specify custom paths to their dependencies, but apparently not this one. https://chromium.googlesource.com/external/github.com/AOMediaCodec/libavif/+/refs/heads/chromium-m119/CMakeLists.txt#216

jzern:
> > Maybe it's unrelated, but even with this change I see AVIF encode errors in a stock Linux Chrome.
> 
> @jzern I can't replicate this behaviour. Were you building from the forked branch?
> 

I had tried both the tip of tree and this PR, they both had the same issue.

> There's a deploy preview generated by the CLI that you can test with
> 
> * https://deploy-preview-1409--squoosh.netlify.app/
> 

This fails in the same way using any of the sample images.

> And I've also hosted a version that will always use the single-threaded variants
> 
> * https://deploy-preview-1409--squoosh--single-thread.surge.sh
> 

This passes with all of them.

> If you get a chance to try with those, do you get the same result?

jzern:
> > > Maybe it's unrelated, but even with this change I see AVIF encode errors in a stock Linux Chrome.
> > 
> > 
> > @jzern I can't replicate this behaviour. Were you building from the forked branch?
> 
> I had tried both the tip of tree and this PR, they both had the same issue.
> 
> > There's a deploy preview generated by the CLI that you can test with
> > 
> > * https://deploy-preview-1409--squoosh.netlify.app/
> 
> This fails in the same way using any of the sample images.
> 

Works on Windows 10: 123.0.6312.122 (Official Build) (64-bit) (cohort: M123 Rollout).

> > And I've also hosted a version that will always use the single-threaded variants
> > 
> > * https://deploy-preview-1409--squoosh--single-thread.surge.sh
> 
> This passes with all of them.
> 
> > If you get a chance to try with those, do you get the same result?

jzern:
> > > > Maybe it's unrelated, but even with this change I see AVIF encode errors in a stock Linux Chrome.
> > > 
> > > 
> > > @jzern I can't replicate this behaviour. Were you building from the forked branch?
> > 
> > 
> > I had tried both the tip of tree and this PR, they both had the same issue.
> > > There's a deploy preview generated by the CLI that you can test with
> > > 
> > > * https://deploy-preview-1409--squoosh.netlify.app/
> > 
> > 
> > This fails in the same way using any of the sample images.
> 
> Works on Windows 10: 123.0.6312.122 (Official Build) (64-bit) (cohort: M123 Rollout).
> 

macOS: 123.0.6312.123 (Official Build) (arm64)
Linux (intel): 123.0.6312.122 (Official Build) (64-bit)
both pass. The failing machine has a similar version of Chrome and Linux, but is AMD; I'm not sure if that's the issue, though. Let me disable any experiments I'm in there.

> > > And I've also hosted a version that will always use the single-threaded variants
> > > 
> > > * https://deploy-preview-1409--squoosh--single-thread.surge.sh
> > 
> > 
> > This passes with all of them.
> > > If you get a chance to try with those, do you get the same result?

jzern:
> > > > > Maybe it's unrelated, but even with this change I see AVIF encode errors in a stock Linux Chrome.
> > > > 
> > > > 
> > > > @jzern I can't replicate this behaviour. Were you building from the forked branch?
> > > 
> > > 
> > > I had tried both the tip of tree and this PR, they both had the same issue.
> > > > There's a deploy preview generated by the CLI that you can test with
> > > > 
> > > > * https://deploy-preview-1409--squoosh.netlify.app/
> > > 
> > > 
> > > This fails in the same way using any of the sample images.
> > 
> > 
> > Works on Windows 10: 123.0.6312.122 (Official Build) (64-bit) (cohort: M123 Rollout).
> 
> macOS: 123.0.6312.123 (Official Build) (arm64) Linux (intel): 123.0.6312.122 (Official Build) (64-bit) both pass. The failing machine has a similar version of Chrome and Linux, but is AMD; I'm not sure if that's the issue, though. Let me disable any experiments I'm in there.
> 

I don't have any flags set actually. This seems like a separate issue to this PR, so I'll file an issue if I can get more detail.

> > > > And I've also hosted a version that will always use the single-threaded variants
> > > > 
> > > > * https://deploy-preview-1409--squoosh--single-thread.surge.sh
> > > 
> > > 
> > > This passes with all of them.
> > > > If you get a chance to try with those, do you get the same result?

surma (approved):
I think this is good to go. 

@jakearchibald wanna do any checks or reviews? Or :shipit: ?

Branchverse:
I came across the same issue explained in https://github.com/GoogleChromeLabs/squoosh/issues/1397 which is supposed to be solved with this PR, I see that it has been approved.
Is there something missing for the completion since I won't be the only one that tries to solve this for a few hours before realizing there is a fix ready for 4 months?

laoneo:
Thanks!

AshleyScirra:
Thanks for the fix folks! Appreciate it.

PR #1413, opened 2024-04-19
Enable zopfli on oxipng

Csabesz07:
This feature lets the user further optimise the compression of the image by adding new options under oxipng.

- The deflater of oxipng is now selectable (zopfli / libdeflater)
- Both deflater's parameters are modifiable (iterations / compress level)


src/shared/missing-preact-types.d.ts
```diff
@@ -1,7 +1,11 @@
 declare module 'preact' {
   namespace JSX {
     interface HTMLAttributes {
-      decoding?: 'sync' | 'async' | 'auto';
+      decoding?:
+        | 'sync'
+        | 'async'
+        | 'auto'
+        | SignalLike<'sync' | 'async' | 'auto' | undefined>;
```
jakearchibald: Should this be here? (it's causing the build to fail)

jakearchibald:
This is exciting!

src/shared/missing-preact-types.d.ts
Csabesz07: Without that I get this error: `Subsequent property declarations must have the same type.  Property 'decoding' must be of type '"sync" | "async" | "auto" | SignalLike<"sync" | "async" | "auto" | undefined> | undefined', but here has type '"sync" | "async" | "auto" | undefined'`.

Csabesz07: @jakearchibald After further investigation I found that the error was caused by a package version error. After fixing that, the error is gone, so the code you mentioned

surma:
@Csabesz07 Your code seems to be working :) If you compress a bigger image, the time it take to compress is hugely different although in this case, the compressed file size is _almost_ identical, which is not completely unexpected :D

Screenshot 2024-04-22 at 09 00 01

I’ll do a review in a bit :)

codecs/oxipng/src/lib.rs
```diff
@@ -20,6 +24,23 @@ pub fn optimise(
     } else {
         Interlacing::None
     });
+    match deflater {
```
surma: Would be nice to introduce a type here first, [like we have in the `resize` module](https://github.com/GoogleChromeLabs/squoosh/blob/19beb1a7ab5ab7df9625edaf7c3bf71a50e183ae/codecs/resize/src/lib.rs#L71-L77), so we can match on semantically meaningful words rather than numbers.

Also, invalid numbers should be `panic()`’d, not just use a default value.

codecs/oxipng/src/lib.rs
```diff
@@ -20,6 +24,23 @@ pub fn optimise(
...
+        0 => {
+            options.deflate = Deflaters::Libdeflater {
+                compression: compression_level,
+            }
+        }
+        1 => {
+            options.deflate = Deflaters::Zopfli {
+                iterations: Option::unwrap(NonZeroU8::new(iterations)),
```
surma:
```suggestion
                iterations: NonZeroU8::new(iterations).unwrap(),
```

surma (changes requested):

jakearchibald:
> the time it take to compress is hugely different although in this case, the compressed file size is _almost_ identical, which is not completely unexpected :D

I'm not seeing any difference when changing the deflater, or the iterations / compress level setting, which feels a bit sus

codecs/oxipng/src/lib.rs
Csabesz07: Thanks for the review @surma !

Is this what you were thinking of e3bb3a946cd06cc491b4682ac02fa80f6dc00c7f ?

Csabesz07:
> I'm not seeing any difference when changing the deflater, or the iterations / compress level setting, which feels a bit sus

Exactly! This is my concern as well, but the way I see, all options are being applied at the moment. 
Oxipng documentation mentions that if any of the optimize's options's properties are defined manually, the initial preset options will be overriden with those.

surma:
I rebuilt the codec locally when testing. I wonder if the binaries on the PR are outdated or something. I’ll take another look tomorrow at that.

Csabesz07:
Hi @surma
Sorry for the slow follow-up on your change requests. After a busy year, I’m back with full focus. If this feature is still useful for the project, I’d be happy to pick up where we left off and push it forward. What’s the best next step?

PR #1423, opened 2024-08-07
Replace deprecated terser plugin

surma:
`rollup-plugin-terser` has been long deprecated and replaced by `@rollup/plugin-terser`.

I discovered this during some Wasm work, where Emscripten now emits code using `??=` which the older terser does not understand. The switch to `@rollup/plugin-terser` fixes that.

There are lots of outdated dependencies, but rather than do a massive PR I figured I’d do it one step at a time.


surma:
... Turns out we were also still on Node 14 in `.nvmrc`, which doesn’t have `someArray.at()`, which the terser plugin relies on.

So this PR also updates to Node 20 LTS.

surma:
No idea why Windows is failing or what I need to do to fix it 😕

aryanpingle:
Judging purely from the error report, [this could be the issue](https://stackoverflow.com/questions/78420489/error-when-running-npm-script-with-yarn-err-spawn-einval)? It's a breaking change in node for Windows users

surma:
@aryanpingle Great find! That seems to have resolved _that_ issue, however now there’s a new path issue 🙄 

I’ll see if I can spin up a Windows workstation somewhere...

aryanpingle:
@surma GH actions runs `npm run build`, not `npm run dev` - so it hasn't caught the fact that you can't set env vars like `DEV_PORT=\"${DEV_PORT:=5000}\"` on Windows. Can we go back to the hardcoded `--listen=5000`?

In the future if we need to use the port number elsewhere we can just declare it separately

jakearchibald (approved):

jakearchibald:
Thank you!

PR #1430, opened 2024-10-05
New feature

yogender-316:
As for new users like me who is using windows the script is not working so i made changes so that windows user can easily run this project on their machine

you just have to change the script to
"scripts": {
    "build": "rollup -c && node lib\\move-output.js",
    "debug": "node --inspect-brk node_modules\\.bin\\rollup -c",
    "dev": "set DEV_PORT=5000 && run-p watch serve",
    "watch": "rollup -cw",
    "serve": "serve --listen=%DEV_PORT% --config ../../../serve.json .tmp\\build\\static",
    "prepare": "husky install"
  },
in package.json and after that run as mentioned
npm i
then npm run build
then npm run dev

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,1 +1,2 @@
 npx lint-staged
+npm test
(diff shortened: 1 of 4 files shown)


ngdream:
thanks , very helpful

Issue #1449, opened 2025-08-30
FAQ (in Wiki) links to non-existant `/cli`


Kissaki:
https://github.com/GoogleChromeLabs/squoosh/wiki/FAQ

> Where is the CLI?: Right [here](https://github.com/GoogleChromeLabs/squoosh/tree/dev/cli).

bailey-coding:
It was removed in [13a185def2a48818dfc2aad5c933b7ab0b5cb95d](https://github.com/GoogleChromeLabs/squoosh/commit/13a185def2a48818dfc2aad5c933b7ab0b5cb95d)

PR #1468, opened 2026-05-05
Remove reference to non-existent FAQ from bug report template (issue #1449)

85ED:
Removed the line referencing the FAQ page in the bug report template, as the page no longer exists. This fixes issue #1449 and prevents users from being directed to a broken link when opening a bug report.


85ED:
CLA signed.

PR #1473, opened 2026-05-29
Updating AVIF & JXL

jakearchibald:


tomByrer:
Would be nice if updated & merged, since browsers re-added JXL support.

jakearchibald:
JXL support was not removed from any browser that had official support for it.

